Issue #92, opened 2020-09-29
Support local backend for TF driver


findkim:
The goal of tech preview is for development, testing, and feedback. With testing it's likely that users will connect Consul Sync with a temporary instance of Consul, like dev mode `consul agent -dev`, and spin it up and down as needed resulting in this common scenario of wiping state files for tasks #83 (non-persisting KV).

It would be useful to offer the option to use the local backend and store state files on disk with the generated root modules.

PR #101, opened 2020-10-05
Add support for local Terraform backend

lornasong:
We want to support local backend in addition to Consul backend. This will support users who need to persist their state file when they spin up/down their Consul agent.

Include e2e tests to document how configuring local backend’s `path` and `workspace_dir` plays out for two tasks. Note: configuring `path` is not meaningful for Sync since `path` is used for the default workspace. Sync creates a custom workspace for each task and therefore does not work in the default workspace. This configuration information will be documented separately. See https://github.com/hashicorp/terraform/blob/master/backend/local/backend.go#L445 for more details on `path`

Resolves: https://github.com/hashicorp/consul-terraform-sync/issues/92

### DIFF
--- a/config/terraform.go
+++ b/config/terraform.go
@@ -239,7 +239,9 @@ func (c *TerraformConfig) Validate() error {
 	}
 
 	for k := range c.Backend {
-		if k != "consul" {
+		switch k {
+		case "consul", "local":
+		default:
 			return fmt.Errorf("unsupported Terraform backend by Sync %q", k)
 		}
 	}
@@ -247,6 +249,7 @@ func (c *TerraformConfig) Validate() error {
 	return nil
 }
 
+// GoString defines the printable version of this struct.
 func (c *TerraformConfig) GoString() string {
 	if c == nil {
 		return "(*TerraformConfig)(nil)"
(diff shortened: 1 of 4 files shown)


lornasong:
@findkim - thanks for the feedback. I added two e2e tests that show how `workspace_dir` plays out for two tasks. It shows the path for where the state file is stored when `workspace_dir` is set. I also include an e2e test for how `path` plays out and an explanation for why it is ignored. Let me know anymore feedback!

config/terraform.go
```diff
@@ -239,14 +239,15 @@ func (c *TerraformConfig) Validate() error {
 	}
 
 	for k := range c.Backend {
-		if k != "consul" {
+		if k != "consul" && k != "local" {
```
eikenb, line 242: This is super small thing and not really much of an issue, particularly with only 2 possibilities, but I find that once you start having more test cases like this it reads better as a switch statement. 
```
switch k {
case "consul", "local": // good
default:
    fmt.Errorf(...)
}

findkim (approved):

config/terraform.go
lornasong: Thanks for raising this! I debated going with the switch-case statement for awhile but opted for the if-statement. We'll want to support remote backend (tfc/e) in the near future... I switched to the switch-statement so we can be set-up for success :)

PR #120, opened 2020-10-15
HCL marshal services variable

findkim:
Now that the services dep is public, we can create a custom function to marshal into HCL instead of manually templating each attribute.

No UX changes, just improves the code management of templating the services variable


lornasong (approved):
Having HCLService makes things look a lot cleaner!

PR #121, opened 2020-10-19
set default retry count

findkim:
I think the worker retry https://github.com/hashicorp/consul-terraform-sync/pull/72 was somehow missed in the refactor efforts https://github.com/hashicorp/consul-terraform-sync/pull/81. `defaultRetry` is 2 and was used in #72 and no longer referenced anywhere after #81


lornasong (approved):
yes! thanks for catching this :)

Issue #60, opened 2020-09-14
Only call out-of-band handlers when changes have occurred


lornasong:
Currently, after units of work are cycled through in controller, all post-apply handlers are executed regardless if units of work were actually updated.

Optimize this so that handlers are called only as needed e.g. only execute panos out-of-band commit when a panos resource has been updated

PR #122, opened 2020-10-20
Update handlers to triggered per/task

lornasong:
This is an improvement to trigger handlers at the driver-layer rather than the controller-layer. Previously handlers for all tasks were retrieved at the controller-layer so that all handlers were triggered after each task completed. This leads to handlers unnecessarily being triggered by unrelated tasks. This changes retrieves handlers for a given task and stores it on the task’s driver.

Resolves: https://github.com/hashicorp/consul-terraform-sync/issues/60 - work was previously done to determine a task has been updated. This PR only ensures the right handler is called (vs. all handlers) when a task is updated.


findkim (approved):
This is great shift of code to get the out-of-band commit closer to the task's corresponding `terraform apply`.  It's testament to how well the implementation was executed to move a modular feature around.

driver/task.go
```diff
@@ -26,8 +26,8 @@ type Service struct {
 type Task struct {
 	Description  string
 	Name         string
-	Providers    []map[string]interface{}
-	ProviderInfo map[string]interface{}
+	Providers    []map[string]interface{} // task.providers config info
+	ProviderInfo map[string]interface{}   // driver.required_provider config info
```
findkim, line 30: I struggled naming these. thank you for adding comments 🙇‍♂️

PR #123, opened 2020-10-21
Remove inspect mode from config file option

findkim:
This aligns the daemon modes to be controlled only by CLI flags, similar to the support for Once.

Having inspect mode as a config option requires multiple steps for a debugging option to change modes that would be better without -- write config file, save, run the program, then later open config, delete option, save, run again

Note, at this point `inspect` flag is not available and is not implemented yet, just shifts existing code around

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -106,18 +106,17 @@ func (cli *CLI) Run(args []string) int {
 		return ExitCodeOK
 	}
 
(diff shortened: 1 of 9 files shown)


lornasong (approved):
nice! agree, having inspect mode as a cli will be a smoother ux

PR #126, opened 2020-10-22
Refactor worker and driver

findkim:
PR proposes to condense down the layers of abstraction. It started growing to where all information were being passed down to each level controller > driver > worker > task.

## Worker
* Moves worker code to separate file
* Refactors worker to reduce the context required for a worker - no longer needs info on client, init history, task.

## Driver
 * Moves driver install outside of controller workflow so it is only checked and installed once and earlier, instead of installing/verifying install for each task since there can only be one driver version used across all tasks
* Doubles down on the recent design choice for a driver instance per task.

### Terraform Driver
* TF driver now manages the TF client directly and initializes it within the constructor. This no longer requires the driver to store extra info that only pertains to the client (log info, path, client type). Previously the driver passed info to the worker to then initialize the client.
* Moves initialization calls outside of `InitTask` and into `NewTerraform`: worker, handlers
* Condensed redundant workspace init handling to a helper

### DIFF
--- a/driver/terraform.go
+++ b/driver/terraform.go
@@ -5,9 +5,8 @@ import (
 	"fmt"
 	"log"
 	"os"
-	"path/filepath"
-	"strings"
 
+	"github.com/hashicorp/consul-terraform-sync/client"
 	"github.com/hashicorp/consul-terraform-sync/handler"
 	"github.com/hashicorp/consul-terraform-sync/templates/tftmpl"
 	"github.com/pkg/errors"
@@ -38,21 +37,22 @@ var (
 // Terraform is an Sync driver that uses the Terraform CLI to interface with
 // low-level network infrastructure.
 type Terraform struct {
-	log               bool
-	persistLog        bool
-	path              string
-	workingDir        string
-	worker            *worker
+	task              Task
 	backend           map[string]interface{}
 	requiredProviders map[string]interface{}
-	postApply         handler.Handler
 
-	version    string
+	workingDir string
+	worker     *worker
 	clientType string
+	client     client.Client
+	postApply  handler.Handler
+
+	inited bool
 }
 
 // TerraformConfig configures the Terraform driver
 type TerraformConfig struct {
+	Task              Task
 	Log               bool
 	PersistLog        bool
 	Path              string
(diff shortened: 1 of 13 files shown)


driver/terraform.go
```diff
@@ -65,71 +65,52 @@ type TerraformConfig struct {
...
 }
 
 // Version returns the Terraform CLI version for the Terraform driver.
 func (tf *Terraform) Version() string {
-	return tf.version
+	return TerraformVersion
 }
 
 // InitTask initializes the task by creating the Terraform root module, creating
 // client to execute task, and retrieving post-Terraform apply handlers
```
lornasong, line 111: I think you could remove the stuff about creating client and retrieving the handlers from the comment

driver/terraform.go
```diff
@@ -65,71 +65,52 @@ type TerraformConfig struct {
...
-func (tf *Terraform) InitTask(task Task, force bool) error {
-	modulePath := filepath.Join(tf.workingDir, task.Name)
-	if _, err := os.Stat(modulePath); os.IsNotExist(err) {
-		if err := os.Mkdir(modulePath, workingDirPerms); err != nil {
-			log.Printf("[ERR] (driver.terraform) error creating task work directory: %s", err)
-			return err
-		}
-	}
+func (tf *Terraform) InitTask(ctx context.Context, force bool) error {
```
lornasong, line 112: I'm not seeing context being consumed. I might have missed it. is it being used or plans for later?

driver/worker.go
```diff
@@ -0,0 +1,103 @@
...
+	"log"
+	"math/rand"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+// worker executes a unit of work and has a one-to-one relationship with a client
+// that will be responsible for executing the work. Currently worker is not safe for
+// concurrent use by multiple goroutines
```
lornasong, line 15: I think you can remove the stuff about client from the comment since this PR removes the client from worker

lornasong (approved):
Looking forward to these changes. Particularly excited about the driver install stuff!

driver/terraform.go
```diff
@@ -65,71 +65,52 @@ type TerraformConfig struct {
...
-	return nil
+
+	return &Terraform{
+		task:              config.Task,
+		backend:           config.Backend,
+		requiredProviders: config.RequiredProviders,
+		workingDir:        config.WorkingDir,
+		worker:            newWorker(defaultRetry),
+		client:            tfClient,
+		postApply:         handler,
```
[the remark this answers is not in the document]
findkim: Great catch, removing!

driver/terraform.go
findkim: yeah it's not being used. i was iffy on whether to keep the function signature consistent across the driver methods (Inspect and Apply both have ctx). Would you prefer this to be removed until used?

driver/terraform.go
findkim: Appreciate your attention to detail, i missed these comments when updating. Will do!

driver/terraform.go
lornasong: ah I see. I personally lean towards removing until use but I also see what you're saying about consistency. I'm ok either way!

driver/worker.go
findkim: 👍 thanks for drawing my attention to the godocs. I made a pass and updated a number of other ones that were missing or affected by the refactor.

driver/terraform.go
findkim: Ok, will remove. `InitTask` in nature is different than the others that are longer running functions and has more of a consisten/predictable execution time.

findkim: I also added context handling in the `baseController.init(ctx)` to stop looping through tasks early if needed.

Issue #127, opened 2020-10-23
Expand support for Terraform 0.14


findkim:
#### Description
The v0.1.0-techpreview1 release only supports Terraform v0.13 ([source](https://github.com/hashicorp/consul-terraform-sync/blob/v0.1.0-techpreview1/templates/tftmpl/root.go#L22)). Terraform v0.14 is nearing release and there are non-breaking syntax changes between 0.13 and 0.14. This would allow users and module authors to expand the features available for CTS module integrations.

[Terraform v0.14 Changelog](https://github.com/hashicorp/terraform/blob/v0.14/CHANGELOG.md)

#### Use Cases
Expanding support for Terraform 0.14 would allow CTS to run modules with improved security, like using the new `sensitive` argument and would resolve https://github.com/hashicorp/consul-terraform-sync/issues/85

#### Additional context
Is there a way to configure which Terraform version to run? What happens to task execution when running Terraform v0.13 with a module containing newer features only supported by v0.14?

PR #129, opened 2020-10-28
Expand usage of Terraform backends for state store

findkim:
Enables usage of other Terraform backends for state store that support [workspaces](https://www.terraform.io/docs/state/workspaces.html) and state locking.

Tests capture testing the integration of passing through backend configuration to the generated root main.tf file. Validity of the backend configuration options are done at task execution time. Each backend were not tested directly to avoid testing Terraform functionality, but I did test the [postgres](https://www.terraform.io/docs/backends/types/pg.html) and came across a bug that prevents CTS from allowing it to be used with Terraform 0.13.

This PR does not allow usage of the pg backend since CTS is currently bounded to Terraform 0.13. However it will be accessible soon as we're looking to expand support for Terraform 0.14 (#127).

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -116,7 +116,12 @@ func (ctrl *baseController) init(ctx context.Context) error {
 func InstallDriver(ctx context.Context, conf *config.Config) error {
 	if conf.Driver.Terraform != nil {
 		tfConf := *conf.Driver.Terraform
-		return driver.InstallTerraform(ctx, *tfConf.Path, *tfConf.WorkingDir)
+		err := driver.InstallTerraform(ctx, *tfConf.Path, *tfConf.WorkingDir)
+		if err != nil {
+			return err
+		}
+
+		return tfConf.CheckVersionCompatibility(driver.TerraformVersion)
 	}
 	return errors.New("unsupported driver")
 }
--- a/config/terraform.go
+++ b/config/terraform.go
@@ -5,6 +5,8 @@ import (
 	"log"
 	"os"
 	"path"
+
+	"github.com/hashicorp/go-version"
 )
 
 const (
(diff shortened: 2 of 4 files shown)


controller/controller.go
```diff
@@ -116,7 +116,12 @@ func (ctrl *baseController) init(ctx context.Context) error {
 func InstallDriver(ctx context.Context, conf *config.Config) error {
 	if conf.Driver.Terraform != nil {
 		tfConf := *conf.Driver.Terraform
-		return driver.InstallTerraform(ctx, *tfConf.Path, *tfConf.WorkingDir)
+		err := driver.InstallTerraform(ctx, *tfConf.Path, *tfConf.WorkingDir)
```
eikenb, line 119: Maybe check the Terraform version before downloading/installing it?

eikenb:
Good choice of where to draw the testing line. :+1:

controller/controller.go
findkim: the UX of this is a bit awkward tbh. Users can't set which Terraform version to run yet (just the latest 0.13), and the work-around is to manually install the desired version and point to it. An existing Terraform at the path would skip the install in fetch its version to then compare with the config.

An improved order would be something like:
* Configured version > validate compat > install
* No configured version > check existing version > validate compat > install
* No configured version > no existing version > fetch latest version > validate compat > install

I had made these changes with minimal impact in mind but you pointing this out makes it more convincing to reorganize the code so it's more logical.

eikenb: I thought there might be a reasons but digging around I couldn't find them. I'm cool with it as a working solution if you want to move forward for now and circle back to this later.

I'll approve this PR and leave that to you.

eikenb (approved):

controller/controller.go
findkim: Appreciate it John! I'll merge it and iterate on this as a follow-up.

lornasong:
Nice find with the postgres issue with Terraform 0.13!

PR #130, opened 2020-10-28
Capture data on process of updating network infrastructure in `Event`

lornasong:
First part in a series of changes to support a status api. This captures data in
an `Event` struct each time CTS goes through the process needed for updating
network infrastructure for a task. Future changes will store the event data and
return various aggregations of the data at the api endpoint

Changes
 - Create `event` package & structure
 - Consume event in controller checkApply()
 - Update `unit` with provider, service, and source in order to configure `event`
 - Update `Task` with ProviderNames() & ServicesNames to help configure `event`

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -27,6 +27,7 @@ type Controller interface {
 	Run(ctx context.Context) error
 }
 
+// Oncer describes the interface a controller that can run in once mode
 type Oncer interface {
 	Once(ctx context.Context) error
 }
@@ -36,6 +37,10 @@ type unit struct {
 	taskName string
 	driver   driver.Driver
 	template template
+
+	providers []string
+	services  []string
+	source    string
 }
 
 type baseController struct {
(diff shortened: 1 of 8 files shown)


lornasong:
`Event` is the rename of the original proposed `Entry`. It represents the actions that CTS goes through to update network infrastructure. Any other naming ideas/opinions welcome. Here’s all that I had brainstormed: apply, update, run, execution, record.

controller/readwrite_test.go
```diff
@@ -26,7 +26,7 @@ func TestReadWriteRun(t *testing.T) {
 		applyTaskErr      error
 		resolverRunErr    error
 		templateRenderErr error
-		watcherWaitErr    error
+		taskName          string
```
lornasong, line 29: Note: I made 2 additional updates to this test
 - removed the case field for watcher.Wait: this method no longer calls watcher.Wait
 - added the case for template.Render

controller/controller.go
```diff
@@ -101,9 +106,12 @@ func (ctrl *baseController) init(ctx context.Context) error {
...
 		units = append(units, unit{
-			taskName: task.Name,
-			template: template,
-			driver:   d,
+			taskName:  task.Name,
+			template:  template,
+			driver:    d,
+			providers: task.ProviderNames(),
+			services:  task.ServiceNames(),
+			source:    task.Source,
```
eikenb, line 114: This is starting to seem like a copy of a task, we're pulling so much from it. At the moment I think it is fine but we might want to review this in the future to see if we want to keep copying all these values or just use the task itself.

eikenb (approved):
LGTM

controller/controller.go
lornasong: :nod: I debated just using task as a field but didn't feel quite ready. Agree, we'll keep an eye on this and switch over to task if another field needs to be added.

PR #132, opened 2020-10-30
switch to hashicorp docker mirror

alvin-huang:


eikenb:
This should take care of our team's repos. Thanks @alvin-huang!

PR #133, opened 2020-11-03
Pull in hcat bug fix on Consul connect timeouts

findkim:
The changes fixes the case where CTS would hang indefinitely when
running on a machine with DNS issues that cause timeouts while
connecting to Consul. Added a log line to provide users context
about attempting Consul connection for easier debugging connection
issues instead of hanging silently.

Context: https://github.com/hashicorp/hcat/issues/23

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -57,10 +57,13 @@ func newBaseController(conf *config.Config) (*baseController, error) {
 	if err != nil {
 		return nil, err
 	}
+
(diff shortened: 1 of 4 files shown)


findkim:
Looks like i need to update the template helper functions with recent changes to hcat lib.

findkim:
Updated! hcat helper funcs were friendly to include 🎉

lornasong (approved):
looks good to me!

PR #135, opened 2020-11-10
Store event data

lornasong:
Second part in a series of changes to support a status api. This stores event
data in memory when running in daemon mode. Future changes will return
aggregations of the stored events at the api endpoint.

Changes
 - Create `store` struct with ability to add and read
 - Update readwrite controller (+ tests) with a `store` field
 - Store events that are crated in checkApply()

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/hashicorp/consul-terraform-sync/config"
 	"github.com/hashicorp/consul-terraform-sync/driver"
+	"github.com/hashicorp/consul-terraform-sync/event"
 	mocksD "github.com/hashicorp/consul-terraform-sync/mocks/driver"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
@@ -19,11 +20,6 @@ import (
 
 func TestNewControllers(t *testing.T) {
 	t.Parallel()
-
-	newCtrlFuncs := map[string]func(*config.Config) (Controller, error){
-		"readwrite": NewReadWrite,
-		"readonly":  NewReadOnly,
-	}
 
 	cases := []struct {
 		name        string
@@ -65,17 +61,24 @@ func TestNewControllers(t *testing.T) {
 				tc.conf.Finalize()
 			}
 
-			for name, newCtrlFunc := range newCtrlFuncs {
-				t.Run(name, func(t *testing.T) {
-					controller, err := newCtrlFunc(tc.conf)
-					if tc.expectError {
-						assert.Error(t, err)
-						return
-					}
-					assert.NoError(t, err)
-					assert.NotNil(t, controller)
-				})
-			}
+			t.Run("readwrite", func(t *testing.T) {
+				controller, err := NewReadWrite(tc.conf, event.NewStore())
+				if tc.expectError {
+					assert.Error(t, err)
+					return
+				}
+				assert.NoError(t, err)
+				assert.NotNil(t, controller)
+			})
+			t.Run("readonly", func(t *testing.T) {
+				controller, err := NewReadOnly(tc.conf)
+				if tc.expectError {
+					assert.Error(t, err)
+					return
+				}
+				assert.NoError(t, err)
+				assert.NotNil(t, controller)
+			})
 		})
 	}
 }
--- a/controller/readwrite_test.go
+++ b/controller/readwrite_test.go
@@ -119,7 +119,6 @@ func TestReadWrite_CheckApply(t *testing.T) {
 				assert.Error(t, err)
 				assert.Contains(t, err.Error(), tc.name)
 				assert.False(t, event.Success)
-				fmt.Println(event)
 				assert.NotNil(t, event.EventError.Message)
 				assert.Contains(t, event.EventError.Message, tc.name)
 			} else {
--- a/controller/readwrite_test.go
+++ b/controller/readwrite_test.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/hashicorp/consul-terraform-sync/config"
 	"github.com/hashicorp/consul-terraform-sync/driver"
+	"github.com/hashicorp/consul-terraform-sync/event"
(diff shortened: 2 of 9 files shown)


controller/controller_test.go
```diff
@@ -20,11 +21,6 @@ import (
 func TestNewControllers(t *testing.T) {
 	t.Parallel()
 
-	newCtrlFuncs := map[string]func(*config.Config) (Controller, error){
-		"readwrite": NewReadWrite,
-		"readonly":  NewReadOnly,
-	}
-
```
lornasong, line 27: I couldn't figure out a way to keep this pattern! :(

main.go
```diff
@@ -1,8 +1,13 @@
...
-import "os"
+import (
+	"os"
+
+	"github.com/hashicorp/consul-terraform-sync/event"
+)
 
 func main() {
-	cli := NewCLI(os.Stdout, os.Stderr)
+	store := event.NewStore()
```
findkim, line 10: The intent of why the store is initialized at this level isn't clear to me, could you expand on this?

controller/controller_test.go
findkim: haha no worries, it was just a way i grouped those tests at the time, things will definitely evolve and it might make more sense to separate testing the two in the future.

controller/readwrite_test.go
```diff
@@ -89,24 +90,84 @@ func TestReadWrite_CheckApply(t *testing.T) {
...
+			event := events[0]
+			assert.Equal(t, tc.taskName, event.TaskName)
+			assert.False(t, event.StartTime.IsZero())
+			assert.False(t, event.EndTime.IsZero())
+
+			if tc.expectError {
+				assert.Error(t, err)
+				assert.Contains(t, err.Error(), tc.name)
+				assert.False(t, event.Success)
+				fmt.Println(event)
```
findkim, line 122: 👀 `fmt.Println`

event/store.go
```diff
@@ -0,0 +1,60 @@
...
+	*sync.RWMutex
+
+	events map[string][]*Event // taskname => events
+	limit  int
+}
+
+// NewStore returns a new store
+func NewStore() *Store {
+	return &Store{
+		RWMutex: &sync.RWMutex{},
```
findkim, line 22: For precautionary measures, the mux would be more controlled as a private attribute so callers of the NewStore could not maliciously hold onto a lock outside of the context of the store methods.

event/store.go
```diff
@@ -0,0 +1,60 @@
...
+
+	s.Lock()
+	defer s.Unlock()
+
+	events := s.events[e.TaskName]
+	events = append([]*Event{e}, events...) // prepend
+	if len(events) > s.limit {
+		events = events[:len(events)-1]
+	}
+	s.events[e.TaskName] = events
```
findkim, line 45: Go lists are by reference and modifying `events` variable updates the list within the map too ([example](https://play.golang.com/p/ZbtITuQo-3t)). No need to change but seeing this thought i'd mention it

event/store_test.go
```diff
@@ -0,0 +1,105 @@
...
+			&Event{TaskName: "happy"},
+			false,
+		},
+		{
+			"error: no taskname",
+			&Event{},
+			true,
+		},
+		{
+			"nil even",
```
findkim, line 26: `nil event`?

event/store.go
```diff
@@ -0,0 +1,60 @@
...
+func NewStore() *Store {
+	return &Store{
+		RWMutex: &sync.RWMutex{},
+		events:  make(map[string][]*Event),
+		limit:   defaultEventCountLimit,
+	}
+}
+
+// Add adds an event and manages the limit of number of events stored per task.
+func (s *Store) Add(e *Event) error {
```
findkim, line 29: Can you change the event parameter to pass by copy instead of reference? That way the caller can't unintentionally change event values once it's already added to the store

findkim (commented):
The store is looking good

main.go
lornasong: Yes! Sorry, it's a bit premature. I am planning to initialize the API in the main in the next PR, something like `NewAPI(store)` so that the status API could access the store. Let me know what you think!

controller/readwrite_test.go
lornasong: oh no! thanks for catching

event/store.go
lornasong: good point! let me do this!

event/store.go
lornasong: Ah yes! I tried removing this line and discovered that `append` creates and returns a new slice so events needs to be reassigned. I'll also keep in mind though that events is a reference, which I had forgotten!

event/store_test.go
lornasong: Thank you!

event/store.go
lornasong: Oh yes! Good point, thank you!

event/store.go
findkim: oh true on the `append` 👍

event/store.go
findkim: honestly there's a lot of this that I probably sprinkled throughout the code too for the convenience of `nil`. i'll try to update them as I see.

main.go
findkim: Gotcha. Ok it'll likely require some refactoring since the configs are parsed within `NewCLI` presuming `NewAPI` will need that info too

findkim (approved):
✨

main.go
lornasong: Yes! Exactly as you say. I'll have to figure out how to share the configs across the cli and api and am planning to do that in the next PR. Sorry for the confusion of making early moves in adding store at this level!

lornasong: 🤔 I thought about your question some more. I like the neatness of explicitly creating and executing CLI and API in the main. But sharing the config stuff will be messier than if the cli calls NewAPI. I think I'll try both ways and see how they look. Let me know if you have thoughts!

Issue #134, opened 2020-11-07
Error: Failed to select workspace: input not a valid number


chkp-jguo:
### Describe the bug
I get the following error message the 2nd time I run consul-terraform-sync. 

The currently selected workspace (sample) does not exist.
  This is expected behavior when the selected workspace did not have an

  existing non-empty state. Please enter a number to select a workspace:

  1. default

  Enter a value:
Error: Failed to select workspace: input not a valid number

2020/11/07 21:21:45.256735 [ERR] (driver.terraform) init (skip apply) for 'sample'
2020/11/07 21:21:45.257073 [ERR] (cli) error running controller: [ERR] (controller.readwrite): could not apply: Error tf-init for 'sample': attempt #1 failed '
Error: Failed to select workspace: input not a valid number

### Versions
**Consul Terraform Sync**
```
consul-nia 8885a20 (8885a20)
Compatible with Terraform ~>0.13.0
```

**Consul Version**
```
Consul 1.8.4
```

**Terraform Version**
```
Terraform v0.13.3
```

### Configuration File(s)
_Reminder to redact any sensitive information that may be present in this file_

<details>
<summary>Click to toggle contents of config file </summary>

```hcl

log_level = "info"

consul {
    address = "192.168.0.1:8500"
}

buffer_period {
    min = "5s"
    max = "20s"
}

task {
    name = "sample"
    description = "This task dynamically updates service addresses"
    source = "../../"
    providers = ["checkpoint"]
    services = ["web", "app", "db"]
}

driver "terraform" {
  log = true
  required_providers {
    checkpoint = {
      source = "CheckPointSW/checkpoint"
    }
  }
}

provider "checkpoint" {
  server = "192.168.0.5"
  username = "user123"
  password = "test123"
  context = "web_api"
  timeout = 60
}

```
</details>

### Terraform Configuration Files Generated by Consul-Terraform-Sync
_Reminder to redact any sensitive information that may be present in the files_

<details>
<summary>Click to toggle contents of main.tf </summary>

```terraform
# This file is generated by Consul Terraform Sync.
#
# The HCL blocks, arguments, variables, and values are derived from the
# operator configuration for Sync. Any manual changes to this file
# may not be preserved and could be clobbered by a subsequent update.

terraform {
  required_version = "~>0.13.0"
  required_providers {
    checkpoint = {
      source = "CheckPointSW/checkpoint"
    }
  }
  backend "consul" {
    address = "123.123.123.123:8500"
[...]

PR #136, opened 2020-11-11
Fix err for non-default workspace with empty state

### LINKED ISSUE #83: Improve user error when statefile is wiped on Consul restart
Users who are doing local development and testing with NIA may connect to a local Consul server, running some variant of:

```
$ consul agent -dev
```

when you restart the server, the KV is not persisted and erased. KV is where the Consul NIA stores the TF statefiles for tasks, resulting in task execution errors regarding workspace (visible when setting `driver.terraform.log=true` in the NIA config). The workspace is created by the TF client and is persisted on disk in the `nia-tasks` directory for each task.

```
2020/09/25 18:48:24.213053 [INFO] running Terraform command: PATH/terraform init -no-color -force-copy -input=false -lock-timeout=0s -backend=true -get=true -get-plugins=true -lock=true -upgrade=false -verify-plugins=true
Initializing modules...

Initializing the backend...

The currently selected workspace (print) does not exist.
  This is expected behavior when the selected workspace did not have an
  existing non-empty state. Please enter a number to select a workspace:
  
  1. default

  Enter a value: 

Error: Failed to select workspace: input not a valid number
```

I ran into this often with my local setup and would be useful to improve the error surfaced to end users to help understand and take action on.

findkim:
Executing terraform init for a non-default workspace that has an empty state causes an error. This adds edge case handling of the Terraform error to init, check the error, create workspace (which also sets up the backend), and init again.

This error is commonly ran into when using the default CTS option for Terraform state store, Consul KV. During development, the Consul server may be restarted and wiping the in-memory Consul KV along with the Terraform state files for CTS tasks. This also happens for other backends, in addition to Consul KV, when an existing task's state has been deleted because `.terraform/environment` manifest file with the workspace name is written to disk and subsequent inits have context of that workspace but not its state.

I chose to leave the `.terraform/environment` file as-is, and not editing or deleting it, to possibly preserve any other workspaces for the working directory CTS may not be aware of.

Resolves #83 #134

### DIFF
--- a/client/terraform_cli.go
+++ b/client/terraform_cli.go
@@ -6,12 +6,17 @@ import (
 	"fmt"
 	"log"
 	"path/filepath"
+	"regexp"
 
 	"github.com/hashicorp/consul-terraform-sync/templates/tftmpl"
 	"github.com/hashicorp/terraform-exec/tfexec"
 )
 
-var _ Client = (*TerraformCLI)(nil)
+var (
+	_ Client = (*TerraformCLI)(nil)
+
+	wsFailedToSelectRegexp = regexp.MustCompile(`Failed to select workspace`)
+)
 
 // TerraformCLI is the client that wraps around terraform-exec
 // to execute Terraform cli commands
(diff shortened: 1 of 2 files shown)


findkim:
Ah hold on, this now breaks the happy path. Let me tinker a bit more before review

findkim:
The initial approach was to select the default workspace first, then attempt init. However init would error on a new working directory.

I changed the approach on handling this edge case to follow the logic:
1. init
1. check error
1. create workspace -- this sets up the local manifest `.terraform/environment` as well as the backend if necessary. For Consul KV this step is needed. 
1. init again to setup module and providers.

client/terraform_cli_test.go
```diff
@@ -148,8 +148,42 @@ func TestTerraformCLIInit(t *testing.T) {
 			}
 
 			assert.NoError(t, err)
+			m.AssertExpectations(t)
```
lornasong, line 151: Oooo I've never used before ✨

lornasong (approved):
Cool. The tf flow looks tricky. Thanks for figuring this out!

client/terraform_cli.go
```diff
@@ -85,20 +90,42 @@ func NewTerraformCLI(config *TerraformCLIConfig) (*TerraformCLI, error) {
 // Init initializes by executing the cli command `terraform init` and
 // `terraform workspace new <name>`
 func (t *TerraformCLI) Init(ctx context.Context) error {
+	var wsCreated bool
+
+	// This is special handling for when the workspace has been detected in
+	// .terraform/environment without a non-empty state. This case is common
+	// when the state for the workspace has been deleted.
+TF_INIT_AGAIN:
```
eikenb, line 98: Normally I prefer explicit loops instead of goto's but this does read pretty well. Conveys (better than a loop would) that this should work the first time, except for this one situation where we need to try again.

client/terraform_cli_test.go
```diff
@@ -148,8 +148,42 @@ func TestTerraformCLIInit(t *testing.T) {
...
 		})
 	}
+
+	t.Run("workspace artifact with empty state", func(t *testing.T) {
+		// Edge case to handle https://github.com/hashicorp/terraform/issues/21393
+		initErr := errors.New(`Initializing the backend...
```
eikenb, line 157: Does tfexec/terraform fall back into an interactive mode here? That seems.. strange.

eikenb (approved):

client/terraform_cli_test.go
findkim: Yeah it's a Terraform bug where it errors with an interactive message but TF is ran in automation with tfexec which disables input.

findkim: That being said, i'll include the GH issue where the logic is above in addition to this test case to keep context collocated

client/terraform_cli.go
findkim: 👍 I rarely use goto loops but it felt even worse writing the nesting with 1 retry logic.

PR #137, opened 2020-11-17
Support panos partial commit to admin user

findkim:
The default behavior for panos using API key commits all changes queued up by any user. This poses a security risk of auto-committing unverified changes by other users. Setting admin changes the behavior to a partial commit and only commits the changes made by the user.

**Caveat**
Panos does not have an API to fetch the associated username fro the API key. It is up to the practitioner to configure the correct username for the key. If the username does not match the key, it will attempt to auto-commit for that user, which may or may not exist or have changes, and will not commit changes made by Consul-Terraform-Sync.

Cases tested:
* [x] api key (no username): validation error on startup
* [x] api key + associated username: changes committed
* [x] api key + associated username + changes by another user: only user's changes are committed
* [x] api key + different username: no-op commit, changes made by the token are not committed because commit is looking for changes by another username.
* [x] api key + non-existing username: same as above
* [x] api key + different username with changes from that user: **edge case** it will auto-commit for that user and **not** the changes made by the token and its user.

### DIFF
--- a/handler/panos.go
+++ b/handler/panos.go
@@ -1,7 +1,9 @@
 package handler
 
 import (
+	"errors"
 	"log"
+	"os"
 	"strings"
 
 	"github.com/PaloAltoNetworks/pango"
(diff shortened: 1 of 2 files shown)


lornasong (approved):
looks good! thanks for taking this on. test cases 🎉

PR #138, opened 2020-11-17
Add basic task status endpoint

lornasong:
Third part in a series of changes to support status api. This adds basic use
case for task status endpoints `/status/tasks` and `/status/tasks/{task_name}`.
Future work listed below.

Changes
 - Change event store.Read() to support read-all when empty task name is passed
 - Change event store.Read() to return map[string][]Event instead of []Event
 - Add api package that serves the api and contains handlers
 - Removed store from main.go and serve api in cli.go rather than main.go

Since this PR was growing a bit large, subsequent PRs will have:
 - configuring the api port
 - e2e test
 - filter by status e.g. ?status=critical
 - include events in payload e.g. ?include=events
 - overall status endpoint

Examples:
`/v1/status/tasks`:
```
{
  "test_a": {
    "task_name": "test_a",
    "status": "healthy",
    "providers": [
      "local"
    ],
    "services": [
      "api",
    ],
    "events_url": "/v1/status/tasks/test_a?include=events"
  },
  "test_b": {
    "task_name": "test_b",
    "status": "critical",
    "providers": [
      "null"
    ],
    "services": [
      "web",
    ],
    "events_url": "/v1/status/tasks/test_b?include=events"
  }
}
```

`/v1/status/tasks/test_a`:
```
{
  "test_a": {
    "task_name": "test_a",
    "status": "healthy",
    "providers": [
      "local"
    ],
    "services": [
      "api",
    ],
    "events_url": "/v1/status/tasks/test_a?include=events"
  }
}
```

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -1,0 +1,91 @@
+package api
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"log"
+	"net/http"
+	"time"
+
+	"github.com/hashicorp/consul-terraform-sync/event"
+)
+
+const (
+	defaultAPIVersion = "v1"
+
+	statusHealthy      = "healthy"
+	statusDegraded     = "degraded"
+	statusCritical     = "critical"
+	statusUndetermined = "undetermined"
+)
+
+// API supports api requests to the cts biniary
+type API struct {
+	store   *event.Store
+	port    int
+	version string
+}
+
+// NewAPI create a new API object
+func NewAPI(store *event.Store, port int) *API {
+	return &API{
+		port:    port,
+		store:   store,
+		version: defaultAPIVersion,
+	}
+}
+
+// Serve starts up and handles shutdown for the http server to serve
+// API requests
+func (api *API) Serve(ctx context.Context) {
+	mux := http.NewServeMux()
+
+	// retrieve task status for a task-name
+	mux.Handle(fmt.Sprintf("/%s/%s/", defaultAPIVersion, taskStatusPath),
+		newTaskStatusHandler(api.store, api.version))
+	// retrieve all task statuses
+	mux.Handle(fmt.Sprintf("/%s/%s", defaultAPIVersion, taskStatusPath),
+		newTaskStatusHandler(api.store, api.version))
+
+	srv := &http.Server{
+		Addr:         fmt.Sprintf(":%d", api.port),
+		WriteTimeout: time.Second * 15,
+		ReadTimeout:  time.Second * 15,
+		IdleTimeout:  time.Second * 60,
+		Handler:      mux,
+	}
+
+	go func() {
+		if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			// go ahead and stop cts from continuing
+			log.Fatalf("error starting api server at '%d': '%s'\n", api.port, err)
+		}
+	}()
+
+	log.Printf("[INFO] (api) server started at port %d", api.port)
+
+	ctxShutDown, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				log.Printf("[INFO] (api) stopping api server")
+				if err := srv.Shutdown(ctxShutDown); err != nil {
+					log.Printf("[ERROR] (api) error stopping api server: '%s'", err)
+				}
+				cancel()
+				return
+			}
+		}
+	}()
+}
+
+// jsonResponse adds the return response for handlers
+func jsonResponse(w http.ResponseWriter, code int, response interface{}) {
+	w.Header().Add("Content-Type", "application/json")
+	w.WriteHeader(code)
+	json.NewEncoder(w).Encode(response)
+}
--- a/api/api.go
+++ b/api/api.go
@@ -14,10 +14,40 @@ import (
 const (
 	defaultAPIVersion = "v1"
 
-	statusHealthy      = "healthy"
-	statusDegraded     = "degraded"
-	statusCritical     = "critical"
-	statusUndetermined = "undetermined"
+	// StatusHealthy is the healthy status. This is determined based on status
+	// type.
+	//
+	// Task Status: Determined by the success of a task updating. The 5 most
+	// recent task updates are stored as an ‘event’ in CTS. A task is healthy
+	// when all the stored events are successful.
+	StatusHealthy = "healthy"
+
+	// StatusDegraded is the degraded status. This is determined based on status
+	// type.
+	//
+	// Task Status: Determined by the success of a task updating. The 5 most
+	// recent task updates are stored as an ‘event’ in CTS. A task is degraded
+	// when more than half of the stored events are successful _or_ less than
+	// half of the stored events are successful but the most recent event is
+	// successful.
+	StatusDegraded = "degraded"
+
+	// StatusCritical is the critical status. This is determined based on status
+	// type.
+	//
+	// Task Status: Determined by the success of a task updating. The 5 most
+	// recent task updates are stored as an ‘event’ in CTS. A task is critical
+	// when less than half of the stored events are successful and the most
+	// recent event is not successful
+	StatusCritical = "critical"
+
+	// StatusUndetermined is when the status is unknown. This is determined
+	// based on status type.
+	//
+	// Task Status: Determined by the success of a task updating. The 5 most
+	// recent task updates are stored as an ‘event’ in CTS. A task is
+	// undetermined when no event data has been collected yet.
+	StatusUndetermined = "undetermined"
 )
 
 // API supports api requests to the cts biniary
@@ -25,39 +55,41 @@ type API struct {
 	store   *event.Store
 	port    int
 	version string
+	srv     *http.Server
 }
 
 // NewAPI create a new API object
 func NewAPI(store *event.Store, port int) *API {
-	return &API{
-		port:    port,
-		store:   store,
-		version: defaultAPIVersion,
-	}
-}
-
-// Serve starts up and handles shutdown for the http server to serve
-// API requests
-func (api *API) Serve(ctx context.Context) {
 	mux := http.NewServeMux()
 
 	// retrieve task status for a task-name
 	mux.Handle(fmt.Sprintf("/%s/%s/", defaultAPIVersion, taskStatusPath),
-		newTaskStatusHandler(api.store, api.version))
+		newTaskStatusHandler(store, defaultAPIVersion))
 	// retrieve all task statuses
 	mux.Handle(fmt.Sprintf("/%s/%s", defaultAPIVersion, taskStatusPath),
-		newTaskStatusHandler(api.store, api.version))
+		newTaskStatusHandler(store, defaultAPIVersion))
 
 	srv := &http.Server{
-		Addr:         fmt.Sprintf(":%d", api.port),
+		Addr:         fmt.Sprintf(":%d", port),
 		WriteTimeout: time.Second * 15,
 		ReadTimeout:  time.Second * 15,
 		IdleTimeout:  time.Second * 60,
 		Handler:      mux,
 	}
 
+	return &API{
+		port:    port,
+		store:   store,
+		version: defaultAPIVersion,
+		srv:     srv,
+	}
+}
+
+// Serve starts up and handles shutdown for the http server to serve
+// API requests
+func (api *API) Serve(ctx context.Context) {
 	go func() {
-		if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+		if err := api.srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
 			// go ahead and stop cts from continuing
 			log.Fatalf("error starting api server at '%d': '%s'\n", api.port, err)
 		}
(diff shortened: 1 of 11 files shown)


api/api.go
```diff
@@ -0,0 +1,91 @@
...
+		WriteTimeout: time.Second * 15,
+		ReadTimeout:  time.Second * 15,
+		IdleTimeout:  time.Second * 60,
+		Handler:      mux,
+	}
+
+	go func() {
+		if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			// go ahead and stop cts from continuing
+			log.Fatalf("error starting api server at '%d': '%s'\n", api.port, err)
```
lornasong, line 62: I tried a couple of things here and couldn't find something that didn't start to get complicated for what I wanted to achieve. If people have ideas here, please share!

api/api.go
```diff
@@ -0,0 +1,91 @@
...
+	mux.Handle(fmt.Sprintf("/%s/%s", defaultAPIVersion, taskStatusPath),
+		newTaskStatusHandler(api.store, api.version))
+
+	srv := &http.Server{
+		Addr:         fmt.Sprintf(":%d", api.port),
+		WriteTimeout: time.Second * 15,
+		ReadTimeout:  time.Second * 15,
+		IdleTimeout:  time.Second * 60,
+		Handler:      mux,
+	}
```
findkim, line 57: This setup part of `Serve` seems like it would be better suit for the constructor `NewAPI`

api/task_status.go
```diff
@@ -0,0 +1,155 @@
...
+	return &taskStatusHandler{
+		store:   store,
+		version: version,
+	}
+}
+
+// ServeHTTP serves the task status endpoint which returns a map of taskname to
+// task status
+func (h *taskStatusHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	log.Printf("[TRACE] (api.taskstatus) requesting task status '%s'", r.URL.Path)
```
findkim, line 40: nit to rename as `api.status.tasks` to be more predictable with the API endpoint?

api/task_status.go
```diff
@@ -0,0 +1,155 @@
...
+
+	taskName, err := getTaskName(r.URL.Path, h.version)
+	if err != nil {
+		log.Printf("[TRACE] (api.taskstatus) bad request: '%s'", err)
```
findkim, line 44: nit to remove the single quotes around the error string, I think the colon before it makes it clear it's a new component to the log

api/api.go
findkim: Yeah let's look into updating this so it doesn't fatal mid-process and cause the other go routines (task execution / terraform runs) to stop abruptly.

findkim (commented):
I have reviewed about half of the PR so far. Pausing for a bit and will pick up tomorrow.

@lornasong could you include in the PR summary the routes and example JSON responses? It would help me visualize the end-goal with the code changes. Thanks!

`/v1/status/tasks`
`/v1/status/tasks/:task_name`
`/v1/status/tasks/:task_name?include=events`

api/api.go
lornasong: Yea, I debated putting it in the `New` for a little bit. I'll go for it!

api/task_status.go
lornasong: I like this idea to have it align with the API endpoint. It does sort of breakaway from the other logs which seems to be approximately `<package-name>.<file/object name>`. It could be confusing. Maybe `(api./status/tasks)`? The slashes might make it more clear that it's not a hierarchy and visually look like the api endpoint? Let me know if you have other ideas

api/task_status.go
lornasong: 👍

lornasong:
@findkim, thanks for the review so far! Good idea on including the endpoints + payload. I added examples to the description.

api/api.go
```diff
@@ -0,0 +1,93 @@
...
+		if err := api.srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			// go ahead and stop cts from continuing
+			log.Fatalf("error starting api server at '%d': '%s'\n", api.port, err)
+		}
+	}()
+
+	log.Printf("[INFO] (api) server started at port %d", api.port)
+
+	ctxShutDown, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
```
eikenb, line 71: With the only thing under here a goroutine, wouldn't this defer be in a race with the goroutine's use of it?

eikenb: The deferred cancel that is.

api/api.go
```diff
@@ -0,0 +1,93 @@
...
+			}
+		}
+	}()
+}
+
+// jsonResponse adds the return response for handlers
+func jsonResponse(w http.ResponseWriter, code int, response interface{}) {
+	w.Header().Add("Content-Type", "application/json")
+	w.WriteHeader(code)
+	json.NewEncoder(w).Encode(response)
```
eikenb, line 92: As all responses are supplied internally and can be sure to be encode-able w/o error, I don't see a problem with ignoring the error here as long as all the calls using this have their json encoding tested.

api/api.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/hashicorp/consul-terraform-sync/event"
+)
+
+const (
+	defaultAPIVersion = "v1"
+
+	statusHealthy      = "healthy"
+	statusDegraded     = "degraded"
+	statusCritical     = "critical"
+	statusUndetermined = "undetermined"
```
findkim, line 20: With status being a public API, it would be friendly for us to make these public constants in the event someone creates a client to consume this API, and can you add human readable descriptions on what each status means based on the rolling window of task statuses?

findkim (commented):
The API is shaping up nicely and the tests are 💯. Once the go routine in Serve is improved and John's attention to the racing `defer cancel()` this should be on its way 🎉

api/api.go
lornasong: I see what you mean about the two cancels racing. I'm new to working directly with go routines / context 🤔 I thought it would make the most sense to remove the cancel in the go routine. Let me know what you think!

api/api.go
lornasong: Thanks for pointing this out! I went with writing a test to test the json encoding. It ended up being more code overall but kept the handler code tidier 🤔.

api/api_test.go
```diff
@@ -0,0 +1,65 @@
...
+		{
+			"task status: single",
+			"status/tasks/task_b",
+			http.StatusOK,
+		},
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer ctx.Done()
+	defer cancel()
```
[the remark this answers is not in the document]
lornasong: Ok, I see! I was thinking it would increase the coverage (it doesn't).

api/api.go
lornasong: Good point about adding a client! Let me add this.

api/task_status.go
lornasong: Agree with not mixing delimiters! All right, I'll leave it as is for now. We've talked about switching to structured logs - we can revisit this when we do that. Let me know if your thoughts change.

api/api.go
findkim: Documentation 🥇

api/api.go
eikenb: When thinking about where to put it you first think of where/how you're using the context paired with it. You want the cancel to pair up with the end of the context's life. With it where you currently have the cancel defer, it will cancel the context to soon, possibly even before it gets used, as it will cancel when it hits the end of the function and will not wait for that goroutine to get started (even if it did, it'd still cancel it way before that time out is reachec).

I think you want the `cancel()` only in the goroutine. Either in  a defer, or just before the return call. That will cancel the context as it's use ends. In fact you could move creating the context into the goroutine, just before the for loop... I think that is probably what I'd do. Then you can just create and cancel in a defer together as you have it, which is the more common pattern.

lornasong: Thanks for your guidance here and taking the time to explain this! That makes sense. I updated to create the context and cancel within the go-routine (see last commit).

api/api.go
lornasong: Thanks for pairing on this @findkim and talking to me about go-routines and context! I've made a pass at this as we talked about and tested that it works for a request that takes a long time to handle. Let me know what you think!

api/api.go
```diff
@@ -0,0 +1,127 @@
...
+func (api *API) Serve(ctx context.Context) error {
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	go func() {
+		defer wg.Done()
+		for {
+			select {
+			case <-ctx.Done():
+				ctxShutDown, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
eikenb, line 100: I'd think you'd want to pass the context object `ctx` (the aurgment) instead of `context.Background()` to `context.WithTimout()` so that if the passed in context was canceled, the wrapped one would be as well. Generally this is the pattern, that as  you add new context at deeper levels that you always wrap the parent context. Used this way contexts keep the hierarchical aspect, so when you cancel one you know that all contexts under it also get canceled.

api/api.go
eikenb: This looks good. Made one new related comment below.

api/api.go
lornasong: I had tried passing in context i.e. `context.WithTimeout(ctx, 5*time.Second)` and noticed that when I stopped cts, cts would immediately shut down the server with a 'context canceled' and would not wait for open requests to finish. Looking in `Shutdown()` it returns as soon as it receives `<-ctx.Done()`. I think wrapping the parent context makes a lot of sense but isn't respecting the timeout as I thought it would. I might be missing something - let me know!

cli.go
```diff
@@ -246,13 +279,19 @@ func (cli *CLI) Run(args []string) int {
...
+			counter := 0
+			for {
+				select {
+				case <-exitCh:
+					counter++
+					log.Printf("[INFO] (cli) graceful shutdown api / controller")
+					if counter >= exitBufLen {
+						return ExitCodeOK
+					}
+				case <-time.After(10 * time.Second):
```
findkim, line 291: Reading the channel 2x is a good way make sure the API and the controller Runs have exited. However, I think the added for-loop now resets the 10 second timer. Is this intentional?

api/api.go
findkim: I believe your findings are correct Lorna. The argument `ctx` is already closed and so the `Shutdown` functionality wouldn't proceed if taking that closed `ctx`.

Waiting on the `exitCh` 2x in `cli.go` accounts for this to wait on the server shutdown + timeout.

cli.go
```diff
@@ -246,13 +279,19 @@ func (cli *CLI) Run(args []string) int {
...
-				return ExitCodeOK
-			case <-time.After(10 * time.Second):
-				log.Printf("[INFO] (cli) graceful shutdown timed out, exiting")
-				return ExitCodeInterrupt
+			counter := 0
+			for {
+				select {
+				case <-exitCh:
+					counter++
+					log.Printf("[INFO] (cli) graceful shutdown api / controller")
```
findkim, line 287: This is a bit confusing to see 2x since it's not clear on the order of which component shut down API or controller. I think it's more appropriate to log that the API has shutdown within this part of the code https://github.com/hashicorp/consul-terraform-sync/blob/7289c1e6ed0ad22f0b794258863b20b3eb50f7d9/api/api.go#L103-L106

That being said, I need to add a log that the controller shut down here 😛  https://github.com/hashicorp/consul-terraform-sync/blob/7289c1e6ed0ad22f0b794258863b20b3eb50f7d9/cli.go#L262
https://github.com/hashicorp/consul-terraform-sync/blob/7289c1e6ed0ad22f0b794258863b20b3eb50f7d9/cli.go#L245

api/api.go
findkim: Nice. The rework of the go-routine handling looks good!

cli.go
lornasong: Yea, I can see how seeing this log twice would be confusing. I'll go ahead and add the logs you mentioned. I can also update this graceful shutdown log to be like "[INFO] (cli) graceful shutdown (1/2)" and then "[INFO] (cli) graceful shutdown (2/2)". Might still be weird to see though. What do you think?

cli.go
lornasong: Not intentional! I wondered if it was reseting but on testing it "looked fine" which I shouldn't have trusted. Let me fix this. Thanks for pointing this out!

cli.go
findkim: yeah a bit weird, could be perceived as shutdown attempts as opposed to different components shutting down.

lornasong: Good point about being misconstrued as shutdown attempts. I moved the graceful shutdown log so that it logs once (on the final exit) and log in the api. I ended up not adding the controller logs in cli since there are shutdown logs in the controller code themselves. It looked a little redundant when I tried it.

cli.go
lornasong: I updated this so that the timer would decrement on the loop so that it wouldn't reset to 10 seconds.

cli.go
findkim: Ohh right. Sorry about that!

cli.go
findkim: Works for me, thanks for updating it

findkim (approved):
Excited to have the status API! Going to approve based on the areas I had commented on.

api/api.go
lornasong: Thanks for explanation @findkim. That makes sense! @eikenb let me know if you have any remaining thoughts on this

eikenb: Ok, this makes sense to me now as well. The usual pattern doesn't apply. :+1:

lornasong:
@findkim & @eikenb thanks for the review!

Issue #53, opened 2020-09-04
Specify environment to inherit for configuring Terraform providers


findkim:
By default, tasks have full access to the parent process' environment. Add a config option to Consul NIA, possibly similar to Consul-Template's `env { allowlist = [] }` option to specify which environment variables to inherit.

findkim:
Rescoping this back to Terraform provider configuration and have the tasks assume context from providers used. 😅

PR #139, opened 2020-11-18
Add vault config option

findkim:
PR to add configuration option for connecting with Vault. The vault configuration is not used by the project at the moment and is setting up for upcoming PRs to support loading configuration for Terraform providers from various sources.

Related #53

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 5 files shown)


findkim:
thanks for the review! I can't take credit on writing those comments, they're largely taken from Consul-Template and trimmed down to Consul-Terraform-Sync use-case.

PR #140, opened 2020-11-19
Deprecate provider block for terraform_provider

findkim:
This is to further distinguish the syntax and features supported by Terraform provider blocks and CTS terraform_provider blocks, like supporting dynamic configuration.

image

related #53


lornasong (approved):
nice! there are a few more provider blocks around that I'm not sure if you're planning to update now or later. there are a couple in the e2e tests and I found a few scattered around when searching `provider "`

findkim:
Missed those, will update. Thank you @lornasong!

PR #141, opened 2020-11-20
Support api port configuration

lornasong:
- Add new `port` config at top-level
 - Update api to consume new port config
 - Update e2e test config to dynamically choose next free port
 - Fix a controller test to dynamically choose next free port

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ test-integration:
 # test-e2e runs e2e tests
 test-e2e:
 	@echo "==> Testing ${NAME} (e2e)"
-	@go test ./e2e -count=1 -timeout=60s -tags=e2e -v ./... ${TESTARGS}
+	@go test ./e2e -count=1 -timeout=80s -tags=e2e -v ./... ${TESTARGS}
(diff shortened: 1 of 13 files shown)


findkim (commented):
The code looks good. Just wanted to double check about the API, are we requiring that the consul-terraform-sync daemon will require and always bind to a port?

lornasong:
@findkim thanks for asking! Currently the daemon will always bind to a port i.e. it will always serve the api at the default or configured port. Are you thinking maybe have it be an option to serve the api or not?

findkim:
Ok! Just wanted to double check. I had considered optional API for a bit when we had imagined it to be a purely daemon process. But I think the project has already evolved quickly that it's probably the common case to have the API enabled.

findkim (approved):

lornasong:
I had also imagined that it would be common case but I think it's a good question to raise. I'll go ahead and merge for now - thank you for reviewing! But I'll bring it up for team discussion and we can add a disable/enable option later.

PR #142, opened 2020-11-20
Add overall status endpoint

lornasong:
Part of a series of changes to support status api. Adds an endpoint that gives the overall status of all the tasks that cts executes. Overall status is determined by task status information for all tasks

Example:

`GET /v1/status`

`
{"status":"healthy"}
`

### DIFF
--- a/api/overall_status_test.go
+++ b/api/overall_status_test.go
@@ -1,0 +1,116 @@
+package api
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/hashicorp/consul-terraform-sync/event"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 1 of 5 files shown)


api/overall_status_test.go
```diff
@@ -0,0 +1,116 @@
...
+		statuses []string
+		expected string
+	}{
+		{
+			"healthy: all healthy",
+			[]string{StatusHealthy, StatusHealthy, StatusHealthy, StatusHealthy, StatusHealthy},
+			StatusHealthy,
+		},
+		{
+			"degraded: mix of degraded and healthy ",
```
findkim, line 94:
```suggestion
			"degraded: mix of degraded and healthy",
```

findkim (approved):
Makes sense to me 👍

api/overall_status_test.go
lornasong: Thanks for catching!

PR #143, opened 2020-11-20
Securely configuring providers

### LINKED ISSUE #53: Specify environment to inherit for configuring Terraform providers

findkim:
Adds support to load provider configuration from env, Consul KV, and Vault using template syntax
* env: `"{{ env \"MY_ENV\" }}"`
* Consul KV: `"{{ key \"kv/path\" }}"`
* Vault secret: `"{{ with secret \"my/secret/path\" }}{{ .Data.data.foo }}{{ end }}"`

Example provider block using templates to load sensitive information
```hcl
terraform_provider "example" {
  some_path = "{{ env \"SOME_PATH\" }}/example.txt"
  sensitive_token = "{{ with secret \"my/secret/path\" }}{{ .Data.data.foo }}{{ end }}"
}
```

**Moved**
* Moves hcat interface objs (Watcher, Resolver, Template) to separate package to be imported in various places 
  * `controller/hcat.go` -> `templates/hcat.go`
* Moves `namedBlock` from `templates/tftmpl/hcl_block.go` to `templates/hcltmpl/block.go`
* Shifts converting raw HCL config map to `NamedBlock` object earlier to be used by both dynamic configs and generating Terraform config files
  * This changes the type of `[]map[string]interface{}` in various places to `[]hcltmpl.NamedBlock`

**Added**
* Adds validation around dynamic configs
  * Only supported in `terraform_provider` blocks
  * Checks for configured Vault block if Vault secret template is detected
* Parallelizes loading dynamic provider configs
* Initializes Vault client if configured

Resolves #53


findkim:
This is largely ready for review. One last item i want to include in this PR is vault integration testing for loading provider config from Vault with the template `{{ secret "mysecret" }}`

controller/controller.go
```diff
@@ -263,30 +306,28 @@ func splitProviderID(id string) (string, string) {
...
+		if alias == "" {
+			return p
+		}
 
-			a, ok := pConf["alias"].(string)
-			if ok && a == alias {
-				return *p
-			}
+		// Match by alias
+		a, ok := p.Variables["alias"]
```
eikenb, line 323: The "a" variable, I'm sure, is for "alias"... but I can't not read this without hearing it as the phrase "A-ok". :)

eikenb (approved):
I've read through this a couple times and can't find anything to suggest. I do like the hcat controller->templates refactor, that makes sense to me. Nice testing as well.
I think the dynamic templates organization is also good.

controller/controller.go
findkim: haha. not my proudest variable name but the function is short and seemed a-ok :D

PR #144, opened 2020-11-23
Task status e2e test

lornasong:
Adds an e2e test for task status endpoints. Uses a fake-handler which passes
once-mode but subsequently produce errors to create failed events. Runs cts in
development in order to use fake-handler.

Changes
 - Update fake-handler to return nil error initially and then error afterwards
 - Refactor out finding next free port into a method
 - Move events so that they’re only added once results templates are rendered
   (discovered by e2e test)
 - e2e test for basic task status and overall status  endpoint

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"net"
 	"net/http"
 	"sync"
 	"time"
@@ -133,6 +134,19 @@ func (api *API) Serve(ctx context.Context) error {
 	// wait for shutdown
 	wg.Wait()
 	return ctx.Err()
+}
+
+// FreePort finds the next free port incrementing upwards. Use for testing.
+func FreePort() (int, error) {
+	listener, err := net.Listen("tcp", ":0")
+	if err != nil {
+		return 0, err
+	}
+	port := listener.Addr().(*net.TCPAddr).Port
+	if err = listener.Close(); err != nil {
+		return 0, err
+	}
+	return port, nil
 }
 
 // jsonResponse adds the return response for handlers. Returns if json encode
--- a/controller/readwrite.go
+++ b/controller/readwrite.go
@@ -145,27 +145,28 @@ func (rw *ReadWrite) checkApply(ctx context.Context, u unit) (bool, error) {
 			taskName, err)
 	}
 
-	ev, err := event.NewEvent(taskName, &event.Config{
-		Providers: u.providers,
-		Services:  u.services,
-		Source:    u.source,
-	})
-	if err != nil {
-		return false, fmt.Errorf("error creating event for task %s: %s",
-			taskName, err)
-	}
-	defer func() {
-		ev.End(err)
-		log.Printf("[TRACE] (ctrl) adding event %s", ev.GoString())
-		err = rw.store.Add(*ev)
-	}()
-
 	// result.Complete is only `true` if the template has new data that has been
 	// completely fetched. Rendering a template for the first time may take several
 	// cycles to load all the dependencies asynchronously.
 	if result.Complete {
 		log.Printf("[DEBUG] (ctrl) change detected for task %s", taskName)
+
+		ev, err := event.NewEvent(taskName, &event.Config{
+			Providers: u.providers,
+			Services:  u.services,
+			Source:    u.source,
+		})
+		if err != nil {
+			return false, fmt.Errorf("error creating event for task %s: %s",
+				taskName, err)
+		}
 		ev.Start()
+		defer func() {
+			ev.End(err)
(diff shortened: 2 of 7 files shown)


api/api.go
```diff
@@ -135,6 +136,19 @@ func (api *API) Serve(ctx context.Context) error {
 	return ctx.Err()
 }
 
+// FreePort finds the next free port incrementing upwards. Use for testing.
+func FreePort() (int, error) {
```
eikenb, line 140: Nice refactor pulling this out.

eikenb:
> Move events so that they’re only added once results templates are rendered (discovered by e2e test)

This is a small change in a small PR, so not a blocker but my instinct is to have this in a separate commit as it is unrelated to the rest as a bugfix vs. and e2e enhancement.

eikenb (approved):
Aside from the one thought, for which nothing is required, this looks good.

lornasong:
Thanks for the review @eikenb. I think that calling out the bugfix in its own commit is a great idea and makes a lot of sense. I made pull it out into its own separate commit. Thanks!

e2e/api_test.go
```diff
@@ -0,0 +1,207 @@
...
+	if err := cmd.Wait(); err != nil && err != sigintErr {
+		return err
+	}
+	return nil
+}
+
+// registerService is a helper function to regsiter a service to the Consul
+// Catalog. The Consul sdk/testutil package currently does not support a method
+// to register multiple service instances, distinguished by their IDs.
+func registerService(t *testing.T, srv *testutil.TestServer, s testutil.TestService, health string) {
```
findkim, line 193: This function is now duplicated in two packages, `e2e` and `templates/tftmpl`. It would be better to only use one instance of it in case it gets changed or removed to use the Consul testutil when supported, perhaps the `clients` package?

controller/readwrite.go
```diff
@@ -145,27 +145,28 @@ func (rw *ReadWrite) checkApply(ctx context.Context, u unit) (bool, error) {
...
+			Services:  u.services,
+			Source:    u.source,
+		})
+		if err != nil {
+			return false, fmt.Errorf("error creating event for task %s: %s",
+				taskName, err)
+		}
 		ev.Start()
+		defer func() {
+			ev.End(err)
```
findkim, line 165: Can you change the variable name of `err` to be more identifiable? This seems to not  include `resolver.Run()` which also returns `err` but the variable is reinitialized from `NewEvent`.

Should errors from fetching template dependencies be included in event tracking?

controller/readwrite.go
```diff
@@ -145,27 +145,28 @@ func (rw *ReadWrite) checkApply(ctx context.Context, u unit) (bool, error) {
...
+			log.Printf("[TRACE] (ctrl) adding event %s", ev.GoString())
+			err = rw.store.Add(*ev)
```
findkim, line 167: Since this error is set, was wondering if this error is supposed to go anywhere or supposed to be ignored? If the latter, probably best to omit the error return to make it more clear this error is ignored.

lornasong: Ack! Thank you for catching. I meant to log this error. I'll open a new PR to make this change (and others).

controller/readwrite.go
lornasong: I like the idea of naming the stored error something different! I'll do that.

Re errors from fetching template dependencies: 🤔 I wasn't considering them for event tracking since there's multiple fetchings per task-execution, and I was worried that it would skew the status data. Thinking about this again, to your point, I think that it would be good to include the errors but not the successes. A skew in successful events might wash out failed task executions but a skew in failed events isn't as problematic (since the task can't even execute). Thanks for asking! Let me make this change and let me know if you have thoughts.

e2e/api_test.go
lornasong: Agree that it's duplicated in both packages and maintenance would have to happen in two places. I had tried reusing the method but didn't like how it looked and came from a place that prefers copying until something is used 3x so left it be... Moving it to a separate package sounds it would look better - let me try it.

lornasong: Moving to the client package looked nice but it unfortunately caused an import cycle. I forgot I had tried keeping the function in only e2e or templates/tftmpl but it led to an undefined error since both tests have separate build tags. Another option is to have the function in its own package. For me, it seems a little bit much for one function so I lean towards duplicating the function. Let me know if you have thoughts!

findkim: That's unfortunate, thanks for looking into it! Yeah we can keep it as-is for now, until we have more test stragglers to group up.

lornasong: Of course! Agree - when we have some more cross-package test functions, we'll re-strategize. Thanks for the feedback!

PR #147, opened 2020-11-30
Support `?status=<health_status>` parameter for task status

lornasong:
Update task status endpoint to support a `?status=<health_status>` parameter
where `<health_status>` values are: healthy, degraded, critical, undetermined. When the
parameter is included, the task statuses will be filtered to only include those
which have a status that matches the status query parameter value.

Example:

`GET /v1/status/tasks?status=critical`

```
{
  "task_b": {
    "task_name": "task_b",
    "status": "critical",
    "providers": [
      "local"
    ],
    "services": [
      "api",
    ],
    "events_url": "/v1/status/tasks/task_b?include=events"
  },
}
```

### DIFF
--- a/api/task_status.go
+++ b/api/task_status.go
@@ -49,6 +49,15 @@ func (h *taskStatusHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	filter, err := statusFilter(r)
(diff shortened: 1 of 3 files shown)


eikenb (approved):
LGTM. I really like that you included e2e tests.

PR #148, opened 2020-11-30
Store event when fetching template dependencies errors

lornasong:
Addresses feedback in https://github.com/hashicorp/consul-terraform-sync/pull/144

Additional improvements:
 - Log error on store.Add()
 - Rename errors that will be stored in events so that they can be more easily
identified


eikenb (approved):
LGTM

PR #152, opened 2020-12-08
Flakey api Serve() test returning connection refused

lornasong:
This is likely caused by the test making http requests before the api server is
ready. I was able to reproduce this error message locally by adding a sleep in
Serve(). Logs show no additional error messages i.e. the server looks like it
started successfully and had not stopped

Change: on receiving a connection refused error message, wait and retry once


eikenb (approved):
Seems a fine workaround for slow startup.

PR #153, opened 2020-12-08
Allow a range of events in status e2e test

lornasong:
There is a race condition when waiting after a service is registered to collect
a number of events per task. Previously we were expecting a specific number of
events. However, waiting too short/long led to too few/many events collected per
task, which can cause the test to fail. This change is pre-emptively adding some
flexibility to this test by allowing a range of expected events.


eikenb (approved):
I like that you got rid of the structures in the test cases and the checkEvents is a nice refactor. I don't see anything off. :+1:

PR #156, opened 2020-12-11
use latest hashicat with tracking refactor

eikenb:
Use hashicat with large refactor of dependency tracking that simplifies
the code greatly and allows for fully asynchronous watches/triggers.


go.mod
```diff
@@ -35,6 +35,7 @@ require (
 )
 
 replace (
+	github.com/hashicorp/hcat => ../hashicat
```
findkim, line 38: Tests are failing looking for your replace directive for locally

templates/hcat.go
```diff
@@ -40,12 +40,11 @@ type Resolver interface {
 // used by this project
 // https://github.com/hashicorp/hcat
 type Watcher interface {
-	WaitCh(ctx context.Context) <-chan error
-	Add(d dep.Dependency) bool
-	Changed(tmplID string) bool
+	WaitCh(context.Context) <-chan error
+	Complete(hcat.Notifier) bool
```
findkim, line 44: Having these removed/added without other changes signals these aren't used anywhere, can they be cleaned up?

findkim: Are there follow-up changes to utilize the asynchronous template rendering/task triggering baesd on the async watches?

go.mod
eikenb: Sigh. Fixed.

findkim (approved):

templates/hcat.go
eikenb: No follow up changes for async at this time. This was focus on needed hashicat work and making sure it didn't break CTS.

eikenb: Yes `Complete()` isn't needed for CTS and should be removed from that Watcher interface. I'll remove that and regenerate the mocks and update the PR.

eikenb: Complete was needed to meet the hcat.Watcherer interface, but Register wasn't needed anymore and it is now gone. Also added a note about which ones are there simply to meet the hcat.Watcherer interface.

PR #158, opened 2020-12-14
Change default port that API is served

lornasong:
- Anticipate that CTS will be run on the same node as Consul
 - Port 8501 is reserved by Consul. Other Consul reserved ports:
 https://www.consul.io/docs/install/ports#ports-table
 - Consul team recommends using port in 8500s for http api and has no
 reservations for 8558 for current and future work as of now
 - 8558 is not a known/taken/reserved port with IANA


findkim (approved):
Thorough work to identify a better port to use as default!

PR #159, opened 2020-12-15
Rename health status language

lornasong:
- 'healthy' => 'successful'
 - 'degraded' => 'errored'
 - 'undetermined' => 'unknown'

Subsequent PRs will update with the new 'calculation' of these statuses

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -16,53 +16,39 @@ import (
 const (
 	defaultAPIVersion = "v1"
 
-	// StatusHealthy is the healthy status. This is determined based on status
+	// StatusSuccessful is the successful status. This is determined based on status
 	// type.
 	//
 	// Task Status: Determined by the success of a task updating. The 5 most
-	// recent task updates are stored as an ‘event’ in CTS. A task is healthy
-	// when all the stored events are successful.
-	//
-	// Overall Status: Determined by the health across all task statuses.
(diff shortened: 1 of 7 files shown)


api/api.go
```diff
@@ -16,53 +16,39 @@ import (
...
-	// StatusHealthy is the healthy status. This is determined based on status
+	// StatusSuccessful is the successful status. This is determined based on status
 	// type.
 	//
 	// Task Status: Determined by the success of a task updating. The 5 most
-	// recent task updates are stored as an ‘event’ in CTS. A task is healthy
-	// when all the stored events are successful.
-	//
-	// Overall Status: Determined by the health across all task statuses.
-	// Overall status is healthy when all task statuses are healthy.
```
lornasong, line 27: I removed Overall Status from these comments since overall status is returning a count of task-status rather than it's own interpretation on the health status values

findkim (approved):

api/overall_status_test.go
```diff
@@ -86,24 +86,27 @@ func TestOverallStatus_TaskStatusToOverall(t *testing.T) {
 		expected string
 	}{
 		{
-			"healthy: all healthy",
```
lornasong, line 89: These comments aren't accurate anymore. I'll likely remove this test all together so just doing the renaming for now

PR #161, opened 2020-12-15
Update Overall Status to return status count summary

lornasong:
Rather than returning a status value to represent the overall status across all
tasks, return a field 'task_summary' that shows how many tasks have each health
status value.

Example: `curl localhost:8558/status`

```
{
	"task_summary": {
		"successful":8,
		"errored":2,
		"critical":1
	}
}
```

Note: currently didn't add "unknown" status in the summary since that would require retrieving tasks that we don't have event information for yet. Doable but omitted in this PR. Open to feedback on adding it before release!

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -207,7 +207,13 @@ func TestJsonResponse(t *testing.T) {
 		{
 			"overall status: success",
 			http.StatusOK,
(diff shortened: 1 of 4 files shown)


findkim (approved):
Changes were easy to follow, 🚢 🎉

PR #162, opened 2021-01-06
Add client for CTS api

lornasong:
Add a client to call CTS’s api. This will be used in subsequent PRs to support
CLI commands that will need to make api calls. The client is loosely modeled
after Consul’s api client with simplifications.

[Example of Consul's /catalog/register API](https://github.com/hashicorp/consul/blob/master/api/catalog.go#L117)

Implements calling the two existing CTS apis: overall status + task status


lornasong:
Note: more abstractions can be done to support other types of request e.g. request bodies. I decided to hold off until we have a need for them.

eikenb (approved):
All new code and everything looked cool. Nice test setup with good coverage.

Be mindful with writing APIs as they are really easy to grow and almost impossible to shrink.

eikenb:
That last bit was just meant as some general advice when working on APIs. Even small ones like this, people will use and depend on anything you put in it...

lornasong:
Thanks for the review @eikenb, yes! agree - once it's out there... it's hard to take back. Please feel free to push on anything you think shouldn't go out there.

Issue #146, opened 2020-11-29
unable to use many services


chen23:
### Describe the bug
When configuring 100 services the client does not work reliably
### Versions
**Consul Terraform Sync**
```
 0.1.0-techpreview1 (2862363)
Compatible with Terraform ~>0.13.0
```

**Consul Version**
```
Consul 1.8.6
```

**Terraform Version**
```
Terraform v0.13.5
```

### Configuration File(s)
_Reminder to redact any sensitive information that may be present in this file_

<details>
<summary>Click to toggle contents of config file </summary>

```hcl
driver "terraform" {
  log = true
}
consul {
  address = "10.1.20.51:8500"
}
log_level = "trace"
task {
  name = "ManyApps"
  description = "Testing many apps"
  source = "../../manyapps"
  services = ["app001","app002","app003","app004","app005","app006","app007","app008","app009","app010","app011","app012","app013","app014","app015","app016","app017","app018","app019","app020","app021","app022","app023","app024","app025","app026","app027","app028","app029","app030","app031","app032","app033","app034","app035","app036","app037","app038","app039","app040","app041","app042","app043","app044","app045","app046","app047","app048","app049","app050","app051","app052","app053","app054","app055","app056","app057","app058","app059","app060","app061","app062","app063","app064","app065","app066","app067","app068","app069","app070","app071","app072","app073","app074","app075","app076","app077","app078","app079","app080","app081","app082","app083","app084","app085","app086","app087","app088","app089","app090","app091","app092","app093","app094","app095","app096","app097","app098","app099","app100"]
}
```
</details>

### Terraform Configuration Files Generated by Consul-Terraform-Sync
_Reminder to redact any sensitive information that may be present in the files_

<details>
<summary>Click to toggle contents of main.tf </summary>

```terraform
locals {

  # Create a map of service names to instance IDs
  service_ids = transpose({
    for id, s in var.services : id => [s.name]
  })

  # Group service instances by name
  grouped = { for name, ids in local.service_ids :
    name => [
      for id in ids : var.services[id]
    ]
  }

}

resource "local_file" "appoutput" {
  for_each = local.grouped
  content = jsonencode(each.value)
  filename = "/tmp/${each.key}.json"
}
```
</details>

<details>
<summary>Click to toggle contents of sample service </summary>

```json
[
  {
    "Node": {
      "ID": "",
      "Node": "192.168.128.1",
      "Address": "192.168.128.1",
      "Datacenter": "dc1",
[...]

PR #164, opened 2021-01-08
Update default transport configurations

findkim:
`http.MaxIdleConnsPerHost` was default to a very low value that caused TCP connections to not be reused across hcat dependency monitoring. Only ~1 connections were being reused from the initial set of hcat requests to Consul, and the remaining dependencies required establishing a new connection.

Changes:
* `DefaultMaxIdleConnsPerHost`: CTS default changed from 1 to 100 (the [transport default is 2](https://golang.org/pkg/net/http/#Transport) and [setting -1 disables keep alives altogether](https://golang.org/src/net/http/transport.go?s=3396:10951#L878)). This increases the # of reusable TCP connections for service monitoring. From 1:2 connections per service now to 1:1.
* `DefaultMaxIdleConns` default changed from 100 to 0, which disables the the limit. The previous limit not only caps all the requests to Consul but also requests to other hosts, which is not needed at this time.
* `DefaultIdleConnTimeout`: from 90s to 5s. hcat monitoring of dependencies should theoretically have 1 connection per 1 dependency for blocking queries. Any idle connections not immediately used would likely never be reused, so close it out earlier.

~For example, 100 services required about 200 connections at startup (normal request and then blocking query).~

~This is the primary bug in #146 when monitoring 100 services always errored with EOF/429 when the Consul agent has a default [`http_max_conns_per_client=200`](https://www.consul.io/docs/agent/options#http_max_conns_per_client). This should now allow up to 198 services by default without resulting in connection limits.~

See [comment below](https://github.com/hashicorp/consul-terraform-sync/pull/164#issuecomment-758195773) for updates on ^

Operators can increase `consul.transport.max_idle_conns_per_host` value to reduce the delay between detecting updates and task execution when monitoring 200 services or more.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -51,7 +51,7 @@ var (
 				DialTimeout:         TimeDuration(10 * time.Second),
 				DisableKeepAlives:   Bool(false),
 				IdleConnTimeout:     TimeDuration(1 * time.Minute),
-				MaxIdleConnsPerHost: Int(5),
+				MaxIdleConnsPerHost: Int(100),
 				TLSHandshakeTimeout: TimeDuration(10 * time.Second),
 			},
 		},
@@ -266,7 +266,7 @@ func TestConfig_Finalize(t *testing.T) {
 	expected.BufferPeriod.Enabled = Bool(true)
 	expected.Consul.KVNamespace = String("")
 	expected.Consul.TLS.Cert = String("")
-	expected.Consul.Transport.MaxIdleConns = Int(100)
+	expected.Consul.Transport.MaxIdleConns = Int(0)
 	expected.Vault = DefaultVaultConfig()
 	expected.Vault.Finalize()
 	expected.Driver.consul = expected.Consul
--- a/config/testdata/long.hcl
+++ b/config/testdata/long.hcl
@@ -33,7 +33,7 @@ consul {
     dial_timeout = "10s"
     disable_keep_alives = false
     idle_conn_timeout = "1m"
-    max_idle_conns_per_host = 5
+    max_idle_conns_per_host = 100
     tls_handshake_timeout = "10s"
   }
 }
--- a/config/testdata/long.json
+++ b/config/testdata/long.json
@@ -31,7 +31,7 @@
       "dial_timeout": "10s",
       "disable_keep_alives": false,
       "idle_conn_timeout": "1m",
-      "max_idle_conns_per_host": 5,
+      "max_idle_conns_per_host": 100,
       "tls_handshake_timeout": "10s"
     }
   },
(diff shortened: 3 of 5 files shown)


findkim:
TODO: write an e2e test with a running Consul agent for http_max_conns_per_client (Consul) + max_idle_conns_per_host (CTS)

findkim:
Update: I had a bug within my on repro steps, so this does not address why 100 services caused 200 connections. It was my overlook of curl also contributing to the limit. Once I had waited for the curl connections to close out, I ran CTS with 100 services and it successfully executed prior to these changes.

These changes do however improve the TCP connection reusability of hcat/consul client. This is not something I was able to write an e2e test for since the transport and dialing code is housed in hcat.

Inserting code around the dialing provided insight on how many times a connection was established ([here](https://github.com/hashicorp/hcat/blob/master/internal/dependency/client_set.go#L271))
```diff
+var dialCount uint64

 func newTransport(i *CreateClientInput) (*http.Transport, error) {
        // This transport will attempt to keep connections open to the server.
+       dialer := net.Dialer{
+               Timeout:   i.TransportDialTimeout,
+               KeepAlive: i.TransportDialKeepAlive,
+       }
        transport := &http.Transport{
                Proxy: http.ProxyFromEnvironment,
-               Dial: (&net.Dialer{
-                       Timeout:   i.TransportDialTimeout,
-                       KeepAlive: i.TransportDialKeepAlive,
-               }).Dial,
+               Dial:  dialer.Dial,
+               DialContext: func(ctx context.Context, network, address string) (net.Conn, error) {
+                       atomic.AddUint64(&dialCount, 1)
+                       log.Printf("[INFO] Dialing %s %s: %d", network, address, atomic.LoadUint64(&dialCount))
+                       return dialer.DialContext(ctx, network, address)
+               },
                DisableKeepAlives:   i.TransportDisableKeepAlives,
                MaxIdleConns:        i.TransportMaxIdleConns,
                IdleConnTimeout:     i.TransportIdleConnTimeout,
                MaxIdleConnsPerHost: i.TransportMaxIdleConnsPerHost,
                TLSHandshakeTimeout: i.TransportTLSHandshakeTimeout,
        }
```

Prior to changes: 100 services resulted in 190 dials.
After changes: 100 services resulted in 100 dials.

eikenb (approved):
LGTM

One related question.. would you suggest changes to hcat's default settings in relation to this? Seems like the defaults would be improved if changed to reflect your findings.

findkim:
Thanks for the review @eikenb!

I don't recall hcat having default options and relies on the callers' defaults. But yeah adding default options I think would be valuable. I would imagine this is affecting Consul Template as well (that's where I'm guilty of copy pasta'ing the config defaults from)

PR #165, opened 2021-01-11
Fix Consul KV path appending terraform suffix 2x

findkim:
Updates code comments to include more context on end result of the KV path for stored state files


lornasong (approved):
nice! tests were also helpful for checking my understanding 🙏

Issue #150, opened 2020-12-03
Trigger NIA tasks based on Consul key/value (KV) changes


sandstrom:
#### Description
It would be great if we could trigger tasks based on specific keys in the KV/store. Right now, it only seems to trigger based on changes to a service.

https://www.consul.io/docs/nia/installation/configuration#task
https://www.consul.io/docs/nia/installation/configuration

#### Use Cases
We have two groups of nodes, "green" and "blue". To avoid downtime we shift traffic from one to the other when doing deploys. This shift is handled by our load balancer (AWS ALB) using their rules.

We'd like to use a Consul KV pair `load-balancer/active-group` : `green` to trigger a Network Infrastructure Automation task that would update the load balancer rules, via Terraform, to shift traffic to green.

#### Alternative Solutions
While services have meta data, it seems that metadata is only "inherited" from service definitions on individual nodes. 

If we want to shift traffic from our blue group to our green group of nodes, we want one canonical place where that value is stored (group receiving traffic). It wouldn't make sense to make service definition changes on an individual node.

AFAIK we can't set service-level meta directly in the catalog, for that service globally.

findkim:
@sandstrom thanks for taking the time submit this feature request. And I see 👍 are trickling in 🎉.

Yep you're correct in that tasks are triggered only on services right now, and agreed I believe that the catalog API is limited for updating the service-level metadata after it has been deployed.

Thanks for laying out the use case well. I think this feature request is valuable and am interested in more insight on how the KV pair would change. For your use case (or other ppl interested in this feature!) would the KV pair `load-balancer/active-group`:`green` be updated by a manual or automated process? What does that look like?

sandstrom:
@findkim Thanks for thinking about this!

It could be both automated and manual. In our case probably manual initially, but could be an automated script kicked off by e.g. a new tagged release in Github in the future.

Consul's KV store is quite useful for deployment and infrastructure state, stuff that shouldn't go in your primary database. Things like:

- maintenance mode (true/false)
- currently deployed blue/green group (string value)
- rate limiting thresholds
- rate of requests/traffic sent to canary channel
- etc

Being able to use NIA for KV changes would unlock a lot of possibilities.

findkim:
Awesome, thank for more info on your use case! This is a feature request we anticipated and have been collecting different use cases to help us design and prioritize a KV trigger.

sandstrom:
@findkim Sounds great, let me know how it goes!

[...]

PR #166, opened 2021-01-12
User defined metadata for services

findkim:
Adds a new option `cts_user_defined_meta` to the `service` configuration block for Consul Terraform Sync. This attribute is useful to insert metadata for a service to be used for network infrastructure automation and consumed by a Terraform module.

A driving use case for user-defined metadata within CTS is automating load balancers and associating a service with a virtual IP address. The current method is to insert the VIP into Consul service metadata during registration. However, it might not be desirable to insert NIA scoped metadata into Consul, where it may not be useful by applications aside from CTS. It would also require shared context of service to VIP mappings between a network administrator and the application developer. `cts_user_defined_meta` can now be used to keep the scope of context to CTS and controlled by the CTS operator.

```hcl
service {
  name = "app"
  cts_user_defined_meta = {
    some_key = "value"
  }
}
```

## Summary of changes
* Adds config option `service.cts_user_defined_meta`
* Prepares user-defined metadata for the list of services by task during initialization. This list include service names or CTS service IDs. Passes the map of metadata that's scoped by task to the task's corresponding template functions.
* When the `terraform.tfvars.tmpl` is rendered, it inserts the user-defined metadata into the `var.services` to be available for compatible Terraform modules.
  * This new attribute to the `var.services` is a non-breaking change for modules. If modules choose to utilize the `cts_user_defined_meta` field, module authors can update their module `var.services` to include it within their variable definintion.

### DIFF
--- a/config/service.go
+++ b/config/service.go
@@ -276,6 +276,41 @@ func (c *ServiceConfigs) Validate() error {
 	return nil
 }
 
+// CTSUserDefinedMeta generates a map of service name to user defined metadata
+// from a list of service IDs or service names.
+func (c *ServiceConfigs) CTSUserDefinedMeta(serviceList []string) map[string]map[string]string {
+	if c == nil {
+		return nil
+	}
+
+	services := make(map[string]bool)
+	for _, s := range serviceList {
+		services[s] = true
+	}
+
+	m := make(map[string]map[string]string)
+	for _, s := range *c {
+		if len(s.CTSUserDefinedMeta) == 0 {
+			continue
+		}
+
+		serviceName := *s.Name
+		if s.ID != nil && *s.ID != "" {
+			if _, ok := services[*s.ID]; ok {
+				m[serviceName] = s.CTSUserDefinedMeta
+				continue
+			}
+		}
+
+		if _, ok := services[serviceName]; ok {
+			if s.ID == nil || *s.ID == "" {
+				m[serviceName] = s.CTSUserDefinedMeta
+			}
+		}
+	}
+	return m
+}
+
 // GoString defines the printable version of this struct.
 func (c *ServiceConfigs) GoString() string {
 	if c == nil {
--- a/config/service.go
+++ b/config/service.go
@@ -295,7 +295,7 @@ func (c *ServiceConfigs) CTSUserDefinedMeta(serviceList []string) map[string]map
 		}
 
 		serviceName := *s.Name
-		if s.ID != nil && *s.ID != "" {
+		if StringPresent(s.ID) {
 			if _, ok := services[*s.ID]; ok {
 				m[serviceName] = s.CTSUserDefinedMeta
 				continue
@@ -303,7 +303,7 @@ func (c *ServiceConfigs) CTSUserDefinedMeta(serviceList []string) map[string]map
 		}
 
 		if _, ok := services[serviceName]; ok {
-			if s.ID == nil || *s.ID == "" {
+			if !StringPresent(s.ID) {
 				m[serviceName] = s.CTSUserDefinedMeta
 			}
 		}
--- a/config/service.go
+++ b/config/service.go
@@ -29,6 +29,10 @@ type ServiceConfig struct {
 
 	// Tag is used to filter nodes based on the tag for the service.
 	Tag *string `mapstructure:"tag"`
+
+	// CTSUserDefinedMeta is metadata added to a service automated by CTS for
+	// network infrastructure automation.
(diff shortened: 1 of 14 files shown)


config/service.go
```diff
@@ -276,6 +276,41 @@ func (c *ServiceConfigs) Validate() error {
...
+	}
+
+	m := make(map[string]map[string]string)
+	for _, s := range *c {
+		if len(s.CTSUserDefinedMeta) == 0 {
+			continue
+		}
+
+		serviceName := *s.Name
+		if s.ID != nil && *s.ID != "" {
```
lornasong, line 298: I came across `StringPresent()` previously, which I think you added! I think it could be reused here and potentially line 306 below to help readability.

config/service_test.go
```diff
@@ -271,3 +271,124 @@ func TestServiceConfig_Validate(t *testing.T) {
 		})
 	}
 }
+
+func TestServiceConfigs_CTSUserDefinedMeta(t *testing.T) {
```
lornasong, line 275: These test cases were helpful to understand CTSUserDefinedMeta! I was trying to make sure I understood it and tested out:
```
{
	"service list value matches name or id",
	&ServiceConfigs{{
		Name: String("a"),
		CTSUserDefinedMeta: map[string]string{
			"a": "b",
		},
	}, {
		Name: String("a-with-meta"),
		ID:   String("a"),
		CTSUserDefinedMeta: map[string]string{
			"key": "value",
		},
	}},
	[]string{"a"},
	map[string]map[string]string{
		"a":  {"a": "b"},
	},
}
```
And got the result:
```
expected: map[string]map[string]string{"a":map[string]string{"a":"b"}}
actual  : map[string]map[string]string{"a":map[string]string{"a":"b"}, "a-with-meta":map[string]string{"key":"value"}}
```

It seems like this isn't expected. I was wondering if this test case could actually happen? Let me know if I'm misunderstanding!

lornasong (commented):
The way you split the changes by commits makes reviewing really nice! I left a question to check my understanding of `CTSUserDefinedMeta()`. Also left a minor suggestion that is up to you if you like.

config/service_test.go
findkim: Good edge case! The result is not expected; however, `ServiceConfigs.Validate()` should catch that and error before proceeding to `ServiceConfigs.CTSUserDefinedMeta()`. This case would be considered an invalid set of services because there's a conflict in uniqueness between a service ID and service name.

I added another set of tests to make it a bit more clear, thanks for pointing this out!

config/service.go
findkim: I love it, thanks for reminding me about `StringPresent`

config/service_test.go
lornasong: Ah, I see! Thanks for the tests and context on how this case is invalid.

lornasong (approved):
yay user defined meta ✨✨ thanks for checking my understanding!

sandstrom:
How would this work in practice?

Is this something that only live inside CTS, in the consul KV store, or in the Consul catalog?

If we have a service called e.g. 'app' and it's running via one load-balancer and two target groups (blue/green). Would this allow us to call one of our consul servers, tell it the current target group, and have CTS automatically run a terraform task updating the target load balancer group?

devarshishah3:
@sandstrom. The user defined meta for services will apply to all the service instances within the service. Based on your comments it looks like you want to update the node with metadata of whether it is a part of blue deployment or green deployment. Based on that you want services running on that node to be part of blue or green deployment. I am still not very clear on your ask. Would you have some time to chat about your use case with us so we can better address it in CTS.

sandstrom:
Our setup looks like this:

image

What we would like to do, is to tell Consul which target group is the active one (blue or green), and have CTS + Terraform make a change in the AWS load balance rule.

I originally envisioned that this would be in the Consul KV store, a key could be e.g. `network.active_group.app` with the value `green` or `blue`.

findkim:
hi @sandstrom excited to hear about your interest on a feature hot off the press!

User-defined metadata is only in scope of CTS and is independent from Consul KV and catalog. It's intended to offer a way for operators to insert service-level metadata without updating Consul or current upstream processes.

Thinking through your use-case I don't think it could be easily solved using the user-defined meta, but it might work in a clunky way. This feature is initially static by nature through CTS configuration, and so the flipping of the `active_group` from between green and blue using this feature would require running CTS with changed configuration (since restarting is the only means currently to reload configs). 

👍 for the task triggering discussion over on #150

PR #167, opened 2021-01-13
Fix indefinite hanging when tasks have identical providers and services

findkim:
# Bug
Template content needs to be distinct from another for hcat to identify them as different templates to be rendered with values, and then for CTS to execute Terraform commands. When multiple tasks have the same set of services and providers (or no providers), CTS hangs indefinitely waiting for tasks to execute the first time. This is because those tasks will have identical generated template files `terraform.tfvars.tmpl` and is treated as one by hcat. The other identical templates never render which causes CTS to hang since that task never completes. This bug is present in all modes.

## Repro steps
1. Config file with 2 tasks with same services (no provider needed)
```hcl
log_level = "DEBUG"

driver "terraform" {
  log = true
}

task {
  name      = "a"
  source    = "../../e2e/test_modules/e2e_basic_task/"
  services = ["app"]
}

task {
  name     = "b"
  source   = "../../e2e/test_modules/e2e_basic_task/"
  services = ["app"]
}
```

2. Run Consul Terraform Sync against Consul agent in dev mode
3. Consul Terraform Sync hangs (you can exit the process now)
4. View the `sync-tasks` directory for the tasks "a" and "b", both have identical `terraform.tfvars.tmpl` files but only one has the rendered file `terraform.tfvars`

## Solution
Task name is unique across tasks. Including the task name to the template files would guarantee that each `terraform.tfvars.tmpl` file has unique content even if multiple tasks are monitoring the same set of providers and services.


lornasong (approved):
Nice find! This looks great. I was thinking it could be helpful for future people (myself included) to have your discovery documented somehow within the code as well. e.g. a comment on why  `TaskPreamble` is being added (besides being additional useful info). Not essential and up to you, so feel free to go ahead and merge.

findkim:
Great idea @lornasong, I'll write up some context before merging

PR #169, opened 2021-01-14
Remove deprecated provider block with special error handling

findkim:
`provider` block name was deprecated in [v0.1.0-techpreview2](https://github.com/hashicorp/consul-terraform-sync/blob/master/CHANGELOG.md#010-techpreview2-december-16-2020), it is removed in v0.1.0-beta.
 
image

related PR https://github.com/hashicorp/consul-terraform-sync/pull/140

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


lornasong (approved):
looks good to me!

config/config.go
```diff
@@ -476,6 +463,26 @@ func supportedFormat(format string) bool {
 	return false
 }
 
+func processUnusedConfigKeys(md mapstructure.Metadata, file string) error {
```
eikenb, line 466: Erroring on unused configuration values is a great idea.

eikenb: Is there a test for this?

findkim: The config decoder had [already supported erroring unused configuration](https://github.com/hashicorp/consul-terraform-sync/pull/169/files#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R347
) and is [tested here](https://github.com/hashicorp/consul-terraform-sync/blob/master/config/config_test.go#L136-L149)

However with the way the decoder works, if that is set to `true`, it doesn't populate the metadata to figure out which keys are unused and returns an error. Another approach could string match that error for `'provider'` for this special case handling. So I swapped it to false and manually processed the unused keys.

eikenb (approved):

PR #170, opened 2021-01-15
add auto-commit setting support to providers

eikenb:
Adds logic to support setting an 'auto-commit' boolean setting on providers that will control the out-of-band automatic commit on task run for providers than need it (eg. Panos).

Note there is no documentation. Seems like any relevant docs would live elsewhere.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -244,9 +244,11 @@ func (c *Config) Validate() error {
 		return err
 	}
 
-	// TODO: validate providers listed in tasks exist
-
 	if err := c.BufferPeriod.Validate(); err != nil {
+		return err
+	}
+
+	if err := c.validateTaskProvider(); err != nil {
 		return err
 	}
 
@@ -254,6 +256,38 @@ func (c *Config) Validate() error {
 		return err
 	}
 
+	return nil
+}
+
+// validateTaskProvider checks that task <-> provider relations are good
+func (c *Config) validateTaskProvider() error {
+	// which providers have auto_commit enabled
+	acProviders := make(map[string]bool, len(*c.TerraformProviders))
+	for _, p := range *c.TerraformProviders {
+		for name, values := range *p {
+			if a, ok := values.(map[string]interface{})["alias"]; ok {
+				name = fmt.Sprintf("%s.%s", name, a)
+			}
+			if v, ok := values.(map[string]interface{})["auto_commit"]; ok {
+				if b, ok := v.(bool); ok {
+					acProviders[name] = b
+				}
+			}
+		}
+	}
+
+	autocomUsed := make(map[string]bool)
+	for _, t := range *c.Tasks {
+		for _, pname := range t.Providers {
+			if autocom := acProviders[pname]; autocom {
+				if autocomUsed[pname] {
+					return fmt.Errorf("provider with autocommit (%s)"+
+						" cannot be used by more than one task", pname)
+				}
+				autocomUsed[pname] = true
+			}
+		}
+	}
 	return nil
 }
 
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -304,6 +304,34 @@ func TestConfig_Finalize(t *testing.T) {
 }
 
 func TestConfig_Validate(t *testing.T) {
+	valid := longConfig.Copy()
+
+	// 2 tasks using same provider w/ auto_commit enabled (should err)
+	autoCommit := valid.Copy()
+	autoCommit.TerraformProviders = &TerraformProviderConfigs{
+		&TerraformProviderConfig{"X": map[string]interface{}{"auto_commit": true}}}
+	ts := *autoCommit.Tasks
+	t2 := ts[0].Copy()
+	*t2.Name = "task2"
+	ts = append(ts, t2)
+	autoCommit.Tasks = &ts
+
+	// task configured with no providers configured (default provider)
+	noProvider := *valid.Copy()
+	noProvider.TerraformProviders = &TerraformProviderConfigs{}
+
+	// valid case with multiple tasks w/ different providers
+	validMultiTask := longConfig.Copy()
+	*validMultiTask.Tasks = append(*validMultiTask.Tasks, &TaskConfig{
+		Description: String("test task1"),
+		Name:        String("task1"),
+		Services:    []string{"serviceD"},
+		Providers:   []string{"Y"},
+		Source:      String("Z"),
+	})
+	*validMultiTask.TerraformProviders = append(*validMultiTask.TerraformProviders,
+		&TerraformProviderConfig{"Y": map[string]interface{}{}})
+
 	cases := []struct {
 		name    string
 		i       *Config
@@ -319,8 +347,20 @@ func TestConfig_Validate(t *testing.T) {
 			false,
 		}, {
 			"valid long",
-			longConfig.Copy(),
+			valid.Copy(),
 			true,
+		}, {
+			"multi-task valid",
+			validMultiTask.Copy(),
+			true,
+		}, {
+			"empty provider",
+			noProvider.Copy(),
+			true,
+		}, {
+			"autocommitting provider reuse error",
+			autoCommit.Copy(),
+			false,
 		},
 	}
 
--- a/config/task_test.go
+++ b/config/task_test.go
@@ -411,6 +411,15 @@ func TestTasksConfig_Validate(t *testing.T) {
 				},
 			},
 			false,
+		}, {
+			"duplicate provider instances",
+			[]*TaskConfig{
+				{
+					Name:      String("task"),
+					Providers: []string{"provider.A", "provider.B"},
+				},
+			},
+			false,
 		},
 	}
 
--- a/handler/panos.go
+++ b/handler/panos.go
@@ -45,6 +45,7 @@ type Panos struct {
 	providerConf pango.Client
 	adminUser    string
 	configPath   string
+	autoCommit   bool
 }
 
 // NewPanos configures and returns a new panos handler
@@ -65,6 +66,14 @@ func NewPanos(c map[string]interface{}) (*Panos, error) {
 	if val, ok := c["json_config_file"]; ok {
 		if v, ok := val.(string); ok {
 			configPath = v
+		}
+	}
+
+	// should we auto_commit?
+	var autoCommit bool
+	if val, ok := c["auto_commit"]; ok {
+		if v, ok := val.(bool); ok && v {
+			autoCommit = true
 		}
 	}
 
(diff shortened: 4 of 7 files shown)


config/task_test.go
```diff
@@ -411,6 +411,15 @@ func TestTasksConfig_Validate(t *testing.T) {
 				},
 			},
 			false,
+		}, {
+			"duplicate provider instances",
```
findkim, line 415: Thanks for adding this test!

config/config.go
```diff
@@ -250,13 +250,52 @@ func (c *Config) Validate() error {
...
+func (c *Config) validateTaskProvider() error {
+	// provider name -> auto-commit setting, the 2 things we need to check
+	providers := make(map[string]bool, len(*c.TerraformProviders))
+	for _, p := range *c.TerraformProviders {
+		for name, values := range *p {
+			if a, ok := values.(map[string]interface{})["alias"]; ok {
+				name = fmt.Sprintf("%s.%s", name, a)
+			}
+			providers[name] = false
+			if v, ok := values.(map[string]interface{})["auto-commit"]; ok {
```
findkim, line 274: Most of the Terraform provider arguments I've seen are underscore delimited. Since this would be a part of the `terraform_provider` blocks could you change it to `auto_commit` for consistency?

findkim: One caution to consider when adding a CTS-specific option to the `terraform_provider` block is the potential of overlap with an existing Terraform provider option.

For example `task_env` does have a small chance of overlap, but I don't anticipate it happening. Auto commit seems like a more common term in general which increases overlap.

findkim (commented):
The changes look good!

There is one aspect of the `terraform_provider` block that I'm not proud of as it is growing in new features. The CTS specific arguments need to be removed before writing out the Terraform files. Otherwise, it'll be an unexpected provider argument from Terraform's perspective and always error. That includes `alias` (it's used slightly different than Terraform's usage), `task_env`, and now the new auto commit option.

They're currently done in different places, can you also include logic to omit the auto-commit from the written provider block for Terraform? Feel free to refactor the others as needed if there's better organization
* `alias` is dropped right before writing to file [here](https://github.com/hashicorp/consul-terraform-sync/blob/master/templates/tftmpl/root.go#L260-L264)
* `task_env` is separated [here](https://github.com/hashicorp/consul-terraform-sync/blob/master/driver/terraform_provider.go#L33-L44)

config/config_test.go
```diff
@@ -319,8 +347,20 @@ func TestConfig_Validate(t *testing.T) {
...
+			validMutiTask.Copy(),
+			true,
+		}, {
+			"autocommitting provider reuse error",
+			autoCommit.Copy(),
+			false,
+		}, {
+			"no provider for task error",
+			noProvider.Copy(),
+			false,
```
lornasong, line 363: From my understanding, sometimes it is ok for the user to configure a provider block e.g. when the provider can work with its default values. I sometimes use the local provider in a module without declaring a provider block. Let me know if I misunderstand you here though

lornasong (commented):
I like how clean this change is! I left comments on a couple minor things

config/config.go
eikenb: I wasn't sure between auto-commit, auto_commit and autocommit... and I couldn't seem to find an example of another 2-word name so I just guessed knowing it'd come up in the review if neededd. :)

config/config_test.go
eikenb: So it sounds like it should allow the situation where TerraformProviders is empty (default) but error on the case where TerraformProviders has a value but the Task's listed provider isn't present.

Sound right?

handler/panos.go
```diff
@@ -96,14 +105,18 @@ func NewPanos(c map[string]interface{}) (*Panos, error) {
...
 		adminUser:    username,
 		configPath:   configPath,
+		autoCommit:   autoCommit,
 	}, nil
 }
 
 // Do executes panos' out-of-band Commit API and calls next handler while passing
 // on relevant errors
 func (h *Panos) Do(prevErr error) error {
 	log.Printf("[INFO] (handler.panos) commit. host '%s'", h.providerConf.Hostname)
```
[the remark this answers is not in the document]
eikenb: It still calls through to the next handler, but is a noop for it's functionality. I was thinking it'd be better to know it was called and didn't commit (maybe you had it configured to use the handler but had auto-commit off) than skip the message. 

It does seem rather verbose for very limited gain though. After thinking about it more I'm leaning toward removing it. 

We should talk about this at some point. Maybe look into coming up with a simple guideline about when and what types of log entries we want. For example, we could say we only log on actions (not on skipping them) or something.

config/config.go
eikenb: auto_commit has been committed

config/config_test.go
```diff
@@ -304,6 +304,34 @@ func TestConfig_Finalize(t *testing.T) {
...
+	*t2.Name = "task2"
+	ts = append(ts, t2)
+	autoCommit.Tasks = &ts
+
+	// task configured with provider that doesn't exist
+	noProvider := *valid.Copy()
+	noProvider.TerraformProviders = &TerraformProviderConfigs{}
+
+	// valid case with multiple tasks w/ different providers
+	validMutiTask := longConfig.Copy()
```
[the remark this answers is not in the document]
eikenb: valid :dog2:  task?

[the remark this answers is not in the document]
eikenb: fixed

config/config_test.go
lornasong: Yes, that sounds right to me! To add onto:

> error on the case where TerraformProviders has a value but the Task's listed provider isn't present

This would be if the TerraformProvider didn't exist any of the listed provider across all tasks. I'm not sure if this validation already exists or if you were planning to add it... I personally don't think that auto-commit requires this validation so ok with it not being included in this PR

handler/panos.go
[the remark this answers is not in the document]
lornasong: I found this log helpful for me when I was developing handlers. Maybe change it to be TRACE level instead? Agree, I think it would be great to have a discussion on logging.

🤔 To your point, I'm realizing that this log is pretty ambiguous. I'm understanding your interpretation of this log as something like 'entering the commit code but may or may not do actual commit'. I had interpreted it as 'starting to commit now...'. Maybe one option could be to include the autoCommit value in this log?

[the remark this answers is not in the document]
eikenb: I thought about moving it into the commit() method w/ TRACE or DEBUG (as there is a commit successful log entry at DEBUG level). But wasn't really sure it added anything to the log entries already in commit().

Having it log with the value of autoCommit or moving the log to an `else` clause so it logs when autoCommit is off might be useful.

I guess for now I'll just update it to have the autoCommit value displayed. Something along the lines of...

    "[TRACE] (handler.panos) commit enabled. host foo."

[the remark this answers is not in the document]
eikenb: I made that change. Let me know if you think of something better. Thanks.

config/config_test.go
eikenb: I was attempting to address the the code comment...

    // TODO: validate providers listed in tasks exist

.. while I was working on validating task/provider inter-dependencies. It isn't directly relevant to this ticket but seemed worth-while as I was almost testing for it already.

So it's not attempting to check that all TerraformProvider entries are used (which seems to be what you're saying in that "This would be.." sentence), but that all the Tasks have valid providers.

handler/panos.go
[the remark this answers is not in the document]
lornasong: I didn't realize there was already a successful commit log entry, sorry! In that case, I'm ok with deleting this log as well. Also happy to keep the log as you've updated though - the message is super clear. I'll leave it up to you. Thanks!

config/config_test.go
lornasong: Gotcha! Just didn't want you to feel obligated. I actually interpret that code comment as you originally had: that all task listed providers should have an associated TerraformProviders block 🤔

eikenb:
> There is one aspect of the `terraform_provider` block that I'm not proud of as it is growing in new features. The CTS specific arguments need to be removed before writing out the Terraform files. Otherwise, it'll be an unexpected provider argument from Terraform's perspective and always error. That includes `alias` (it's used slightly different than Terraform's usage), `task_env`, and now the new auto commit option.

Seems like the `root.go` (where `alias` is filtered) is the logical place for these things as that is the final rendering place. The place for filtering `task_env` seems more like it was included where the value was used, where it could have been done in root.go like alias.

For now I think I'll just add in `root.go` where it does the `alias` a filter for `auto_commit`, along with some tests for this. I'll push these changes up in a separate commit.

I think filtering `task_env` should probably be refactored to `root.go` as well but I'm less sure about it and how to test that so I'm just going to leave it for now.

config/config_test.go
eikenb: I reworked this to allow task providers w/o a matching terraform provider as long as the terraform providers is empty. I've added a test to confirm this and adjusted the other test to use a setup with terraform providers set, just not the one the Task wants.

lornasong: the noProvider and noTaskProvider tests look great to me!

For "task providers w/o a matching terraform provider as long as the terraform providers is empty" - I can see a possibility where a module uses an f5 and local provider, where f5 provider needs a TerraformProvider block but the local provider doesn't since it can use default. I think this should be valid but would not pass the validation added here. I'm personally starting to think that there isn't a generalized way to validate the the task provider list to TerraformProvider blocks. Let me know if I misunderstand in anyway.

config/config_test.go
```diff
@@ -304,6 +304,39 @@ func TestConfig_Finalize(t *testing.T) {
...
+	t2 := ts[0].Copy()
+	*t2.Name = "task2"
+	ts = append(ts, t2)
+	autoCommit.Tasks = &ts
+
+	// task configured with provider that doesn't exist
+	noProvider := *valid.Copy()
+	noProvider.TerraformProviders = &TerraformProviderConfigs{}
+
+	// task configured with provider that doesn't exist
```
lornasong, line 323: this comment doesn't quite match my understanding of this config. maybe a copy-pasta from line 319?

config/config_test.go
findkim: Oh no sorry about my comment I left in. Yeah I had initially imagined that it would be nice to validate and error early. However Lorna is right about the edge case and that it's not necessarily required to have a `terraform_provider` block. It just occurred to me the other day it could also be that they want to use the `driver.terraform { required_providers {} }` block that denotes the namespace.

So removing the added validation (and my troublesome comment) would allow for those two cases.

findkim (approved):
Great, it did seem that `root.go` would be a decent place for auto commit filtering. Approving on the items we discussed!
* `auto-commit` -> `auto_commit`
* Update to exclude `auto_commit` from being written to file

config/config_test.go
eikenb: Total c-n-p error. :facepalm:

config/config_test.go
eikenb: Great. I'll just remove the related code then. Pushing updates soon...

eikenb: I've removed the verify and test code as well as that comment.

lornasong (approved):
Looks good to me! All those tests ✨ . Thanks for all the discussion!

Issue #171, opened 2021-01-19
Nested blocks for provider definition become templated as variables


joatmon08:
### Describe the bug

When using a nested blocks for provider definition, it will be templated to variable in the provider block instead of preserving the nested structure.

### Versions
**Consul Terraform Sync**
```
 0.1.0-techpreview2 (d844653)
Compatible with Terraform >= 0.13.0, < 0.15
```

**Consul Version**
```
Consul 1.9.1
```

**Terraform Version**
```
Terraform v0.14.4
```

### Configuration File(s)
_Reminder to redact any sensitive information that may be present in this file_

<details>
<summary>Click to toggle contents of config file </summary>

```hcl
log_level = "INFO"

port = 8558

syslog {
  enabled = true
}

buffer_period {
  enabled = true
  min     = "5s"
  max     = "20s"
}

consul {
  address = "127.0.0.1:8500"
}

driver "terraform" {
  log         = true
  persist_log = true

  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "~>3.23"
    }
  }
  backend "consul" {
    gzip = false
  }
}

terraform_provider "aws" {
  region     = "us-east-2"
  access_key = "{{ env \"AWS_ACCESS_KEY_ID\" }}"
  secret_key = "{{ env \"AWS_SECRET_ACCESS_KEY\" }}"
  token      = "{{ env \"AWS_SESSION_TOKEN\" }}"
  assume_role {
    role_arn     = "{{ env \"AWS_ROLE_ARN\" }}"
  }
}

service {
  name        = "my-application"
  datacenter  = "cloud"
  description = "all instances of the service my-application in datacenter cloud"
}

task {
  name           = "canary"
  description    = "send canary traffic to my-application in datacenter cloud"
  providers      = ["aws"]
  services       = ["my-application"]
  source         = "joatmon08/listener-rule/aws"
  version        = "0.1.3"
  variable_files = ["../../datacenter.module.tfvars"]
}
```
</details>

### Terraform Configuration Files Generated by Consul-Terraform-Sync
_Reminder to redact any sensitive information that may be present in the files_

<details>
<summary>Click to toggle contents of main.tf </summary>

```terraform
terraform {
  required_version = ">= 0.13.0, < 0.15"
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "~>3.23"
    }
  }
  backend "consul" {
    address = "127.0.0.1:8500"
    gzip    = false
[...]

PR #173, opened 2021-01-20
Add support for provider configuration containing object-type attributes

lornasong:
When converting user-defined provider configuration into provider definition
blocks with variables, object-type attributes were formatted like
`obj = var.provider_name.obj`. This led to "Error: Unsupported argument".

This PR formats object-type attributes into a block like:
```
obj {
  attr = var.provider_name.obj.attr
  ...
}
```

Fixes https://github.com/hashicorp/consul-terraform-sync/issues/171


findkim (approved):
👏 this is great!

eikenb (approved):
Nice. Simpler than I would have guessed.

PR #174, opened 2021-01-22
include service kind in module input

### LINKED ISSUE #168: Include service type in module input
Services set for module input are not easily distinguishable by Consul service type. Service types can include ([source](https://github.com/hashicorp/consul/blob/v1.9.1/api/agent.go#L15-L40)): "", connect-proxy, mesh-gateway, terminating-gateway, ingress-gateway

The service type is available by the Consul API `/v1/health/service/:service` under the field `service.kind` but not through Consul Terraform Sync.
```
$ curl localhost:8500/v1/health/service/api-sidecar-proxy
[{
  "Service": {
      "Kind": "connect-proxy",
      "ID": "api-sidecar-proxy",
      "Service": "api-sidecar-proxy",
      ...
  }
}]
```
* generated Terraform file [variables.tf](https://github.com/hashicorp/consul-terraform-sync/blob/v0.1.0-techpreview2/templates/tftmpl/variables.go)
* [hcat.HealthService](https://github.com/hashicorp/hcat/blob/cc9b9b904d72cff2d3df0caedaf1750968eca851/dep/template_function_types.go#L46)

eikenb:
Takes the newly supported hcat consul health service Kind value and
propagates it through to the module input fields.

Updated to use latest hashicat.
Add it to the services variable fields.
Rendered into templates via mapstructure rendering of data structs.
Tests and testdata updated.

Closes #168

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/hashicorp/go-syslog v1.0.0
 	github.com/hashicorp/go-uuid v1.0.2
 	github.com/hashicorp/go-version v1.2.1
-	github.com/hashicorp/hcat v0.0.0-20201211001333-cc9b9b904d72
+	github.com/hashicorp/hcat v0.0.0-20210122002113-a06a4c3ec618
 	github.com/hashicorp/hcl v1.0.0
 	github.com/hashicorp/hcl/v2 v2.6.0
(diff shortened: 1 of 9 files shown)


lornasong (approved):
makes sense to me!

PR #175, opened 2021-01-26
Support querying service by namespace

findkim:
Service config blocks with namespaces defined can now query services based on namespace when deployed with Consul enterprise.

```hcl
service {
  name = "app"
  namespace = "app-team"
}
```

Configuring the service namespace without Consul enterprise will result in an exiting error:
```
2021/01/25 18:14:43.253097 [ERR] (cli) error running controller in Once mode: health.service(app?ns=app-team&filter="Checks.Status == passing"): Unexpected response code: 400 (Bad request: Invalid query parameter: "ns" - Namespaces are a Consul Enterprise feature)
```

Pulls in changes from  https://github.com/hashicorp/hcat/pull/35
Marking this with a bug label since the namespace option was available before it was fully supported.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/hashicorp/consul v1.8.0
 	github.com/hashicorp/consul/sdk v0.5.0
 	github.com/hashicorp/go-checkpoint v0.5.0
+	github.com/hashicorp/go-cleanhttp v0.5.1
 	github.com/hashicorp/go-syslog v1.0.0
 	github.com/hashicorp/go-uuid v1.0.2
 	github.com/hashicorp/go-version v1.2.1
@@ -24,7 +25,7 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/mitchellh/mapstructure v1.3.3
 	github.com/pkg/errors v0.9.1
-	github.com/stretchr/testify v1.5.1
+	github.com/stretchr/testify v1.6.1
 	github.com/tidwall/pretty v1.0.2 // indirect
 	github.com/xdg/scram v0.0.0-20180814205039-7eeb5667e42c // indirect
 	github.com/xdg/stringprep v1.0.0 // indirect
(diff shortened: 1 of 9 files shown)


go.mod
```diff
@@ -10,10 +10,11 @@ require (
 	github.com/hashicorp/consul v1.8.0
 	github.com/hashicorp/consul/sdk v0.5.0
 	github.com/hashicorp/go-checkpoint v0.5.0
+	github.com/hashicorp/go-cleanhttp v0.5.1
```
eikenb, line 13: This isn't probably an issue with a project this size but we eventually might want to consider breaking out module updates that aren't directly related to the changeset.

findkim: I ran `go mod tidy` as a part of these changes, it looks like it picked up one that was missing for some time

```sh
$ go mod why github.com/hashicorp/go-cleanhttp
# github.com/hashicorp/go-cleanhttp
github.com/hashicorp/consul-terraform-sync/e2e/internal
github.com/hashicorp/go-cleanhttp
```

This would be relevant to the `stretchr/testify` version update below 😛 -- however since it is only used for testing I think I'll keep it as a part of this PR

eikenb:
Looks good but I wanted to see if I could get that integration test to pass but I've kicked it a couple times now and it keeps timing out. It has been causing more trouble recently but this is the first time a kick wouldn't get it to pass.

I'd rather not approve something that just won't pass the tests until we know for sure what the issue is. I'm pulling down the branch now to see if I can make more sense of it.

findkim:
The `ns` addition to the integration test for the template caused us to find a bug within the integration test. The template would never render since it was running against Consul OSS since namespace is not supported. However, the integration test would not stop on those errors and indefinitely loop.

I'm going to keep the namespace out of the integration test since our CI pipeline is running against OSS

lornasong (approved):
ooo nice find with the integration test. implementation of the hcat changes are really clean and minimal! approving on what I'm aware of!

eikenb (approved):

PR #176, opened 2021-01-27
Update examples with new template

findkim:
A few additions have been made to the generated TF files and template, updating the example files to include:
* #174 kind
* #167 include task info to generated files
* #166 user-defined metadata
* #120 simplify HCL service template

This is a manual process, can't commit to looking into automating these updates right now


lornasong (approved):
thanks for updating these! a separate thought not for this PR, I wonder if we can use the e2e test help (remember) to update this 🤔 (I had forgotten about these files!).

findkim:
@lornasong using e2e seems in the realm of possible!

The example in this directory was quickly written for visual understanding, it does not execute as-is. I've been wanting to work on a module + CTS config that is runnable and shows enough features.

PR #178, opened 2021-01-28
simple, single retry for panos' commit

eikenb:
Recommendation at this point was to retry only once after a decent delay
based on SE feedback.

Retry is very simple and groups together the commit and wait calls as
either could fail (both make http calls to the panos API with failure
conditions).

Added checks to existing tests as they already provided decent coverage.

--

PR notes:
1) what do you think of the the retry tests using expectedRetries() vs. including the expected retries in the table data?

### DIFF
--- a/handler/panos.go
+++ b/handler/panos.go
@@ -6,9 +6,11 @@ import (
 	"log"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/PaloAltoNetworks/pango"
 	"github.com/PaloAltoNetworks/pango/commit"
+	"github.com/hashicorp/consul-terraform-sync/retry"
 	"github.com/mitchellh/mapstructure"
 )
 
@@ -19,6 +21,9 @@ const (
 	// Users with custom roles currently return an error for an empty commit with
 	// this server response prefix. See GH-73 for more details.
 	emptyCommitServerRespPrefix = `<response status="success" code="13">`
+
+	// max number of retries
+	maxRetris = 4
 )
 
 //go:generate mockery --name=panosClient  --structname=PanosClient --output=../mocks/handler
@@ -47,6 +52,7 @@ type Panos struct {
 	adminUser    string
 	configPath   string
 	autoCommit   bool
+	retry        retry.Retry
 }
 
 // NewPanos configures and returns a new panos handler
@@ -107,6 +113,7 @@ func NewPanos(c map[string]interface{}) (*Panos, error) {
 		adminUser:    username,
 		configPath:   configPath,
 		autoCommit:   autoCommit,
+		retry:        retry.NewRetry(maxRetris, time.Now().UnixNano()),
 	}, nil
 }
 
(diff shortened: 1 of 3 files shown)


handler/panos_test.go
```diff
@@ -206,18 +206,39 @@ func TestPanosCommit(t *testing.T) {
...
+				case c && w: // everyone happy
+					return 1, 1
+				case c && !w: // wait fails
+					return 2, 2
+				}
+				return -1, -1 // never reached
+			}
+			commitTries, waitTries := expectedRetries()
+			m.AssertNumberOfCalls(t, "Commit", commitTries)
+			m.AssertNumberOfCalls(t, "WaitForJob", waitTries)
```
lornasong, line 241: I've never used this! Good to know about!

lornasong (approved):
looks good to me! the changes are very readable, particularly the switch-statement and comments in the tests

handler/panos.go
```diff
@@ -138,18 +144,29 @@ func (h *Panos) commit() error {
...
 	}
-	if err != nil {
-		log.Printf("[ERR] (handler.panos) error committing: %s. Server response: '%s'", err, resp)
-		return err
-	}
 
-	if err := h.client.WaitForJob(job, nil); err != nil {
-		log.Printf("[ERR] (handler.panos) error waiting for panos commit to finish: %s", err)
-		return err
+	if err := tryCommit(); err != nil {
```
findkim, line 164: Not sure if you're already aware, but wanted mention that there's an existing [`retry` package](https://github.com/hashicorp/consul-terraform-sync/blob/master/retry/retry.go) that is pretty small and is currently used for [retrying task execution in daemon mode](https://github.com/hashicorp/consul-terraform-sync/blob/master/controller/readwrite.go#L203). It has slightly different logic in that it uses a random delay based on seed. Would you want to repurpose/modifie for this single retry too?

eikenb: I missed retry getting added. I had thought about using a more standard retry algorithm but was waiting to talk to Tristan about it. Now (post meeting) that I have and determined exponential backoff is appropriate I can just use it as is... Thanks!

Also, more importantly IMO, that reminded me we really should be passing the Context through the handler Do() calls to allow for interruptions.

I'm going to rework handlers to pass Context and this case to use that retry package.

eikenb: All done with these changes. Thanks again.

handler/panos.go
```diff
@@ -19,6 +21,9 @@ const (
 	// Users with custom roles currently return an error for an empty commit with
 	// this server response prefix. See GH-73 for more details.
 	emptyCommitServerRespPrefix = `<response status="success" code="13">`
+
+	// max number of retries
+	maxRetris = 4
```
lornasong, line 26: small typo: "maxRetris" => "maxRetries"

retry/retry.go
```diff
@@ -25,7 +27,7 @@ func NewRetry(maxRetry uint, seed int64) Retry {
 }
 
 // Do calls a function with exponential retry with a random delay.
-func (r *Retry) Do(ctx context.Context, f func(context.Context) error, desc string) error {
+func (r Retry) Do(ctx context.Context, f func(context.Context) error, desc string) error {
```
lornasong, line 30: ahhh👍

retry/retry.go
```diff
@@ -81,3 +86,11 @@ func (r *Retry) waitTime(attempt uint) int {
 	total := (baseTimeSeconds + delay) * float64(time.Second)
 	return int(total)
 }
+
+// Test version, returns retry in test mode (nanosecond retry delay).
+func NewTestRetry(maxRetry uint) Retry {
```
lornasong, line 91: good idea! (note to self to use this in the future)

lornasong (commented):
this looks good to me! 🎉 reusing the retry library as @findkim mentioned. i only noticed a very minor typo that we can include as part of some other pr for the next time we work in the panos code area

handler/panos.go
eikenb: Saw the approved and missed this. Will fix in a small follow up commit.

retry/retry.go
eikenb: I considered a 'Retrier' interface that I could swap out with a test version but the semantics of the interface meant the nil value resulted in a panic and the changes to fix that were ugly. So I preserved using the struct directly and added a test mode to it.

PR #180, opened 2021-01-29
propagate context through handlers

eikenb:
Handler's Do() methods need to be cancellable via context and having the context helps. This makes sure the context is passed along.

Split this out from panos-retry work.

### DIFF
--- a/handler/handler.go
+++ b/handler/handler.go
@@ -1,6 +1,7 @@
 package handler
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/pkg/errors"
@@ -14,7 +15,7 @@ type Handler interface {
 
 	// Do executes the handler. Receives previous error and returns previous
 	// error wrapped in any new errors.
-	Do(error) error
+	Do(context.Context, error) error
 
 	// SetNext sets the next handler that should be called
 	SetNext(Handler)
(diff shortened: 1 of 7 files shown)


handler/handler.go
```diff
@@ -43,11 +44,11 @@ func TerraformProviderHandler(providerName string, config interface{}) (Handler,
 }
 
 // callNext should be called by a handler's Do() to call the next handler
-func callNext(nextH Handler, prevErr, err error) error {
+func callNext(ctx context.Context, nextH Handler, prevErr, err error) error {
 	nextErr := nextError(prevErr, err)
 
 	if nextH != nil {
-		return nextH.Do(nextErr)
+		return nextH.Do(ctx, nextErr)
```
findkim, line 51: To make use of the context, it would need to check if the context has been cancelled before calling the next handler to determine if it should continue or stop further execution.

eikenb: I was focusing on just getting it plumbed through but that is an obvious case for it.
I'll add a check there.

findkim: Thanks for looking into it. Context handling within the scope of the panos commit seems appropriate for the other PR, this one just stuck out to me if we could stop earlier.

findkim (approved):

PR #181, opened 2021-02-03
Set provider intermediate variable as sensitive

### LINKED ISSUE #85: Look into using Terraform sensitive attribute
Terraform feature to [add sensitive attribute to variables](https://github.com/hashicorp/terraform/pull/26183) is currently slated for version 0.14. This allows adding a `sensitive` attribute to variables which will prevent outputting those variable values during plan. This will be helpful for read-only mode and potentially elsewhere.

findkim:
Sensitive argument won't prevent provider args from being stored in the state files, but will omit them from Terraform output and logs.

Conditionally adds the `sensitive = true` to the generated `variables.tf` when the discovered/installed Terraform version is 0.14+

Resolves #85

### DIFF
--- a/templates/tftmpl/golden_test.go
+++ b/templates/tftmpl/golden_test.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"testing"
 
+	goVersion "github.com/hashicorp/go-version"
 	"github.com/hashicorp/consul-terraform-sync/templates/hcltmpl"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -69,6 +70,7 @@ func TestNewFiles(t *testing.T) {
 			Func:   newVariablesTF,
 			Golden: "testdata/variables.tf",
 			Input: RootModuleInputData{
+				TerraformVersion: goVersion.Must(goVersion.NewSemver("0.14.2")),
 				Providers: []hcltmpl.NamedBlock{hcltmpl.NewNamedBlock(
 					map[string]interface{}{
 						"testProvider": map[string]interface{}{
--- a/templates/tftmpl/golden_test.go
+++ b/templates/tftmpl/golden_test.go
@@ -70,7 +70,7 @@ func TestNewFiles(t *testing.T) {
 			Func:   newVariablesTF,
 			Golden: "testdata/variables.tf",
 			Input: RootModuleInputData{
-				TerraformVersion: goVersion.Must(goVersion.NewSemver("0.14.2")),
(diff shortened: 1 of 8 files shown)


templates/tftmpl/golden_test.go
```diff
@@ -69,6 +70,7 @@ func TestNewFiles(t *testing.T) {
 			Func:   newVariablesTF,
 			Golden: "testdata/variables.tf",
 			Input: RootModuleInputData{
+				TerraformVersion: goVersion.Must(goVersion.NewSemver("0.14.2")),
```
eikenb, line 73: My nose started twitching when I saw a version hardcoded in a string but it looks like it is more an example than something that actually uses that version. If I'm wrong and it is actually using that maybe it would be better moved to a constant. If I'm right maybe change it to something that would make it more obvious that it isn't real (like 9.99.9 or something).

templates/tftmpl/variables.go
```diff
@@ -7,11 +7,16 @@ import (
...
 	"github.com/hashicorp/consul-terraform-sync/templates/hcltmpl"
+	goVersion "github.com/hashicorp/go-version"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
 	"github.com/hashicorp/hcl/v2/hclwrite"
 	"github.com/zclconf/go-cty/cty"
 )
 
+// tfVersionSensitive is the first version to support the sensitive argument
+// for variables
+var tfVersionSensitive = goVersion.Must(goVersion.NewSemver("0.14.0"))
```
eikenb, line 18: Will this version need to be adjusted at some point or will it become a legacy thing (ie. when we're on version 0.42.0 a few years down the road and no one uses anything this old we'd just remove it)? If we want to have a minimal version we support then maybe store these values in an obvious, common place... maybe a requirements.go or something?

findkim: This version is only relevant to sensitive vars and not the Terraform version constraint supported. The supported Terraform version exists in version/terraform.go

I would much prefer to keep this variable where it's used since it only makes sense in the context of this file.

findkim: I'd like to follow the pattern to support version n-2. And so if 0.16 is a thing or if Terraform 1.0s, then this can be dropped and assumed always sensitive

templates/tftmpl/golden_test.go
findkim: Not an example. This is a test case. It's no different than another test value that has hard coded strings. Yeah I can change it to a nonsensical version

lornasong (approved):
makes sense to me!

eikenb:
all works for me as well.

PR #182, opened 2021-02-04
Separate provider vars from services var

findkim:
Separating the provider vars allows the `terraform.tfvars` file to be more easily shareable now that potentially sensitive values are not included. This also removes provider info from being unnecessarily included in the `.tfvars.tmpl` template file.

Resolves #172

PR #183, opened 2021-02-04
Include providers.tfvars file to TF commands

findkim:
Follow-up to include passing the `providers.tfvars` to the Terraform commands. This file is always created, although it may not have variables if there are no providers configured for the task.

Related PR https://github.com/hashicorp/consul-terraform-sync/pull/182


client/terraform_cli_test.go
```diff
@@ -14,11 +14,14 @@ import (
...
+		tfvars := tfexec.VarFile("terraform.tfvars")
+		ptfvars := tfexec.VarFile("providers.tfvars")
+
 		m := new(mocks.TerraformExec)
 		m.On("SetEnv", mock.Anything).Return(nil)
 		m.On("Init", mock.Anything).Return(nil)
-		m.On("Apply", mock.Anything, mock.Anything).Return(nil)
-		m.On("Plan", mock.Anything, mock.Anything).Return(true, nil)
+		m.On("Apply", mock.Anything, tfvars, ptfvars).Return(nil)
+		m.On("Plan", mock.Anything, tfvars, ptfvars).Return(true, nil)
```
eikenb, line 24: It took me a minute to realize this is mocking the tfexec calls and not the above methods.

Commenting only to point this out in case someone else pauses on the same thing.  Nothing wrong, just the layout of this PR made it a bit confusing.

eikenb (approved):
:+1:

PR #184, opened 2021-02-05
Fix task_env type validation

findkim:
HCL config unmarshals `task_env` to interface and type casting nested interfaces requires converting each item and not the collection -- `map[string]interface{}` and so the type validation expecting `map[string]string` would always fail and CTS would error and exit on configuration.

Related #157

Repro steps
1. Config file with `task_env` block for any provider

```hcl
terraform_provider "local" {
  task_env {
    "my_env" = "foo"
  }
}

task {
  name        = "test"
  source      = "<any module>"
  services = ["api", "web"]
}
```

2. Run CTS
```shell
$ consul-terraform-sync -config-file task_env.hcl -inspect
2021/02/04 18:14:47 [ERR] (cli) error validating configuration: unexpected task_env block format
```

### DIFF
--- a/config/provider.go
+++ b/config/provider.go
@@ -146,9 +146,14 @@ func (c *TerraformProviderConfig) Validate() error {
 		// Validate task_env format if exists
 		taskEnv, exists := block["task_env"]
 		if exists {
-			_, ok := taskEnv.(map[string]string)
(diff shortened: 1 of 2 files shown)


lornasong (approved):
makes sense to me!

PR #185, opened 2021-02-10
Refactor hashicat templating to driver-level

lornasong:
This is the first in a series of changes to support enable/disable CLI

Refactors creating a task template into a new driver method `initTaskTemplate()`
and refactors out rendering the template into a new driver method
`RenderTemplate()`. In order to support a CLI that will update task lifecycle,
we will to be able to recreate task templates and render templates on demand.

Changes:
 - controller.newTaskTemplate() => driver.initTaskTemplate()
 - unit.template => driver.template
 - controller.fileReader => driver.fileReader
 - controller.checkApply's resolver.Run + tmpl.Render => driver.RenderTemplate()
 - Added UserDefinedMeta field to task to avoid passing configuration into driver


controller/controller.go
```diff
@@ -250,58 +233,21 @@ func newDriverTasks(conf *config.Config, providerConfigs driver.TerraformProvide
...
-	if err != nil {
-		return nil, err
-	}
-
-	renderer := hcat.NewFileRenderer(hcat.FileRendererInput{
-		Path:  tfvarsFilepath,
-		Perms: filePerm,
-	})
-
-	metaMap := conf.Services.CTSUserDefinedMeta(task.Services)
```
lornasong, line 296: Moved storing this meta map in driver.Task so that the refactored method wouldn't depend on `conf` down the line

controller/controller_test.go
```diff
@@ -93,35 +93,19 @@ func TestBaseControllerInit(t *testing.T) {
 	cases := []struct {
 		name        string
 		expectError bool
-		initErr     error
```
lornasong, line 96: `initErr` wasn't being used anymore. just some cleanup.

controller/readonly_test.go
```diff
@@ -18,64 +18,48 @@ func TestReadOnlyRun(t *testing.T) {
...
-			tmpl := new(mocks.Template)
-			tmpl.On("Render", mock.Anything).Return(hcat.RenderResult{}, tc.templateRenderErr).Once()
-
-			r := new(mocks.Resolver)
-			r.On("Run", mock.Anything, mock.Anything).
-				Return(hcat.ResolveEvent{Complete: true}, tc.resolverRunErr)
-
 			w := new(mocks.Watcher)
-			w.On("Wait", mock.Anything).Return(tc.watcherWaitErr).
-				On("Size").Return(5)
```
lornasong, line 70: the call to watcher.Wait wasn't needed so I also cleaned up `watcherWaitErr`

controller/readwrite_test.go
```diff
@@ -23,38 +22,26 @@ func TestReadWrite_CheckApply(t *testing.T) {
 	t.Parallel()
 
 	cases := []struct {
-		name              string
-		expectError       bool
-		applyTaskErr      error
-		resolverRunErr    error
-		templateRenderErr error
```
lornasong, line 30: `templateRenderErr` also just some cleanup

controller/controller.go
findkim: User-defined meta is on the service but is used by the task, so adding it to `driver.Task` makes sense 👍

findkim (approved):
A few months ago with the decision to have 1 task per driver (instead of the initial 1 driver for all tasks) it resulted in a mess of mixing config + task mangling + driver in various functions, i.e. `newTaskTemplate`.

This is a great refactor that cleans up the responsibilities of the controller to manage the watcher + driver, and the driver to manage the template (initialization, rendering) and execution of the task using the template.

I looked through your changes a few times because it touches a lot of files, but I think you did a good job combing through it already I didn't find anything else!

lornasong:
thanks for the thorough review and feedback @findkim! as I was doing this refactor I was feeling very thankful that we went with the 1 task to 1 driver decision ❤️

PR #186, opened 2021-02-10
Refactor buffer_period to driver-level

lornasong:
Discovered a bug in earlier refactor https://github.com/hashicorp/consul-terraform-sync/pull/185.
If a task is configured as disabled from the start, a template will not be
generated and template == nil. readwrite.setTemplateBufferPeriods() calls
driver.TemplateID() which throws a nil pointer panic.

Refactor setTemplateBufferPeriods():
 - Add controller.getTemplateBufferPeriods() and then set buffer info on driver.Task
 - Add driver.SetBufferPeriods() which sets buffer info from driver.Task on watcher
 - Update readwrite.Run() to call driver.SetBufferPeriods()
 - Remove the need for driver.TemplateID()

### DIFF
--- a/driver/task.go
+++ b/driver/task.go
@@ -2,6 +2,7 @@ package driver
 
 import (
 	"log"
+	"time"
 
 	"github.com/hashicorp/consul-terraform-sync/client"
 	mocks "github.com/hashicorp/consul-terraform-sync/mocks/client"
@@ -16,6 +17,13 @@ type Service struct {
 	Tag         string
 }
 
+// BufferPeriod contains the task's buffer period configuration information
+// if enabled
+type BufferPeriod struct {
+	Min time.Duration
+	Max time.Duration
+}
+
 // Task contains task configuration information
 type Task struct {
 	Description     string
@@ -27,6 +35,7 @@ type Task struct {
 	VarFiles        []string
 	Version         string
 	UserDefinedMeta map[string]map[string]string
+	BufferPeriod    *BufferPeriod // nil when disabled
 }
 
 // ProviderNames returns the list of providers that the task has configured
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -242,6 +242,7 @@ func newDriverTasks(conf *config.Config, providerConfigs driver.TerraformProvide
 			VarFiles:        t.VarFiles,
 			Version:         *t.Version,
 			UserDefinedMeta: conf.Services.CTSUserDefinedMeta(t.Services),
+			BufferPeriod:    getTemplateBufferPeriod(conf, t),
 		}
 	}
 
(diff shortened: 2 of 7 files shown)


lornasong:
Named this PR after the refactor since the bug can't be replicated with existing code (we don't support disabled tasks). But from a code perspective, it is a bug (unhandled pointer).

driver/task.go
```diff
@@ -27,6 +34,7 @@ type Task struct {
 	VarFiles        []string
 	Version         string
 	UserDefinedMeta map[string]map[string]string
+	BufferPeriod    *BufferPeriod
```
eikenb, line 37: I don't think this needs to be a pointer. For changes to the Task, it is already a pointer. I don't see any mutations of BufferPeriod in place and I don't see any cases where the nil case is needed (checking for unset doesn't seem required over 0-case) in either this PR now #185.

findkim (approved):
If I'm understanding correctly, the bug exists for your upcoming work for disabling tasks and not reproducible in main branch. The changes seem to work and follows the existing functionality:
* Defaults to global buffer period, overrides with task buffer period if enabled
* Disabled global + task enabled, buffer is set
* Disabled global + task, no buffer is set

driver/task.go
eikenb: I suggest this as the semantics for 0-values is much nicer (no panics) and values allow for better performance (can be allocated on the stack, so less GC).

lornasong: Yea, there isn't any mutations of BufferPeriod. I made BufferPeriod a pointer so that the nil case would be meaningful: nil = buffer_period is disabled.

The alternative would be instead to have an enabled field on the BufferPeriod. This was a little less intuitive to me from the backend since disabled BufferPeriod should not have mix/max values. However, this would be more in line with the configuration structure 🤔 .

What do you think? You mention a lot of other positives for not having a pointer. Is your preference still for not having a pointer? Happy to change, just want to double-check.

lornasong:
thanks @findkim! yes, you are understanding correctly that the bug is not reproducible in the main branch and exists for upcoming work. thanks for confirming the different scenarios!

driver/task.go
eikenb: If nil is used as an unset/disabled value where the 0 value won't work, then it makes sense. I probably missed or misinterpreted where that was done. Using an enabled field is an option but I don't think it is a slam dunk improvement as it has downsides as well.

TLDR; it is being used to represent 'disabled' and no change is necessary.

eikenb (approved):

driver/task.go
lornasong: Gotcha! Thanks for your input and helpful to hear your train of thought. Let me add a comment to clarify that the nil value is 'disabled' since it's not not very apparent.

PR #187, opened 2021-02-11
Attempt to improve consul service templating flakiness

lornasong:
`TestRenderTFVarsTmpl/happy_path` occasionally fails where the templated service
information is missing one or two of the registered services. Of the 4 fails
observed, two are missing api.worker-01.dc1 and api-2.worker-01.dc1. The other
two are missing web.worker-01.dc1

I haven’t been able to debug this successfully (it started passing when I tried).
This change is a pre-emptive measure to increase the context cancel a little and
retry the test once in case the race is between the test and the services
registering.

### DIFF
--- a/templates/tftmpl/integration_test.go
+++ b/templates/tftmpl/integration_test.go
@@ -229,14 +229,26 @@ func TestRenderTFVarsTmpl(t *testing.T) {
 			}
 			tmpl := hcat.NewTemplate(input)
 
-			ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+			ctx, cancel := context.WithTimeout(context.Background(), time.Second*8)
 			defer cancel()
+
+			gld, err := ioutil.ReadFile(tc.goldenFile)
+			if err != nil {
+				require.NoError(t, err)
+			}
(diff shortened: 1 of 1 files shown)


findkim (approved):

templates/tftmpl/integration_test.go
```diff
@@ -229,14 +229,26 @@ func TestRenderTFVarsTmpl(t *testing.T) {
 			}
 			tmpl := hcat.NewTemplate(input)
 
-			ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+			ctx, cancel := context.WithTimeout(context.Background(), time.Second*8)
 			defer cancel()
+
+			gld, err := ioutil.ReadFile(tc.goldenFile)
```
eikenb, line 235: This will help more than anything as it should give Consul more time to fully register everything which is probably the issue.

lornasong: Good point! I also noticed that the e2e test has a similar Consul registration setup. It doesn't have similar flakiness though probably since it has a buffer when CTS starts up, similar to this ReadFile.

PR #189, opened 2021-02-11
Consume `enabled` field in task driver

lornasong:
- Previously added `enabled` field to task configuration
 - Update driver methods to not make changes if task is disabled
 - e2e test for configuring a disabled task

### DIFF
--- a/driver/terraform_test.go
+++ b/driver/terraform_test.go
@@ -70,7 +70,7 @@ func TestRenderTemplate(t *testing.T) {
 			tmpl.On("Render", mock.Anything).Return(hcat.RenderResult{}, tc.renderErr).Once()
 
 			tf := &Terraform{
-				task:     Task{Name: "RenderTemplateTest"},
+				task:     Task{Name: "RenderTemplateTest", Enabled: true},
 				resolver: r,
 				template: tmpl,
 			}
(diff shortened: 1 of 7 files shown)


driver/terraform_test.go
```diff
@@ -278,6 +278,35 @@ func TestGetTerraformHandlers(t *testing.T) {
 	}
 }
 
+func TestDisabledTask(t *testing.T) {
+	t.Run("disabled-tasks", func(t *testing.T) {
+		// tests that disabled tasks don't require mocking any calls and does
+		// not throw any errors
```
findkim, line 284: Helpful context here!

e2e/e2e_test.go
```diff
@@ -218,6 +218,40 @@ func TestE2ELocalBackend(t *testing.T) {
...
+	err = makeTempDir(tempDir)
+	// no defer to delete directory: only delete at end of test if no errors
+	require.NoError(t, err)
+
+	configPath := filepath.Join(tempDir, configFile)
+	err = makeConfig(configPath, disabledTaskConfig(srv.HTTPAddr, tempDir))
+	require.NoError(t, err)
+
+	// deliberately running sync in long running mode. This catches a bug in
+	// when setting buffer_period for a disabled task
```
findkim, line 238: This comment might lose context as it ages, like it's catching an active bug. Do you mind rewording it to be more generic? Just throwing out a suggestion to give express better what I mean

> ... This may catch bugs where functions execute unexpectedly on disabled tasks.

findkim (approved):
I'm impressed on this simple solution. I for some reason had imaged disabling tasks during run time to be messy. Nice

e2e/e2e_test.go
lornasong: Good point! I wrote the comment when it was active and it's definitely lost context since. Let me reword. Thank you!

PR #190, opened 2021-02-13
Share controller drivers with API

lornasong:
In order to support an update task API for the disable/enable cli feature, APIs
will need to call the drivers for the associated task.

Changes
 - Update controller.Init() to also return a map of drivers
 - Update API with a `drivers` fields
 - Refactor cli to serve API after controller.Init() is called in order to
 consume drivers
 - Update driver to be concurrent with locks (separate commit)

(I have a subsequent PR that will consume the new drivers field in the API. The changes got sort of big so wanted to make separate PRs)

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/hashicorp/consul-terraform-sync/driver"
 	"github.com/hashicorp/consul-terraform-sync/event"
 )
 
@@ -54,13 +55,14 @@ const (
 // API supports api requests to the cts biniary
 type API struct {
 	store   *event.Store
+	drivers map[string]driver.Driver
 	port    int
 	version string
 	srv     *http.Server
 }
 
 // NewAPI create a new API object
-func NewAPI(store *event.Store, port int) *API {
+func NewAPI(store *event.Store, drivers map[string]driver.Driver, port int) *API {
 	mux := http.NewServeMux()
 
 	// retrieve overall status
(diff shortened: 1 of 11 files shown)


eikenb:
> The changes got sort of big so wanted to make separate PRs
:100: .. Thank you. I much prefer more broken down PRs.

eikenb (approved):
Everything looked fine to me. Broken up into good, logical commits.
:+1:

api/api.go
```diff
@@ -54,13 +55,14 @@ const (
 // API supports api requests to the cts biniary
 type API struct {
 	store   *event.Store
+	drivers map[string]driver.Driver
```
findkim, line 58: Maps are by reference, and so it now seems that there are two objects accessing the same map of drivers. Albeit both are only reads (API and Controller), it does pose a potential area where one object could unintentionally modify the map.

^ the above callout isn't too concerning now, but an option could be to upgrade the map to a full on object to obscure the map because `driver.Drivers` is shorter to type than `map[string]driver.Driver`  😝

lornasong: Yes! Thanks for calling this out. I was considering a driver.Drivers since I think, to your point about currently there is only reads on the map, creating and deleting tasks will need to write. I chose the map so that I could focus on disable/enable tasks but hadn't considered the potential for unintentionally modifying the map. Making this change will definitely be first on the list for supporting create/delete taks cli.

findkim: With create/delete on the road _map_ `driver.Drivers` makes more sense, good to wait for then for those changes

lornasong: Just following up here, this is now done (https://github.com/hashicorp/consul-terraform-sync/pull/218)! Thanks for the feedback!

PR #191, opened 2021-02-16
Add Update Task API with initial support to disable/enable a task

lornasong:
Add support for `PATCH /v1/tasks/<task_name> -data '{"enabled:"true}'`. This will
be used for the upcoming disable task CLI. Additional functionality will be
added in subsequent PRs for enable task CLI.

more details):
 - Refactor api's getTaskName() to be shared across handlers
 - Add driver.UpdateTask() to allow patch updating a task
 - Add Update Task API

Note: ^ I tried to split out changes into commits. Not perfect. There's a couple lines
(particularly imports) that aren't in the right commit.

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -225,3 +225,57 @@ func TestJsonResponse(t *testing.T) {
 		})
 	}
 }
+
+func TestGetTaskName(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name      string
+		path      string
+		expectErr bool
+		expected  string
+	}{
+		{
+			"all task statuses",
+			"/v1/status/tasks",
+			false,
+			"",
+		},
+		{
+			"task status for a specific task",
+			"/v1/status/tasks/my_specific_task",
+			false,
+			"my_specific_task",
+		},
+		{
+			"empty task name",
+			"/v1/status/tasks/",
+			false,
+			"",
+		},
+		{
+			"tasks task name",
+			"/v1/status/tasks/tasks",
+			false,
+			"tasks",
+		},
+		{
+			"invalid name",
+			"/v1/status/tasks/mytask/stuff",
+			true,
+			"",
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			actual, err := getTaskName(tc.path, taskStatusPath, "v1")
+			if tc.expectErr {
+				assert.Error(t, err)
+				return
+			}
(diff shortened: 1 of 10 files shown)


eikenb (approved):
Everything looks good. Great job breaking into logical commits, they read very nicely, and I really liked how you re-used the Handler system to manage the tasks.

I do think the driver.Driver interface is getting a bit large but given the current structure of things, reworking it now doesn't make sense. However should keep an eye out for opportunities to break it apart.

lornasong:
Thanks for the review @eikenb! Helpful to know that breaking into the commits is more readable :). I'll make sure to continue doing that.

Good point about the driver.Driver interface. I have prepared changes that will add on one method (eek). Yes, definitely on the look out for opportunities to break it apart.

api/task_test.go
```diff
@@ -0,0 +1,204 @@
...
+			"/v1/tasks/xyz",
+			http.MethodPut,
+			``,
+			http.StatusMethodNotAllowed,
+		},
+	}
+
+	drivers := make(map[string]driver.Driver)
+	patchUpdateD := new(mocks.Driver)
+	patchUpdateD.On("UpdateTask", mock.Anything).Return(nil).Once()
```
findkim, line 65: I'm seeing a lot of `mock.Anything` in mocked method signatures throughout and just wanted to callout that this would skip out on testing that the function is receiving the correct parameter. It's useful for tests that aren't testing that specific parameter.

Not looking for any change here, just wanted to share more about the `mock` testing options. You could optionally make the test more granular to specify `UpdateTaskConfig{Enabled: true}` for the test "patch update task".

api/api_test.go
```diff
@@ -5,10 +5,14 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 	"time"
 
+	mocks "github.com/hashicorp/consul-terraform-sync/mocks/driver"
+
```
findkim, line 13: Not sure if this is intentional to have this import grouped by itself

lornasong: Not intentional, I have a future set of changes that modifies this files as well. I've updated that set to include regrouping this import with the other cts imports. Thanks!

lornasong: Note: following up, here's the commit to regroup the import https://github.com/hashicorp/consul-terraform-sync/pull/196/commits/4b040ed32228aff0fdadc6049f46939afcb16e5e, part of PR-196

PR #192, opened 2021-02-16
Add API client for Update Task API

lornasong:
more details):
 - Add driver.Task() so that we can view state of task for testing (and Status
 API down the line)
- Add API client

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -31,6 +31,8 @@ type Driver interface {
 	// UpdateTask supports updating certain fields of a task
 	UpdateTask(task PatchTask) error
 
+	Task() Task
+
 	// Version returns the version of the driver.
 	Version() string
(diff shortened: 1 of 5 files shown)


driver/driver.go
```diff
@@ -31,6 +31,8 @@ type Driver interface {
 	// UpdateTask supports updating certain fields of a task
 	UpdateTask(task PatchTask) error
 
+	Task() Task
```
eikenb, line 34: looks like it needs a doc comment

eikenb (approved):
Other than maybe adding that doc comment it looks good.
:+1:

driver/driver.go
lornasong: Ack! Thank you. I will add that now.

PR #193, opened 2021-02-16
Make a testutils package to hold helper functions shared across packages

lornasong:
Changes:
 - Refactor out makeTempDir to testutils

It was copied in two packages and we have at least one other test-helper-type method that would benefit from consolidating in a testing package


eikenb (approved):
LGTM.

I like the pattern of returning the cleanup func.. so an additional :+1: there. :)

PR #194, opened 2021-02-17
Support disable CLI

lornasong:
`consul-terraform-sync task disable <task_name>`

more details):
 - Some initial minimal CLI command support. Taken from Nomad and Terraform
 - Support disable CLI (fixes some typos in previous change too)

Note: planning to write an E2E test once I also update Status API with enabled/disabled. In the test, then I will be able to run the disable CLI and use the status API to confirm changes in disabled/enabled.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -13,8 +13,6 @@ import (
 	"syscall"
 	"time"
 
-	mcli "github.com/mitchellh/cli"
-
 	"github.com/hashicorp/consul-terraform-sync/api"
 	"github.com/hashicorp/consul-terraform-sync/command"
 	"github.com/hashicorp/consul-terraform-sync/config"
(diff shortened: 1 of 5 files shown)


command/meta.go
```diff
@@ -0,0 +1,67 @@
...
+		c.Name())
+	help = wordwrap.WrapString(help, width)
+
+	m.UI.Output(help)
+	return false
+}
+
+func (m *meta) client() *api.Client {
+	return api.NewClient(&api.ClientConfig{
+		// TODO: add support for configuring port when doing general options
```
lornasong, line 64: I'm planning to circle back to this in the next couple days when I work on adding support for TLS as well

eikenb (approved):
All looks good to me. :+1: 

"mitchellh/cli" is a pretty good cli library and (of course) the standard around here and adopting it will make sub-commands much easier.

I'm liking the amount of change per-commit. Nice balance.

cli.go
```diff
@@ -8,11 +8,15 @@ import (
 	"log"
 	"os"
 	"os/signal"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
 
+	mcli "github.com/mitchellh/cli"
+
```
findkim, line 17: Not sure if this is intentional to have this import grouped by itself

lornasong: Not intentional, I usually add the an import somewhere and assumed goimports organized it nicely. Looks like goimports respects newlines. I'll move it to be with the other non-standard library imports.

PR #195, opened 2021-02-18
Support run options 'inspect' and 'now' in driver.UpdateTask

lornasong:
This work will support the enable cli to be able to generate a plan for the
operator to confirm and then run immediately if confirmed. The Update Task API
will consume these as ?run=now and ?run=inspect query parameters.

more details):
 - Refactor out the logic from the driver public methods into private methods:
 RenderTemplate(), InspectTask(), ApplyTask(). UpdateTask will call the private
 methods.
 - Update inspectTask() so that it can have the option to return the plan
 information rather than simply logging it out
 - Add watcher as a field on driver so that UpdateTask can call renderTemplate()
 which requires watcher
 - Refactor SetBufferPeriod and RenderTemplate to use driver's watcher field
 rather than passing in the watcher is a method parameter
 - Add support for run=now (by calling applyTask) and run=inspect (by calling
 inspectTask) in UpdateTask


driver/terraform.go
```diff
@@ -224,24 +225,78 @@ func (tf *Terraform) ApplyTask(ctx context.Context) error {
...
+	default:
+		// onward!
 	}
 
 	if reinit {
+		if err := tf.initTask(true); err != nil {
+			return "", fmt.Errorf("Error updating task '%s'. Unable to init "+
+				"task: %s", tf.task.Name, err)
+		}
 		tf.inited = false
```
findkim, line 264: This took me a couple of reads to understand `tf.initTask` is different than `tf.inited` because it's been a while since I've viewed this file. Not an ask but something in the future we can do is add context to what `inited` is pertaining to, the Terraform workspace.

driver/terraform.go
```diff
@@ -224,24 +225,78 @@ func (tf *Terraform) ApplyTask(ctx context.Context) error {
...
+	if patch.RunOption == RunOptionInspect {
+		log.Printf("[TRACE] (driver.terraform) update task '%s'. inspect run "+
+			"option", tf.task.Name)
+		plan, err := tf.inspectTask(ctx, true)
+		if err != nil {
+			return "", fmt.Errorf("Error updating task '%s'. Unable to inspect "+
+				"task: %s", tf.task.Name, err)
+		}
+
+		plan = strings.ReplaceAll(plan, "Terraform", "CTS")
```
findkim, line 288: Can you expand the benefits of modifying the Terraform plan output?

api/task.go
```diff
@@ -116,7 +117,10 @@ func (h *taskHandler) updateTask(w http.ResponseWriter, r *http.Request) {
 		patch.Enabled = config.BoolVal(conf.Enabled)
 	}
 
-	err = d.UpdateTask(patch)
+	ctx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
findkim: On second thought, seeing how this context is handed to Terraform apply later in `UpdateTask`. What would the suggested behavior be? Should it be an asynchronous apply or could the client cancel the apply?

driver/terraform.go
```diff
@@ -439,6 +494,22 @@ func (tf *Terraform) initTaskTemplate() error {
 	return nil
 }
 
+// logDepSize logs the watcher dependency size every nth iteration. Set the
+// iterator to a negative value to log each iteration.
+//
+// This is the same as the controller.logDepSize(). TODO: consider consolidating
+func (tf *Terraform) logDepSize(n uint, i int64) {
```
findkim, line 501: I believe this is the same watcher on the controller shared across all of the drivers, logging at at the controller level should be sufficient and this duplicate logging per driver.

driver/task.go
```diff
@@ -8,9 +8,20 @@ import (
 	mocks "github.com/hashicorp/consul-terraform-sync/mocks/client"
 )
 
+const (
```
eikenb, line 11: As this grows it might be worthwhile to convert it to follow Go's enum-like pattern using iota with a custom type to allow for more compile time checks with these options.

eikenb: This in "this grows" meaning the list of possible options.

eikenb: Note that next time I'll put this below the constant values so they are in the displayed snippet.

api/task.go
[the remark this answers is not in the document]
eikenb: I was also wondering about this... A quick reading of it makes it seem pointless as the d.UpdateTask() call would need to block until complete in order for this to work. So the cancel is always called after the context's use has already finished, in which case a context.Background() would suffice.

If the plan is to convert this to an async call later that would use the cancel, I'd suggest waiting until that time to add it.

driver/terraform.go
lornasong: It is the same watcher as the controller. I hadn't thought of that the controller logDepSize() will encompass any watcher calls in UpdateTask. Thanks! Let me remove it.

driver/terraform.go
lornasong: Yea, so this plan will be outputted in the enable CLI something like:
```
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create

CTS will perform the following actions:

 // diff
```

It looked a little confusing to me to say "Terraform will perform the following actions" but could just be me. 🤔 it felt a bit wrong in spirit to replace "Terraform" though too. Maybe `strings.ReplaceAll(plan, "Terraform", "Terraform (on behalf of CTS)")`? Do you have thoughts?

driver/terraform.go
lornasong: Agree, we have a lot of 'init's going on. This tf.inited is an optimization to track if we've done a `terraform init` so that we don't need to call it every time before `terraform apply`. One thing I can do now, I'll move the `tf.inited = false` within tf.initTask() since it's likely every time we call tf.initTask(), we will likely need to call `terraform init` again.

api/task.go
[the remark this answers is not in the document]
lornasong: Agree, the cancel is not used meaningfully. I think I might have seen this in some CLI implementation elsewhere and used it without much thought! Thanks for catching this and giving feedback, @findkim and @eikenb! I'll remove the cancel and we can add it back if we ever find it meaningful.

driver/task.go
lornasong: Interesting! I had seen the iota pattern around and didn't realize it was to solve the go enum problem. I'll keep this in mind. In my next PR, I'll use these values so that they can be passed as query parameter values in the API e.g. ?run=now and ?run=inspect. I'm not sure if the iota types will translate ok for that?

driver/terraform.go
findkim: You're right, it won't encompass any watcher calls. To clarify about what I mean on duplicate: the watcher size is across all templates, and so this would log the same value per driver(task) as the one in the controller, not providing any new information for users.

driver/terraform.go
findkim: I see thanks for the example output!

> it felt a bit wrong in spirit to replace "Terraform"

Yeah I share that sentiment too, but I can see the value you're bringing to bridge that CTS is calling Terraform. How about appending the CTS context before Terraform's output?

```
<CTS info here executing Terraform plan>

An execution plan has been generated...
  + create

  // diff
```

driver/terraform.go
findkim: I like that minor change, it makes sense.

driver/terraform.go
lornasong: Ah. Gotcha. Thanks for clarifying!

driver/terraform.go
lornasong: I like the idea of a preface mentioning Terraform! There is actually a preface in the CLI-side prior to printing out the plan. I might actually modify the preface in the CLI to mention Terraform there instead and keep this just the 'pure' plan. I'll delete the string parsing here. Thanks!

findkim (approved):

PR #196, opened 2021-02-19
Modify Update Task API & client with run query parameter

lornasong:
Support `?run=now` and `?run=inspect` in the Update Task API and client. The
enable CLI will use these to output a plan for the operator (?run=inspect) and
on approval execute the changes (?run=now).

more details):
 - Update the API to parse for run param and call driver.UpdateTask accordingly
 - Update the API client to pass in the run param and return plan
 - Reorganize imports (from feedback https://github.com/hashicorp/consul-terraform-sync/pull/191#pullrequestreview-592717549)

### DIFF
--- a/api/task_test.go
+++ b/api/task_test.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"encoding/json"
 	"errors"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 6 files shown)


findkim (approved):
Changes make sense to me 👍

api/task_test.go
```diff
@@ -203,3 +236,64 @@ func TestTask_decodeJSON(t *testing.T) {
...
+func TestTask_RunOption(t *testing.T) {
+	cases := []struct {
+		name        string
+		path        string
+		option      string
+		expectError bool
+	}{
+		{
+			"happy path run now",
+			"/v1/tasks?run=now",
```
findkim, line 249: Are the run options available for `/v1/tasks`? or `/v1/tasks/:task`?

lornasong: Thanks for catching this! It should be `/v1/tasks/:task`. Let me update!

PR #198, opened 2021-02-19
Support enable task CLI command

lornasong:
Example:

```
./consul-terraform-sync task enable task_disabled
==> Inspecting changes to resource if enabling 'task_disabled'...

    Generating plan that CTS will use Terraform to execute

    
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create

  // <resource diff. truncated>

Plan: 3 to add, 0 to change, 0 to destroy.

==> Enabling the task will perform the actions described above.
    Do you want to perform these actions for 'task_disabled'?
     - CTS cannot guarantee Terraform will perform these exact actions if
       monitored services have changed.

Enter a value: yes

==> Enabling and running 'task_disabled'...

==> 'task_disabled' enable complete!
```

Example if no resources are expected to change:
```
./consul-terraform-sync task enable task_b
==> Inspecting changes to resource if enabling 'task_b'...

    Generating plan that CTS will use Terraform to execute

    
No changes. Infrastructure is up-to-date.

This means that Terraform did not detect any differences between your
configuration and real physical resources that exist. As a result, no
actions need to be performed.

==> 'task_b' enable complete!
```


command/meta.go
```diff
@@ -76,3 +76,44 @@ func (m *meta) client() *api.Client {
...
+	return 0, true
+}
+
+// changesDetected attempts to detect if a change plan has been generated.
+// Does a few string parse checks to try to prevent a false negative
+func (m *meta) changesDetected(plan string) bool {
+	plan = strings.TrimSpace(plan)
+
+	// look for phrases that there are changes
+	if strings.HasPrefix(plan, "An execution plan has been generated") {
```
lornasong, line 110: Looking for these phrases is a little iffy. I have in my to-dos to write an e2e test to check terraform output for these phrases. If people have other ideas, let me know!

command/meta.go
```diff
@@ -76,3 +76,44 @@ func (m *meta) client() *api.Client {
...
 	}, nil)
 }
+
+// requestUserApproval returns an exit code and boolean describing if the user
+// approved. If the user did not approve (false is returned), exit code is provided.
+func (m *meta) requestUserApproval(taskName string) (int, bool) {
+	m.UI.Info("Enabling the task will perform the actions described above.")
+	m.UI.Output(fmt.Sprintf("Do you want to perform these actions for '%s'?", taskName))
+	m.UI.Output(" - This action cannot be undone.")
+	m.UI.Output(" - CTS cannot guarantee Terraform will perform these exact actions if")
```
findkim, line 86: CTS -> Consul Terraform Sync

CTS is fine for internal comments and colloquial usage, but since this is user-facing I think it would be best to expand to the full name (a few more cases below too).

eikenb:
The commit message for the first commit, 139cdc2, seems to have been cut off. 

 > UpdateTask will return the Terraform plan as is. The CLI

"The CLI" ...?

command/meta.go
findkim: I'm not sure if this would be accessible from this package, but it looks like the tf-exec package has a boolean to provide info on diffs

> The returned boolean is false when the plan diff is empty (no changes) and true when the plan diff is non-empty (changes present).

https://pkg.go.dev/github.com/hashicorp/terraform-exec@v0.13.0/tfexec#Terraform.Plan

eikenb:
I don't see the `-y|--yes` and `-n|--no` options in here to automate running these commands. 
Are they planned or already in a follow up PR?

eikenb:
Quick related thought... instead of the default being interactive, the default could be `no` where you pass `-y|--yes` for yes or `-i|--interactive` for the prompt. Defaults to not changing anything, just showing the plan. But this also doesn't block the TTY (terminal input/output) which, in the case of a script/automation, can make it hard to know what's going on.

Just a passing thought. Don't worry about it unless you really like the idea.

command/meta.go
lornasong: Ooo thanks for finding this. This seems way better than trying to parse for changes. Let me look into consuming the 'changes present' boolean.

command/meta.go
lornasong: Good point! I will change this to be the full name

lornasong:
@eikenb, thanks for the review and comments!

> The commit message for the first commit, 139cdc2, seems to have been cut off.

🙈 I'm not sure how that got in there. The commit message should only be "Update enable CLI to include -port flag changes". Let me clean that up. Thanks for flagging!

> I don't see the -y|--yes and -n|--no options in here to automate running these commands.
> Are they planned or already in a follow up PR?

Yes! They are planned in sense that I agree we want to do them. I'll make tickets to capture them. However, they aren't planned for Beta.

> the default could be 'no' where you pass -y|--yes for yes or -i|--interactive for the prompt

Interesting idea! Thanks for sharing it! My main hesitation with it is that our other products (Terraform in particular) have default interactive mode. I think users might assume CTS will follow this precedent and use non-interactive mode by accident.

eikenb:
> My main hesitation with it is that our other products (Terraform in particular) have default interactive mode.

If there's internal precedent for this pattern, it is probably better to stay with that. Particularly with Terraform and CTS, as they will have some overlap in users and keeping things consistent is good.

eikenb (approved):
My comments have all been addressed. :+1:

command/meta.go
lornasong: The work to consume the 'changes present' boolean got a bit large so I opened a separate PR: https://github.com/hashicorp/consul-terraform-sync/pull/200. I'll rebase this PR off of those changes once they are merged.

lornasong: Rebased! Thanks for this finding @findkim !

command/meta.go
findkim: There are two more occurrences @lornasong then 👍

lornasong: Thanks for pointing that out! Ok! I've pushed a commit to catch the other two occurrences.

findkim (approved):

PR #200, opened 2021-02-19
Modify driver.UpdateTask to return whether changes are present for inspect runs

lornasong:
Enable CLI needs to determine if changes are present after generating an inspect
plan using driver.UpdateTask. It is difficult to determine from the plan text if
changes are actually present. tf-exec's Plan() returns a boolean if changes
are present. It is better to pass this boolean to the CLI rather than attempt
to parse the plan text.

From discussion: https://github.com/hashicorp/consul-terraform-sync/pull/198#discussion_r579463821

Changes
 - Update terraform client to return changes-present boolean
 - Update driver.UpdateTask to return InspectPlan{} struct rather than plan string


findkim (approved):
The ChangesPresent looks good 👍

PR #201, opened 2021-02-19
Change API error from string to object

findkim:
**Breaking change**: This will give room for future details to be added to the API error responses.

Prior to change
```
{
  "error": "unsupported status parameter value. only supporting status values successful, errored, critical, and unknown but got xxx"
}
```

After change
```
{
  "error": {
    "message": "unsupported status parameter value. only supporting status values successful, errored, critical, and unknown but got XXX"
  }
}
```


lornasong (approved):
the change is so clean and minimal ✨

PR #202, opened 2021-02-22
Update Task Status API to support enabled/disabled tasks

lornasong:
Example of updated return payload:
```
  "task_a": {
    "task_name": "task_a",
    "status": "successful",
    "enabled": false, // new field
	...
  }
```

Commits:
 - Add driver to taskStatusHandler and use driver to add `enabled` information
 to task status payload
 - Support 'unknown' status for tasks that are disabled from the start (since
 these tasks have never run, there is no event data  collected on them)
 - Fix issue in controller where events were being stored even though tasks were
 disabled

### DIFF
--- a/api/task_status.go
+++ b/api/task_status.go
@@ -75,7 +75,7 @@ func (h *taskStatusHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			jsonErrorResponse(w, http.StatusNotFound, err)
 			return
 		}
-		status := makeTaskStatus(taskName, events, d.Task(), h.version)
+		status := makeTaskStatus(events, d.Task(), h.version)
 
 		if filter != "" && status.Status != filter {
 			continue
@@ -110,7 +110,7 @@ func (h *taskStatusHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 }
 
 // makeTaskStatus takes event data for a task and returns a task status
-func makeTaskStatus(taskName string, events []event.Event, task driver.Task,
+func makeTaskStatus(events []event.Event, task driver.Task,
 	version string) TaskStatus {
 
 	successes := make([]bool, len(events))
@@ -131,12 +131,12 @@ func makeTaskStatus(taskName string, events []event.Event, task driver.Task,
 	}
 
 	return TaskStatus{
-		TaskName:  taskName,
+		TaskName:  task.Name,
 		Status:    successToStatus(successes),
 		Enabled:   task.Enabled,
 		Providers: mapKeyToArray(uniqProviders),
 		Services:  mapKeyToArray(uniqServices),
-		EventsURL: makeEventsURL(events, version, taskName),
+		EventsURL: makeEventsURL(events, version, task.Name),
 	}
 }
 
--- a/api/task_status.go
+++ b/api/task_status.go
@@ -88,10 +88,30 @@ func (h *taskStatusHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		statuses[taskName] = status
 	}
 
+	// if user requested a specific task that does not have events, check if
+	// a driver exists
+	if taskName != "" {
+		if _, ok := data[taskName]; !ok {
+			if d, ok := h.drivers[taskName]; ok {
+				statuses[taskName] = makeTaskStatusUnknown(d.Task())
+			}
+		}
+	}
+
+	// if user requested all tasks and status filter applicable, check driver
+	// for tasks without events
+	if taskName == "" && (filter == "" || filter == StatusUnknown) {
+		for tN, d := range h.drivers {
+			if _, ok := data[tN]; !ok {
+				statuses[tN] = makeTaskStatusUnknown(d.Task())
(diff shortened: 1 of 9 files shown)


api/task_status.go
```diff
@@ -64,7 +68,17 @@ func (h *taskStatusHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	statuses := make(map[string]TaskStatus)
 	for taskName, events := range data {
-		status := makeTaskStatus(taskName, events, h.version)
+		d, ok := h.drivers[taskName]
+		if !ok {
+			err := fmt.Errorf("task '%s' does not exist", taskName)
+			log.Printf("[TRACE] (api.updatetask) %s", err)
+			jsonResponse(w, http.StatusNotFound, map[string]string{
+				"error": err.Error(),
+			})
```
findkim, line 77:
```suggestion
			jsonErrorResponse(w, http.StatusNotFound, err)
```

api/task_status.go
```diff
@@ -74,11 +88,33 @@ func (h *taskStatusHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		}
+	}
+
 	jsonResponse(w, http.StatusOK, statuses)
 }
 
-// makeTaskStatus takes event data for a task and returns an overall task status
-func makeTaskStatus(taskName string, events []event.Event, version string) TaskStatus {
+// makeTaskStatus takes event data for a task and returns a task status
+func makeTaskStatus(taskName string, events []event.Event, task driver.Task,
```
findkim, line 115: the `task driver.Task` has the name, and the caller of this function should already ensure that `taskName == driver.Task.Name`, so an option could be to remove `taskName` parameter to condense the function signature.

api/task_status_test.go
```diff
@@ -168,30 +227,15 @@ func TestTaskStatus_ServeHTTP(t *testing.T) {
...
-			map[string]TaskStatus{
-				"task_nonexistent": TaskStatus{
-					TaskName:  "task_nonexistent",
-					Status:    StatusUnknown,
-					Providers: []string{},
-					Services:  []string{},
-					EventsURL: "",
-				},
-			},
+			// http.StatusOK,
```
findkim, line 230:
```suggestion
```

e2e/api_test.go
```diff
@@ -175,6 +189,14 @@ func TestE2E_StatusEndpoints(t *testing.T) {
 
 			checkEvents(t, taskStatuses, fakeFailureTaskName, tc.expectFailureTask)
 			checkEvents(t, taskStatuses, fakeSuccessTaskName, tc.expectSuccessTask)
+
+			task, ok := taskStatuses[disabledTaskName]
+			if tc.expectDisabledTask {
+				require.True(t, ok)
+				assert.Nil(t, task.Events)
+			} else {
+				require.False(t, ok)
```
findkim, line 198: Both of the `require` reads to me that they should be `assert` because it's a part of the testing expected values instead of test setup.

e2e/api_test.go
```diff
@@ -118,7 +125,11 @@ func TestE2E_StatusEndpoints(t *testing.T) {
 			require.NoError(t, err)
 			defer resp.Body.Close()
 
-			require.Equal(t, resp.StatusCode, http.StatusOK)
+			require.Equal(t, tc.statusCode, resp.StatusCode)
+
+			if tc.statusCode != http.StatusOK {
+				return
+			}
```
findkim, line 132: Prior, only 200 was accepted and `require` made sense. Now that the status code is a value to test for, I would suggest changing it `require` to `assert`.

api/task_status.go
```diff
@@ -24,13 +26,15 @@ type TaskStatus struct {
 // taskStatusHandler handles the task status endpoint
 type taskStatusHandler struct {
 	store   *event.Store
+	drivers map[string]driver.Driver
```
findkim, line 29: Keeping track of where this map is used -- now a third obj sharing view of the same map.

findkim (approved):
Thanks for including the example payload. The status API changes look good. Only had a few minor suggestions that are non-blocking.

api/task_status.go
lornasong: :nod: and there's fourth object (overall status API) down the line. I may go ahead and refactor this after Beta (thinking about the code freeze) unless you think it should be done beforehand.

api/task_status.go
lornasong: Thank you for catching!

api/task_status.go
lornasong: I like that! Thanks for the suggestion.

api/task_status_test.go
lornasong: Thank you for catching this!

e2e/api_test.go
lornasong: I agree, thanks for pointing that out!

e2e/api_test.go
lornasong: Yes, agree! Thank you for catching this :)

PR #203, opened 2021-02-22
Update Overall Status API to support enabled/disabled tasks

lornasong:
**Breaking change:** Reorganize the returned status payload to allow for
different breakdowns of task summary information

Example of new payload:
```
{
  "task_summary": {
    "status": { // move status info into this status sub-struct
      "successful": 2,
      "errored": 1,
      "critical": 0,
      "unknown": 0 // new field for disabled tasks with no event data
    },
    "enabled": { // add new enabled information
      "true": 2,
      "false": 1
    }
  }
}
```

Commits:
 - Add driver to overallStatusHandler and use driver to add enabled information
to overall status payload. Refactor task summary to have status and enabled
sub-structures
 - Refactor some testing methods
 - Support 'unknown' status for tasks that are disabled from the start (since
these tasks have never run, there is no event data collected on them)

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -68,7 +68,7 @@ func NewAPI(store *event.Store, drivers map[string]driver.Driver, port int) *API
 
 	// retrieve overall status
 	mux.Handle(fmt.Sprintf("/%s/%s", defaultAPIVersion, overallStatusPath),
-		newOverallStatusHandler(store, defaultAPIVersion))
+		newOverallStatusHandler(store, drivers, defaultAPIVersion))
 	// retrieve task status for a task-name
 	mux.Handle(fmt.Sprintf("/%s/%s/", defaultAPIVersion, taskStatusPath),
 		newTaskStatusHandler(store, drivers, defaultAPIVersion))
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -240,9 +240,15 @@ func TestJsonResponse(t *testing.T) {
 			http.StatusOK,
 			OverallStatus{
 				TaskSummary: TaskSummary{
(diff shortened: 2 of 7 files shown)


api/api.go
```diff
@@ -68,7 +68,7 @@ func NewAPI(store *event.Store, drivers map[string]driver.Driver, port int) *API
 
 	// retrieve overall status
 	mux.Handle(fmt.Sprintf("/%s/%s", defaultAPIVersion, overallStatusPath),
-		newOverallStatusHandler(store, defaultAPIVersion))
+		newOverallStatusHandler(store, drivers, defaultAPIVersion))
```
findkim, line 71: 4!

findkim (approved):

api/api.go
lornasong: :nod-intensifies:

eikenb: ???

lornasong: Sorry for the missing context! This is the 4th place where I've passed in 'drivers' which is a `map[string]driver` and has the danger of being modified. We want to refactor the map into something an actual `Drivers{}` object. I'm currently planning to do this after Beta to respect the code freeze and testing. Open to feedback on the timing!

findkim: This is in context `map[string]drivers` across the recent series of PRs

https://github.com/hashicorp/consul-terraform-sync/pull/190#discussion_r578003640
https://github.com/hashicorp/consul-terraform-sync/pull/202#discussion_r580399453

eikenb: Thanks!

PR #204, opened 2021-02-22
Add e2e tests for Update Task API

lornasong:
Commits:
 - Reuse TestE2EDisabledTask for Update Task API e2e test due to overlaps in
 checks. Expand on disabled task to enable then disable and do checks
 - Refactor functions that run CTS

### DIFF
--- a/e2e/api_test.go
+++ b/e2e/api_test.go
@@ -263,7 +263,7 @@ func TestE2E_TaskEndpoints_UpdateEnableDisable(t *testing.T) {
 	err = makeConfig(configPath, disabledTaskConfig(srv.HTTPAddr, tempDir, port))
 	require.NoError(t, err)
 
-	cmd, err := runSyncLongMode(configPath)
+	cmd, err := runSync(configPath)
 	defer stopCommand(cmd)
 	require.NoError(t, err)
 	time.Sleep(5 * time.Second)
@@ -343,8 +343,7 @@ func runSyncDevMode(configPath string) (*exec.Cmd, error) {
 	return cmd, nil
 }
 
-// runSyncLongMode - placeholder name
-func runSyncLongMode(configPath string) (*exec.Cmd, error) {
+func runSync(configPath string) (*exec.Cmd, error) {
 	cmd := exec.Command("consul-terraform-sync",
 		fmt.Sprintf("--config-file=%s", configPath))
 	cmd.Stdout = os.Stdout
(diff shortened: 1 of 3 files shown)


e2e/api_test.go
```diff
@@ -242,6 +244,92 @@ func TestE2E_StatusEndpoints(t *testing.T) {
...
+	t.Parallel()
+
+	srv, err := newTestConsulServer(t)
+	require.NoError(t, err, "failed to start consul server")
+	defer srv.Stop()
+
+	tempDir := fmt.Sprintf("%s%s", tempDirPrefix, "disabled_task")
+	delete, err := testutils.MakeTempDir(tempDir)
+	// no defer to delete directory: only delete at end of test if no errors
+	require.NoError(t, err)
```
findkim, line 257: Seeing how you refactored `apiRequest` to check errors there, it would be a neat test utility for this function too!

e2e/api_test.go
```diff
@@ -242,6 +244,92 @@ func TestE2E_StatusEndpoints(t *testing.T) {
 	delete()
 }
 
+func TestE2E_TaskEndpoints_UpdateEnableDisable(t *testing.T) {
```
findkim, line 247: Just checking if I'm understanding the test correctly to make sure I'm reviewing with the right context.

`TestE2E_TaskEndpoints_UpdateEnableDisable` replaces the need for `TestE2EDisabledTask`.
1. The test starts off with a disabled task
2. API to enable it
3. API to disable it then trigger a task run by adding a new service instance to very no runs because it's disabled.

findkim (approved):

e2e/api_test.go
lornasong: Ooo, good idea! I have on my to-do list to revisit `testutils` and move some functions there (registering consul services, even apiRequest can probably benefit from a move). I will add this to my list! I'm planning to do this after we release Beta.

e2e/api_test.go
lornasong: Yes! Your understanding is correct. `TestE2EDisabledTask` does step 1. `TestE2E_TaskEndpoints_UpdateEnableDisable` needs to start off with step 1, so I combined the two tests. Let me add a comment at the top of the test with a summary similar to what you have here.

findkim: The test summary comment is 💯 thank you!

PR #206, opened 2021-02-22
Task update API to be consistent JSON payload with snake_case

findkim:
All other API responses are snake case. Instead of creating a dedicated API object that is 1:1 to the internal object, will just added JSON struct tags as a means to have the Update Task API to be consistent with snake case.

```
$ curl localhost:8558/v1/tasks/a -X PATCH -d '{"enabled":true}' | jq .
```

```diff
{
-  "ChangesPresent": false,
-  "Plan": ""
+  "changes_present": false,
+  "plan": ""
}
```


lornasong (approved):
thanks for catching 🙏

PR #208, opened 2021-02-22
Add e2e tests for Disable and Enable Task CLIs

lornasong:
Changes
 - Add test cases for cross subcommand errors that are defined in the meta object
 - Test enable and disable task CLIs
 - Mild refactor oneTaskConfig so that a port can be passed in

### DIFF
--- a/e2e/command_test.go
+++ b/e2e/command_test.go
@@ -1,0 +1,233 @@
+// +build e2e
+
+package e2e
+
+import (
+	"bytes"
+	"fmt"
+	"os/exec"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/hashicorp/consul-terraform-sync/api"
+	"github.com/hashicorp/consul-terraform-sync/testutils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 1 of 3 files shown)


e2e/command_test.go
```diff
@@ -0,0 +1,233 @@
...
+	port, err := api.FreePort()
+	require.NoError(t, err)
+
+	err = makeConfig(configPath, disabledTaskConfig(srv.HTTPAddr, tempDir, port))
+	require.NoError(t, err)
+
+	cmd, err := runSyncDevMode(configPath)
+	defer stopCommand(cmd)
+	require.NoError(t, err)
+	time.Sleep(5 * time.Second)
```
findkim, line 112: For API-dependent e2e tests, I can see a future improvement to have the CTS client support a health endpoint to verify that the CLI is good to interact with API server.

findkim (approved):
The new e2e tests are easy to follow

e2e/command_test.go
lornasong: Great idea! I'd love that compared to these arbitrary sleeps

findkim: It'll still poll/sleep probably but yeah a bit less arbitrary 🙂

PR #210, opened 2021-02-23
update all remaining modules

eikenb:
An update of consul/sdk/testutis eliminated the test server setup call
that didn't have a testing.TestingTB passed to it so I added a dummy
version to testutils/utils.go to use in those cases where needed.


lornasong (approved):
looks good to me!

PR #213, opened 2021-02-23
Use consul configuration for the default consul backend for Terraform

findkim:
The default Terraform backend when ran by CTS is the Consul KV. Users would configure the Consul block in CTS configuration and assume the same configuration would be used for the Terraform Consul backend. However there were two edge cases not handled for configuring the default Consul backend.
1. Configuration values in Consul block for CTS configuration file were not passed to the Consul client used by Terraform backend (i.e. the `backend "consul" {}` block in Terraform `main.tf`).
2. Configuration values for Consul through ENV were not passed to the Consul client used by Terraform backend if a provider `task_env` was set because it no longer consumed from the os environment ([tfexec docs](https://pkg.go.dev/github.com/hashicorp/terraform-exec@v0.13.0/tfexec#Terraform.SetEnv)).

The proposed changes convert Consul configuration to the supported ENV to set for the task's Terraform workspace. These changes do not write Consul configuration values to any Terraform files unless explicitly configured within the Terraform [backend block](https://www.consul.io/docs/nia/configuration#backend).

### DIFF
--- a/driver/terraform.go
+++ b/driver/terraform.go
@@ -96,7 +96,7 @@ func NewTerraform(config *TerraformConfig) (*Terraform, error) {
 		log.Printf("[ERR] (driver.terraform) init client type '%s' error: %s", config.ClientType, err)
 		return nil, err
 	}
-	if env := config.Task.Providers.Env(); len(env) > 0 {
+	if env := config.Task.Env; len(env) > 0 {
 		tfClient.SetEnv(env)
 	}
 
@@ -171,7 +171,7 @@ func (tf *Terraform) SetBufferPeriod() {
 		return
 	}
 
-	log.Printf("[TRACE] (driver.terraform) set buffer period for '%s': %v",
+	log.Printf("[TRACE] (driver.terraform) set buffer period for '%s': %+v",
 		taskName, bp)
 	tf.watcher.SetBufferPeriod(bp.Min, bp.Max, tf.template.ID())
 }
--- a/driver/terraform.go
+++ b/driver/terraform.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 
 	"github.com/hashicorp/consul-terraform-sync/client"
@@ -96,7 +97,17 @@ func NewTerraform(config *TerraformConfig) (*Terraform, error) {
 		log.Printf("[ERR] (driver.terraform) init client type '%s' error: %s", config.ClientType, err)
 		return nil, err
 	}
-	if env := config.Task.Env; len(env) > 0 {
+
+	if taskEnv := config.Task.Env; len(taskEnv) > 0 {
+		// Terraform init requires discovering git in the PATH env.
+		//
+		// The terraform-exec package disables inheriting from the os environment
+		// when using tfexec.SetEnv(). So for CTS purposes, we'll force inheritance
+		// to allow Terraform commands to use the os environment as necessary.
+		env := envMap(os.Environ())
+		for k, v := range taskEnv {
+			env[k] = v
+		}
 		tfClient.SetEnv(env)
 	}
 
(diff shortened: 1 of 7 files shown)


driver/terraform.go
```diff
@@ -96,7 +96,7 @@ func NewTerraform(config *TerraformConfig) (*Terraform, error) {
 		log.Printf("[ERR] (driver.terraform) init client type '%s' error: %s", config.ClientType, err)
 		return nil, err
 	}
-	if env := config.Task.Providers.Env(); len(env) > 0 {
+	if env := config.Task.Env; len(env) > 0 {
```
eikenb, line 99: Not sure if there is anything wrong here, but wanted to draw attention to it in case. 
Do the env settings from `config.Task.Providers.Env()` just not matter here anymore?

lornasong (commented):
Just wanted to double-check I understand how edge-case 1 is being solved. I'm understanding that there won't be any Consul backend information in the Terraform files (per your comment "These changes do not write Consul configuration values to any Terraform files unless explicitly configured within the Terraform backend block") and that the Consul backend information is set by the tf-exec environment variable. A two edge-cases one solution. Is that right?

driver/terraform.go
findkim: The `providers.Env()` is merged by the caller in the changes above within `controller/controller.go`.

However after running into a different but where tfexec doesn't inherit the os env which includes `PATH` used for `git` within Terraform init, I'll be moving some of the merging around now that there are 3 environments: task, providers, and os.

findkim:
@lornasong yep! two edge-cases one solution. For case 1 it won't be written to file but instead the consul configuration from the CTS config file would be converted and passed as ENV for the Terraform commands.

driver/terraform.go
eikenb: Oh... I think I see it now on line 242 above. Cool. 
Sorry for the red herring.

findkim:
Thank for the initial review y'all! After running this with the default path using no custom environments I ran into a bug with how the environment is handled. The error below:

```
2021/02/23 21:59:38.936734 [INFO] running Terraform command: /Users/kngo/dev/hashicorp/consul-terraform-sync/terraform init -no-color -force-copy -input=false -lock-timeout=0s -backend=true -get=true -get-plugins=true -lock=true -upgrade=false -verify-plugins=true
Initializing modules...
Downloading findkim/print/cts 0.1.0 for print...

Error: Failed to download module

Could not download module "print" (main.tf:21) source code from
"git::https://github.com/findkim/terraform-cts-print?ref=v0.1.0": error
downloading 'https://github.com/findkim/terraform-cts-print?ref=v0.1.0': git
must be available and on the PATH.
```

I updated the PR to always inherit from the OS environment so Terraform has access to `PATH` where it can discover `git` used to download modules from the Terraform registry.

I kept the `os.Environ` logic near `tfexec.SetEnv()` to keep the context as to why this is necessary closer together. And I made the logic for building the task environment more clear within `controller/controller.go` to merge the Consul config and task's providers.

driver/terraform.go
```diff
@@ -96,7 +97,17 @@ func NewTerraform(config *TerraformConfig) (*Terraform, error) {
...
+	if taskEnv := config.Task.Env; len(taskEnv) > 0 {
+		// Terraform init requires discovering git in the PATH env.
+		//
+		// The terraform-exec package disables inheriting from the os environment
+		// when using tfexec.SetEnv(). So for CTS purposes, we'll force inheritance
+		// to allow Terraform commands to use the os environment as necessary.
+		env := envMap(os.Environ())
+		for k, v := range taskEnv {
+			env[k] = v
+		}
```
lornasong, line 110: Helpful comment! Also personally like inheriting os env vars close to the tfClient.SetEnv

lornasong (approved):
Thanks for the clarification. This looks good to me! Good catch finding this issue!

driver/terraform.go
findkim: That's good to hear, thanks.

I also found that if this was done in the controller, it would have made the tests a bit more tedious to maintain. https://app.circleci.com/pipelines/github/hashicorp/consul-terraform-sync/711/workflows/f577eac6-6266-432d-8559-ff85aa75fa1c/jobs/1380/parallel-runs/0/steps/0-104?invite=true

eikenb: Good call on keeping code oriented toward being testable. :100:

PR #214, opened 2021-02-24
Modify the Update Task API response when returning inspect

lornasong:
While writing the API docs for Update Task, I found that the returned response didn't make sense to me for all cases

PR changes the response so that the returned driver.InspectPlan is wrapped in an 'inspect' object, and it is only returned when `?run=inspect` is requested.

For other/no run options, return an empty 'inspect' object. Previously, an empty driver.InspectPlan was returned when `?run=inspected` was _not_ requested. The empty values are confusing. In particular, `changes_present` would be false for `?run=now` even though changes might have actually been present.

Example payload when `?run=inspect`
```
{
  "inspect": {
    "changes_present": true,
    "plan": "<Terraform plan output>"
  }
}
```

Payload when no run option or `?run=now`
```
{}
```

Prior to the change, payload when payload when no run option or `?run=now`:
```
{
   "changes_present": false,
   "plan": ""
}
```

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -252,6 +252,16 @@ func TestJsonResponse(t *testing.T) {
 				},
 			},
 		},
+		{
+			"update task inspect",
+			http.StatusOK,
+			UpdateTaskResponse{
+				Inspect: &driver.InspectPlan{
+					ChangesPresent: true,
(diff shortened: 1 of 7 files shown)


findkim (approved):
Good catch

PR #215, opened 2021-02-25
initial dockerfile and entrypoint

eikenb:


docker/Dockerfile
```diff
@@ -0,0 +1,73 @@
+# This Dockerfile creates a production release image for the project. This
+# downloads the release from releases.hashicorp.com and therefore requires that
+# the release is published before building the Docker image.
+#
+# We don't rebuild the software because we want the exact checksums and
+# binary signatures to match the software and our builds aren't fully
+# reproducible currently.
+ARG BASE_ARCH
```
alvin-huang, line 8: We don't need these two lines here anymore since we are using `docker buildx` in the release process rather than building 4 architecture images and then using `docker manifest`.

docker/Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+FROM ${BASE_ARCH}/alpine:3.11
```
alvin-huang, line 9:
```suggestion
FROM alpine:3.13
```

docker/Dockerfile
eikenb: :+1:

docker/Dockerfile
eikenb: Just pushed the commit that changes this.

alvin-huang (approved):
LGTM!

docker/Dockerfile
eikenb: Thanks!

PR #216, opened 2021-02-26
fix docker permissions with a 'home' directory

eikenb:
Changed /consul-terraform-syn directory to be owned by the added user of
the same name and made it the working directory. This allows writing to
the current/working directory which is needed for the sync-tasks
directory to be created and for terraform to be downloaded locally (both
the defaults).

Note that setting the terraform config variables 'path' or 'working_dir'
to something other than "" will not work.


lornasong (approved):
looks good to me!

PR #218, opened 2021-03-04
Refactor map of drivers to a Drivers object

lornasong:
A map of drivers is shared across the API handlers and read-write controller
to manage task lifecycle. This map of drivers is currently used with read-access
only. However, we will want to support write-access in order to support a
create task cli and delete task cli.

Drivers structure will wrap the map of drivers in order to manage concurrent
read/write access to the map.

Changes:
 - Add driver.Drivers{} to replace map[string]driver.Driver
 - New methods to support currently required functionality NewDrivers(),
 drivers.Add(), drivers.Get(), drivers.Map()

### DIFF
--- a/driver/drivers.go
+++ b/driver/drivers.go
@@ -61,7 +61,7 @@ func (d *Drivers) Map() map[string]Driver {
 	d.mu.RLock()
 	defer d.mu.RUnlock()
 
-	copy := make(map[string]Driver)
+	copy := make(map[string]Driver, len(d.drivers))
 	for k, v := range d.drivers {
 		copy[k] = v
 	}
--- a/driver/drivers.go
+++ b/driver/drivers.go
@@ -1,0 +1,69 @@
+package driver
+
+import (
+	"errors"
+	"fmt"
+	"sync"
+)
+
+// Drivers wraps the map of task-name to associated driver so that the map
+// can be accessed concurrently
+type Drivers struct {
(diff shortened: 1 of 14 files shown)


driver/drivers.go
```diff
@@ -0,0 +1,69 @@
...
+
+	return driver, true
+}
+
+// Map returns a copy of the map containing the drivers
+func (d *Drivers) Map() map[string]Driver {
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+
+	copy := make(map[string]Driver)
```
eikenb, line 64: As you have the size of the map, it is (slightly) more efficient to pre-allocate the map here.

eikenb (approved):
When I saw previous discussions on this my first thought was to refactor the `map[string]Driver` into it's own type (eg. `type Drivers map[string]Driver`). But I think after working on it a bit I would very likely have ended up in this same place. 

TLDR; :+1: on the design and, aside from my one tiny comment, it looks good.

driver/drivers.go
lornasong: Ah, good point. Thank you!

lornasong:
@eikenb thank you for the review! Yes, actually I was also originally thinking of `type Drivers map[string]Driver` implementation and had actually done this first. I switched to a struct though so that I could add the lock and have more explicit read/write access to the map.

PR #219, opened 2021-03-05
Fix bug in Task Status API response when requesting task with no events

lornasong:
Previously when requesting task status for a task with a driver and no events (e.g. a task configured as disabled), the API would return:

```
TaskStatus{
	TaskName:  "task_d",
	Status:    StatusUnknown,
	Enabled:   false,
	Providers: []string{}, // this data is missing
	Services:  []string{}, // this data is missing
	EventsURL: "",
},
```

It should return:
```
TaskStatus{
	TaskName:  "task_d",
	Status:    StatusUnknown,
	Enabled:   false,
	Providers: []string{"null"},
	Services:  []string{"local"},
	EventsURL: "",
},
```
See test for more details.

Commits:
 - Fix how store.Read() returns events for tasks without events. This was causing tasks without events to be handled similarly as tasks with events. Therefore code handling tasks without events was unreachable
 - After fixing store.Read, update code handling tasks without events (now that it is reachable) to handle when a task has no events and no driver

### DIFF
--- a/api/task_status.go
+++ b/api/task_status.go
@@ -70,7 +70,7 @@ func (h *taskStatusHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	for taskName, events := range data {
 		d, ok := h.drivers.Get(taskName)
 		if !ok {
-			err := fmt.Errorf("task '%s' does not exist", taskName)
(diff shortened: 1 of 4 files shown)


eikenb (approved):
Looks good.

PR #220, opened 2021-03-09
Add tests for some missing coverage from enable/disable cli development

lornasong:
Commits
 - Removed some unreachable code found while looking at coverage
 - Add a few missing tests

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -2,7 +2,6 @@ package api
 
 import (
 	"encoding/json"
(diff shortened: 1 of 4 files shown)


findkim (approved):
The `api.Client.request` is a nifty method you wrote that manages the repetitive code cleaned up here ✨

eikenb (approved):
:grim-reaper-of-code:

PR #221, opened 2021-03-10
Fix "missing config flag option" when running docker container in daemon-mode

lornasong:
Currently when running CTS docker container in once-mode or inspect-mode,
everything runs great. When running the container in daemon-mode (without any
flags), the configuration files are not passed in. In other modes (with flags), the
configuration file is successfully passed in.

The following error is seen when running`$ docker run -ti -v /my/path/to/config/file:/consul-terraform-sync/config hashicorp/consul-terraform-sync:0.1.0-beta`
> [ERR] config file(s) required, use --config-dir or --config-file flag options...

Change: update docker entrypoint to also pass config-directory when no CTS flags
are passed. With this change, I was able to build the docker image locally and test
daemon-mode, version, once-mode successfully

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -16,7 +16,7 @@ fi
 
 # If user is trying to run consul-terraform-sync with no arguments (daemon-mode),
 # then $1 will default to '/bin/sh'. In this case, do not pass along default arguments.
-if [ "$1" = '/bin/sh' ]
+if [ "$*" = '/bin/sh -c /bin/${NAME}' ]
 then
     set -- /bin/consul-terraform-sync
 fi
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -14,6 +14,13 @@ then
     set -- /bin/consul-terraform-sync "$@"
 fi
 
+# If user is trying to run consul-terraform-sync with no arguments (daemon-mode),
+# then $1 will default to '/bin/sh'. In this case, do not pass along default arguments.
+if [ "$1" = '/bin/sh' ]
+then
+    set -- /bin/consul-terraform-sync
+fi
+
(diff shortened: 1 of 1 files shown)


lornasong:
Hey team, my bash knowledge is minimal. Welcome any feedback on a better way to accomplish this!

findkim (approved):
I'm also not well versed in bash and rusty on docker, but the changes make sense going off of the summary and explanation of the bug.

eikenb:
Please hold off on merging this as it might have an issue, but I need to test it first and am in the middle of something.

docker/docker-entrypoint.sh
```diff
@@ -14,6 +14,13 @@ then
     set -- /bin/consul-terraform-sync "$@"
 fi
 
+# If user is trying to run consul-terraform-sync with no arguments (daemon-mode),
+# then $1 will default to '/bin/sh'. In this case, do not pass along default arguments.
+if [ "$1" = '/bin/sh' ]
```
eikenb, line 19: This should check for more than `$1` equating `/bin/sh` as it is a very common pattern to run the container with `/bin/sh` in order to debug issues. For example..

```
docker run -ti --rm --network host hashicorp/consul-terraform-sync:0.1.0-beta /bin/sh
```

So you don't want to test for `/bin/sh`.. at least by itself. The full command when the docker container is run without arguments is `/bin/sh -c /bin/${NAME}`.  So I suggest a matching the full string here..

```
if test "$*" = '/bin/sh -c /bin/${NAME}'
then ...
```

eikenb: Note I tweaked the above several times... so don't go off the email version.

lornasong: Thanks for catching that and sharing that common use-case! That's super helpful to know. I committed your suggested change and updated the comment as well.

eikenb (approved):
:+1:

PR #222, opened 2021-03-11
Fix service tag erroring on '='

findkim:
Filtering service tag that contained `=` was supported in techpreview2 and broke in beta.

tags.hcl config
```hcl
service {
  name = "my-app"
  tag = "my=tag"
}

task {
  name = "test"
  services = ["my-app"]
  source = "findkim/print/cts"
}
```

```
$ consul-terraform-sync -inspect -config-file ./configs/tags.hcl
2021/03/11 16:58:19.758458 [ERROR] (driver.terraform) checking dependency changes for 'test': execute: template: 3117340e9c3d7ed273cb6efc04266975:11:17: executing "3117340e9c3d7ed273cb6efc04266975" at <service "my-app" "\"my=tag\" in Service.Tags">: error calling service: health.service: invalid query parameter: "\"my=tag\" in Service.Tags" for "my-app"
```

Pulls in changes from https://github.com/hashicorp/hcat/pull/38 that fix the bug

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,6 @@ require (
 	github.com/hashicorp/consul v1.9.3
 	github.com/hashicorp/consul/sdk v0.7.0
 	github.com/hashicorp/errwrap v1.1.0 // indirect
-	github.com/hashicorp/go-bexpr v0.1.7 // indirect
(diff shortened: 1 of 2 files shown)


lornasong (approved):
makes sense!

PR #228, opened 2021-03-18
move cli.go into commands package

eikenb:
This lets us import the cli and use it to run cts similar to the e2e
tests, runs the full app, but keeping it in process. This is
particularly useful for benchmarking, so we can profile the benchmarks.


lornasong (approved):
makes sense to me!

PR #229, opened 2021-03-18
use passed in args instead of os.Args

eikenb:
cli.Run() takes the cli argument list as its argument... but then
ignores it and uses os.Args (which is what is passed from the call in
main).

Using the passed in arguments is required for testing.


lornasong (approved):
Nice catch! Makes sense!

PR #230, opened 2021-03-19
make api-tooling-core team the default reviewer

eikenb:
Documentation on how this works.
https://help.github.com/articles/about-codeowners/


findkim (approved):
This is fine for now. If/when we reach a point we have a lot of contributors this may get very noisy and can reconsider then.

PR #232, opened 2021-03-19
add null provider services/instances benchmarks

eikenb:
Runs benchmarks using the Null provider and working with Consul setup
with multiple Services or Service Instances.

Currently uses HTTP connections as they are faster and we aren't using a
lot of connections.

### DIFF
--- a/e2e/benchmarks/services_test.go
+++ b/e2e/benchmarks/services_test.go
@@ -1,0 +1,279 @@
+// +build e2e
+
+// Runs benchmarks using the Null provider and working with Consul setup with
+// multiple Services or Service Instances.
+
+package benchmarks
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/hashicorp/consul-terraform-sync/command"
+	"github.com/hashicorp/consul-terraform-sync/testutils"
+	"github.com/hashicorp/consul/sdk/testutil"
+	"github.com/stretchr/testify/require"
+)
+
+const configFile = "config.hcl"
+const tempDirPrefix = "tmp_"
+
+// Benchmark N Services Instance (all the same service)
+func Benchmark1Instance(b *testing.B) {
+	benchmarkInstances(b, 1)
+}
+func Benchmark100Instances(b *testing.B) {
+	benchmarkInstances(b, 100)
+}
+func Benchmark1000Instances(b *testing.B) {
+	benchmarkInstances(b, 1000)
+}
+func benchmarkInstances(b *testing.B, N int) {
+	srv := testutils.NewTestConsulServerHTTPS(b, "../../testutils")
+	defer srv.Stop()
+	path, cleanup := setupServiceInstances(b, srv, N)
+	defer cleanup() // comment out if you want to keep the config files
+	//_ = cleanup
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		require.NoError(b, runSyncOnce(path))
+	}
+}
+
+// Benchmark N Services
+func Benchmark1Service(b *testing.B) {
+	benchmarkServices(b, 1)
+}
+func Benchmark100Service(b *testing.B) {
+	benchmarkServices(b, 100)
+}
+func Benchmark1000Service(b *testing.B) {
+	benchmarkServices(b, 1000)
+}
+func benchmarkServices(b *testing.B, N int) {
+	srv := testutils.NewTestConsulServerHTTPS(b, "../../testutils")
+	defer srv.Stop()
+	path, cleanup := setupMultiServices(b, srv, N)
+	defer cleanup() // comment out if you want to keep the config files
+	//_ = cleanup
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		require.NoError(b, runSyncOnce(path))
+	}
+}
+
+// ----------------------------------------------------------------------------
+
+// Run CTS in once mode from the cli level
+func runSyncOnce(configPath string) error {
+	cli := command.NewCLI(os.Stdout, os.Stderr)
+	args := []string{"cts", "-once", fmt.Sprintf("--config-file=%s", configPath)}
+	exitCode := cli.Run(args)
+	errstr := "Exit Code Error: %s (%d)\n"
+	switch exitCode {
+	case command.ExitCodeOK:
+		return nil
+	case command.ExitCodeError:
+		return fmt.Errorf(errstr, "ExitCodeError", exitCode)
+	case command.ExitCodeInterrupt:
+		return fmt.Errorf(errstr, "ExitCodeInterrupt", exitCode)
+	case command.ExitCodeRequiredFlagsError:
+		return fmt.Errorf(errstr, "ExitCodeRequiredFlagsError", exitCode)
+	case command.ExitCodeParseFlagsError:
+		return fmt.Errorf(errstr, "ExitCodeParseFlagsError", exitCode)
+	case command.ExitCodeConfigError:
+		return fmt.Errorf(errstr, "ExitCodeConfigError", exitCode)
+	case command.ExitCodeDriverError:
+		return fmt.Errorf(errstr, "ExitCodeDriverError", exitCode)
+	default:
+		return fmt.Errorf(errstr, "UNKNOWN", exitCode)
+	}
+}
+
+// Sets up consul and config to test setup with N service instances
+func setupServiceInstances(t testing.TB, srv *testutil.TestServer, N int) (string, func() error) {
+	services := make([]string, 1)
+	for _, s := range generateServices(N, true) {
+		testutils.RegisterConsulService(t, srv, s, testutil.HealthPassing)
+		services[0] = s.Name
+	}
+	return makeConfig(t, services, srv.HTTPAddr, false)
+}
+
+// add generated services as separte services
+func setupMultiServices(t testing.TB, srv *testutil.TestServer, N int) (string, func() error) {
+	services := make([]string, N)
+	for i, s := range generateServices(N, false) {
+		testutils.RegisterConsulService(t, srv, s, testutil.HealthPassing)
+		services[i] = s.Name
+	}
+	return makeConfig(t, services, srv.HTTPSAddr, true)
+}
+
+// generate service instance entries
+func generateServices(n int, instances bool) []testutil.TestService {
+	baseport := 30000
+	services := make([]testutil.TestService, n)
+	for i := 0; i < n; i++ {
+		name, id := "test_service", fmt.Sprintf("svc_%d", i)
+		if !instances {
+			name = id
+		}
+		services[i] = testutil.TestService{
+			Name:    name,
+			ID:      id,
+			Address: "127.0.0.2",
+			Port:    int(baseport + i),
+			Tags:    []string{},
+		}
+	}
+	return services
+}
+
+// writes the config files out to disk, returns path to them and cleanup func
+func makeConfig(t testing.TB, services []string, addr string, tls bool,
+) (string, func() error) {
+	tmpDir := fmt.Sprintf("%s%s", tempDirPrefix, "test_services")
+	cleanup := testutils.MakeTempDir(t, tmpDir)
+	modDir := filepath.Join(tmpDir, "module")
+	testutils.MakeTempDir(t, modDir)
+
+	configPath := filepath.Join(tmpDir, configFile)
+	err := writeConfig(configPath, configContents(addr, tmpDir, tls, services))
+	require.NoError(t, err)
+	modulePath := filepath.Join(modDir, "main.tf")
+	err = writeConfig(modulePath, nullModuleContents())
+	require.NoError(t, err)
+
+	return configPath, cleanup
+}
+
+// write the (config) contents to the path
+func writeConfig(configPath, contents string) error {
+	f, err := os.Create(configPath)
+	if err != nil {
+		return nil
+	}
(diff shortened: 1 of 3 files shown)


e2e/benchmarks/services_test.go
```diff
@@ -0,0 +1,276 @@
...
+      node_meta             = map(string)
+
+      cts_user_defined_meta = map(string)
+    })
+  )
+}
+`
+}
+
+func fileModuleContents() string {
```
lornasong, line 242: I'm not seeing that this method is used in this PR. Am I missing something or maybe there are future plans for it?

e2e/benchmarks/services_test.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+// ----------------------------------------------------------------------------
+
+// Run CTS in once mode from the cli level
+func runSyncOnce(configPath string) error {
+	cli := command.NewCLI(os.Stdout, os.Stderr)
+	args := []string{"cts", "-once", fmt.Sprintf("--config-file=%s", configPath)}
+	exitCode := cli.Run(args)
+	errstr := "Exit Code Erro: %s (%d)\n"
```
lornasong, line 76: minor thing "Exit Code Erro:" => "Exit Code Error:"

e2e/benchmarks/services_test.go
```diff
@@ -0,0 +1,276 @@
...
+  source = "../../tmp_stress/module"
+  providers = ["local"]
+  services = ["stress"]
+}
+
+`, address, tls, pwd)
+}
+
+// null resource, services module
+func nullModuleContents() string {
```
lornasong, line 210: Suggestion only if you like, I think it could be worth adding the actual null module files. I could see myself using this in the future.

e2e/benchmarks/services_test.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+// generate service instance entries
+func generateServices(n int, instances bool) []testutil.TestService {
+	baseport := 30000
+	services := make([]testutil.TestService, n)
+	for i := 0; i < n; i++ {
+		name, id := "service", fmt.Sprintf("svc_%d", i)
+		if instances {
+			name, id = fmt.Sprintf("svc_%d", i), "instance"
```
lornasong, line 120: I like how you're generating services vs. service instances here! It's very elegant

e2e/benchmarks/services_test.go
```diff
@@ -0,0 +1,276 @@
...
+	"github.com/hashicorp/consul-terraform-sync/testutils"
+	"github.com/hashicorp/consul/sdk/testutil"
+	"github.com/stretchr/testify/require"
+)
+
+const configFile = "config.hcl"
+const tempDirPrefix = "tmp_"
+
+// Benchmark N Services Instance (all the same service)
+func Benchmark1Instance(b *testing.B) {
```
findkim, line 26: Minor suggestion: To make this regex friendly, structure the function names `BenchmarkInstances1XXX` allows for you to selectively run a set of benchmarks

```
go test ./e2e/benchmarks -bench=BenchmarkInstance
```

would run all 3 and not run the other benchmarks in `tasks_test.go` etc.

e2e/benchmarks/services_test.go
```diff
@@ -0,0 +1,276 @@
...
+		require.NoError(b, runSyncOnce(path))
+	}
+}
+
+// ----------------------------------------------------------------------------
+
+// Run CTS in once mode from the cli level
+func runSyncOnce(configPath string) error {
+	cli := command.NewCLI(os.Stdout, os.Stderr)
+	args := []string{"cts", "-once", fmt.Sprintf("--config-file=%s", configPath)}
```
lornasong, line 74: Suggestion only if you like to make this 'consul-terraform-sync' instead of 'cts' in case people haven't set up the cts alias on their local development.

e2e/benchmarks/services_test.go
```diff
@@ -0,0 +1,276 @@
...
+		require.NoError(b, runSyncOnce(path))
+	}
+}
+
+// Benchmark N Services
+func Benchmark1Service(b *testing.B) {
+	benchmarkServices(b, 1)
+}
+func Benchmark100Service(b *testing.B) {
+	benchmarkServices(b, 1)
```
findkim, line 52:
```suggestion
	benchmarkServices(b, 100)
```

e2e/benchmarks/services_test.go
```diff
@@ -0,0 +1,276 @@
...
+}
+func Benchmark1000Service(b *testing.B) {
+	benchmarkServices(b, 1)
```
findkim, line 55:
```suggestion
	benchmarkServices(b, 1000)
```

e2e/benchmarks/services_test.go
```diff
@@ -0,0 +1,276 @@
...
+}
+func benchmarkServices(b *testing.B, N int) {
```
findkim, line 57: These services benchmarks do not seem to be testing that CTS will monitor N services, the task itself is only monitoring one service "stress".

```
task {
  name = "stress-test"
  source = "../../tmp_stress/module"
  providers = ["local"]
  services = ["stress"]
}
```

The task would need to resemble the below to monitor all N services

```
task {
  services = ["service001", "service002", ...]
}

File: e2e/benchmarks/services_test.go [hunk 10]
```diff
@@ -0,0 +1,276 @@
...
+func Benchmark100Instances(b *testing.B) {
+	benchmarkInstances(b, 100)
+}
+func Benchmark1000Instances(b *testing.B) {
+	benchmarkInstances(b, 1000)
+}
+func benchmarkInstances(b *testing.B, N int) {
+	srv := testutils.NewTestConsulServerHTTPS(b, "../../testutils")
+	defer srv.Stop()
+	path, cleanup := setupServiceInstances(b, srv, N)
```
findkim, line 38: N service instances are registered with Consul as the name "service", however CTS for the `benchmarkInstances` are monitoring the service named "stress" which does not end up querying all of those instances. Can you update either the name they are registered with Consul as or update the service name the CTS task is monitoring?

e2e/benchmarks/services_test.go
eikenb: No plans, just missed that when cleaning out leftovers. Didn't see it at the bottom. Thanks!

e2e/benchmarks/services_test.go
eikenb: :facepalm:

e2e/benchmarks/services_test.go
eikenb: It isn't actually running the command. It is calling cli.Run(), directly running the code in from the benchmark so it can be profiled. I used 'cts' instead of 'consul-terraform-sync' to help, mistakenly perhaps, call out this difference. Maybe there is something better I can put in there?

e2e/benchmarks/services_test.go
eikenb: Yeah... I went the other way. Made it so it was easy to call one without the others via regex. 
Eg.

`go test -tags e2e -bench 1I` 

Where `1I` can be `100I`, `1000I` or `1S` etc. 
Then for more you use `*`, eg. `1*I` or `1*S`.

I'm not married to the idea. It was just the most useful as I was developing them. If the other way is preferred, I'll change it no worries.

e2e/benchmarks/services_test.go
eikenb: Sloppy C-n-P. Sorry.

e2e/benchmarks/services_test.go
eikenb: Your right. I worked on the instances version at first and was still thinking along those lines when I added the services. It needs that added.

e2e/benchmarks/services_test.go
eikenb: Saw this in the last one. Kind of hoped you missed it so I could fix it as this is getting embarrassing.

eikenb: Going to fix this so it uses the same variable. Eliminate the worry of keeping them in sync.

e2e/benchmarks/services_test.go
findkim: Oh! TIL that it works the other way. Go continues to surprise me, thanks for sharing.

e2e/benchmarks/services_test.go
eikenb: Hey... I thought HTTP/2 would be useful here but then it ended up not being. Should have thought about that more at the time as it would have indicated this issue as it does hit the issue with 1000 services. So I switched the MultiServices ones to use HTTP/2 while the Instances ones will stick with HTTP/1.

e2e/benchmarks/services_test.go
eikenb: You mean including them in a directory like the e2e/test_modules/ ones? Instead of it being generated?

lornasong: Yes, sorry for the lack of clarity! Yes, including the files in 'e2e/test_modules/' instead of generating them.

e2e/benchmarks/services_test.go
lornasong: Ah 🤦 I see! I'm ok with leaving it as is. You also have a clarifying comment on the function already. I think it's just me. I've also been working with running commands lately, so it could be because my head is in that space already. Thanks for clarifying for me!

lornasong (approved):
Approving for the things that we chatted about ✨✨✨. Nonblocking optional suggestion to include null module files instead of generating them - but only if you like!

e2e/benchmarks/services_test.go
findkim: New changes look good! Thanks for fixing it.

Are you observing different Consul API behavior between the two MultiServices and Instances over HTTP/1? It's not clear to me how the Consul HTTP limit doesn't affect 1000 instances but does for 1000 services when they're both using the same endpoint.

e2e/benchmarks/services_test.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+	return makeConfig(t, services, srv.HTTPSAddr, true)
+}
+
+// generate service instance entries
+func generateServices(n int, instances bool) []testutil.TestService {
+	baseport := 30000
+	services := make([]testutil.TestService, n)
+	for i := 0; i < n; i++ {
+		name, id := fmt.Sprintf("svc_%d", i), "instance"
```
findkim, line 123: Running the `BenchmarkNService` benchmarks locally, the `terraform.tfvars.tmpl` contains querying the N services, however I'm noticing that the `terraform.tfvars` file is not rendered with the expected services. For the 100 service benchmark, only `svc_099` renders in the tfvars file. For 1000 services I'm seeing an empty `services = { }`

The service ID is unique across all services for an agent. I suspect that `id = "instance"` is overriding the service registration instead of registering unique N services. Could you look into why the `tfvars` is not rendering as expected with 1 instance for N services?

e2e/benchmarks/services_test.go
eikenb: Wouldn't the Instances case be 1 watched service with 1000 entries and so only need the 1 connection?

findkim: Oh I thought this was on the topic of registering the services within `setupServiceInstances` and not the client for watching.

e2e/benchmarks/services_test.go
eikenb: Interesting... I'll need to look into this one. :thinking:

eikenb: I think there's a race for the `services = {}` vs. the single entry. I initially only saw `services = {}` even with `100S` unless I put a sleep in between the population code and the CTS run.

So I need to figure that out as well.

findkim: Interesting, a potential race and maybe the instance iD isn't an issue then 🤔

eikenb: No... not the race I thought it was as the sleep doesn't fix it reliably. Even with a long sleep it sometimes works and sometimes doesn't. I'm going to dig more into it today.

eikenb: Since the  100Instances test works as it should, I tried changing the 100Services test (the one having the problem) so each Service used had unique names and ids (previously only name was unique) as this was one of the differences from the 100Instances. The result it that I get a terraform.tfvars that usually contains 1-3 entries, with all their names/ids being early entries (they are numbered 1-100) so far always being <30.

I also added a call to the consul server to print out all services after the initialization but before the CTS run is triggered. All 100 services are registered.

I'm working to check the health status of those 100 services now.

eikenb: Printed out all the health checks for the services and they are all present via that endpoint as well and all passing. Again, this was printed after the seeding of test data into consul but before the CTS run. So that data is all present and good in Consul prior to the CTS run.

Might be an actual bug.

e2e/benchmarks/services_test.go
eikenb: You are right, I switched context there sort of combining this and the changes related to the previous comment. Made sense to me at the time but was a bit of a jump. Sorry for the confusion.

e2e/benchmarks/services_test.go
eikenb: I'll do this but I'll do it later. Spent all my time getting the benchmarks to work right and forgot this and don't have time today and we want to go ahead and merge this. I'll add a note to refactor this our next week as it won't take long.

findkim (approved):
I pulled in the Wait delay bump from https://github.com/hashicorp/hcat/pull/41 and re-ran these benchmarks. All 100 and 1000 services made it in the rendered file! The work-around is sufficient ✨ 

Service ID changes are good too 👍

PR #235, opened 2021-03-25
Fix exponential backoff calculation for retry methods

lornasong:
Previous "exponential" backoff algorithm was `x**2`, which is incorrect. Adjusted
to be actually exponential:`2**x`


findkim (commented):
x^2 vs 2^x nice catch

findkim (approved):
forgot to approve

PR #238, opened 2021-03-29
Default version name

findkim:
CTS versions 0.1.0-techpreview2 and 0.1.0-beta output excluded the CLI name due to the build process not setting it. We likely missed this because the local development build process using the [Makefile set the name](https://github.com/hashicorp/consul-terraform-sync/blob/master/Makefile#L26).

```
$ consul-terraform-sync -version
 0.1.0-techpreview2 (d844653)
Compatible with Terraform >= 0.13.0, < 0.15
```

should be

```
$ consul-terraform-sync -version
consul-terraform-sync 0.1.0-techpreview2 (d844653)
Compatible with Terraform >= 0.13.0, < 0.15
```

not going to back port this, fixing for 0.1.0 GA and beyond

### DIFF
--- a/command/cli.go
+++ b/command/cli.go
@@ -125,7 +125,7 @@ func (cli *CLI) Run(args []string) int {
 
 		exitCode, err := subcommands.Run()
 		if err != nil {
-			log.Printf("[ERROR] running the CLI command '%s': %s",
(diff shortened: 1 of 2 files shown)


lornasong (approved):
nice catch!

PR #240, opened 2021-03-31
Add benchmark for task to be triggered

findkim:
2.5s from when Catalog changes are detected to re-rendering the `terraform.tfvars` file before Terraform commands are executed on the workspace. The benchmark is testing with the task buffer period disabled. The 2.5s includes the added 1s delay to mitigate race condition (context https://github.com/hashicorp/consul-terraform-sync/pull/232#pullrequestreview-621647245)

```
$ go test ./e2e/benchmarks/ -bench BenchmarkTaskTrigger -benchtime=10s -tags e2e
[INFO] freeport: detected ephemeral port range of [49152, 65535]
[INFO] freeport: reducing max blocks from 30 to 26 to avoid the ephemeral port range
goos: darwin
goarch: amd64
pkg: github.com/hashicorp/consul-terraform-sync/e2e/benchmarks
BenchmarkTaskTrigger-8   	       6	2507039392 ns/op
PASS
ok  	github.com/hashicorp/consul-terraform-sync/e2e/benchmarks	32.846s
```

### DIFF
--- a/e2e/benchmarks/task_trigger_test.go
+++ b/e2e/benchmarks/task_trigger_test.go
@@ -1,0 +1,109 @@
+// +build e2e
+
+// $ go test ./e2e/benchmarks/ -bench BenchmarkTaskTrigger -benchtime=10s -tags e2e
+package benchmarks
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/hashicorp/consul-terraform-sync/controller"
+	"github.com/hashicorp/consul-terraform-sync/event"
+	"github.com/hashicorp/consul-terraform-sync/testutils"
+	"github.com/hashicorp/consul/sdk/testutil"
+	"github.com/stretchr/testify/require"
+)
+
+func BenchmarkTaskTrigger(b *testing.B) {
+	// Benchmarks the time for a Consul catalog change to trigger and re-render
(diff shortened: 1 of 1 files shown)


e2e/benchmarks/task_trigger_test.go
```diff
@@ -0,0 +1,109 @@
...
+	go func() {
+		err := ctrl.Run(ctx)
+		ctrlStopped <- err
+	}()
+
+	// Monitor templated file for updates
+	tfvarsFile := filepath.Join(tempDir, "task_000", "terraform.tfvars")
+	fileLastUpdated := getFileModTime(b, tfvarsFile)
+	fileUpdated := make(chan struct{}, b.N)
+	go func(lastUpdate *time.Time) {
```
eikenb, line 59: Nice. I like this way of detecting file changes.

lornasong (approved):
Good to know the time from a Catalog change to triggering a task is ~2.5s!

eikenb (approved):
Once we have a decent number of these we should have them automatically run (weekly, per-CI run, ???) and store the results with a way for us to easily check (graph?) the results to watch for performance regressions or validate improvements.

findkim:
Thanks for the review y'all!

I like the idea of running these automatically. I was coincidentally reading up on GitHub action scheduling this morning and it would be a good candidate to run these weekly. Storing results to track performance regressions/improvements is something we can look into, will add this to our backlog.

PR #241, opened 2021-04-01
split out null_resource test module for reuse

eikenb:
I already have a case and I know @lornasong thought this would be useful.


findkim (approved):
Nice cleanup! ✨

findkim:
Hm, the CI tests didn't run for this PR -- any idea if you need to reauth CircleCI?

eikenb:
I'm not sure what happened with CircleCI, I never had to be logged in there for it to work before but who knows with the new setup. I'm going to test/figure that out separately as I'd like to get this merged so I can use it. I did login to CircleCI without issue.

Issue #233, opened 2021-03-23
CTS does not re-establish connection with Consul if Consul stops and restarts


lornasong:
### Describe the bug
When CTS is running and Consul stops, CTS throws an error as expected:

`[ERR] (ctrl) error watching template dependencies: health.service(lsong?filter=Checks.Status == "passing"): Get "http://localhost:8500/v1/health/service/api?filter=Checks.Status+%3D%3D+%22passing%22&index=40": dial tcp [::1]:8500: connect: connection refused`

However, once Consul restarts, CTS no longer is triggered when service instances change

### Versions
**Consul Terraform Sync**
```
consul-terraform-sync 0.1.0-beta (6ef6680)
Compatible with Terraform >= 0.13.0, < 0.15
```

**Consul Version**
```
Consul 1.9.4
```
(Experience with other versions as well)
**Terraform Version**
```
Terraform v0.14.4
```

### Configuration File(s)
_Reminder to redact any sensitive information that may be present in this file_

<details>
<summary>Click to toggle contents of config file </summary>

```hcl
log_level = "info"

consul {
  address = "localhost:8500"
}

driver "terraform" {
  log = true
  working_dir = "."
}

task {
  name = "task_c"
  services = ["api"]
  source      = "findkim/print/cts"
  version     = "0.1.0"
}
```
</details>

### Terraform Configuration Files Generated by Consul-Terraform-Sync
_Reminder to redact any sensitive information that may be present in the files_

<details>
<summary>Click to toggle contents of main.tf </summary>

```terraform
terraform {
  required_version = ">= 0.13.0, < 0.15"
  backend "consul" {
    address = "localhost:8500"
    gzip    = true
    path    = "consul-terraform-sync/terraform"
  }
}

module "task_c" {
  source   = "findkim/print/cts"
  version  = "0.1.0"
  services = var.services
}
```
</details>

<details>
<summary>Click to toggle contents of terraform.tfvars </summary>

```terraform
services = {
}

```
</details>

### Expected Behavior
When Consul stops and restarts, expect that CTS would re-establish connection with Consul and update resources with latest Consul Catalog.

### Actual Behavior
When Consul stops and restarts, CTS no longer updates resources with latest Consul Catalog. Any changes in Consul Catalog appear to be unobserved by CTS

### Steps to Reproduce
1. Start Consul with no services: `consul agent -dev`
2. Start consul-terraform sync with the above config file: `consul-terraform-sync -config-file=config.hcl`
[...]

PR #242, opened 2021-04-01
Add retry when hcat connection with Consul has an issue

lornasong:
When hcat polls Consul for changes in Consul catalog, it will now retry on error. It will retry a maximum of 9 times with exponential backoff. 9 retries with some jitter totals to waiting 8.5-12.8 minutes. If after 9 retries there is still an error, then CTS will exit.

Commits (see commit message for more details)
 - Make retry package's waitTime() public
 - Configure hcat watcher with a Consul retry function (this consumes retry.WaitTime()) (modified in last commit)
 - Pull latest hcat version
 - Update CTS to exit when hcat watcher errors since there is a retry now
 - Refactor testutils for changes needed to write a Consul restart e2e test
 - Consul restart e2e test
 - Update naming and comments for clarity

Resolves: https://github.com/hashicorp/consul-terraform-sync/issues/233

### DIFF
--- a/controller/readwrite.go
+++ b/controller/readwrite.go
@@ -26,7 +26,7 @@ type ReadWrite struct {
 	store *event.Store
 	retry retry.Retry
 
-	testMode bool
+	testMode   bool
 	taskNotify chan string
 }
 
@@ -41,7 +41,7 @@ func NewReadWrite(conf *config.Config, store *event.Store) (Controller, error) {
 		baseController: baseCtrl,
 		store:          store,
 		retry:          retry.NewRetry(defaultRetry, time.Now().UnixNano()),
-		taskNotify: make(chan string, len(*conf.Tasks)),
+		taskNotify:     make(chan string, len(*conf.Tasks)),
 	}, nil
 }
 
@@ -71,6 +71,7 @@ func (rw *ReadWrite) Run(ctx context.Context) error {
 		case err := <-rw.watcher.WaitCh(ctx):
 			if err != nil {
 				log.Printf("[ERR] (ctrl) error watching template dependencies: %s", err)
+				return err
 			}
 
 		case <-ctx.Done():
--- a/controller/watcher.go
+++ b/controller/watcher.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"log"
+	"math/rand"
 	"time"
 
 	"github.com/hashicorp/consul-terraform-sync/config"
@@ -69,8 +70,8 @@ func retryConsul(retryCount int) (bool, time.Duration) {
 		return false, 0 * time.Second
 	}
 
-	r := retry.NewRetry(0, time.Now().UnixNano()) // 0 is not used
-	wait := r.WaitTime(uint(retryCount))
+	random := rand.New(rand.NewSource(time.Now().UnixNano()))
+	wait := retry.WaitTime(uint(retryCount), random)
 	dur := time.Duration(wait)
 	log.Printf("[WARN] (hcat) error connecting with Consul. Waiting %v to retry"+
 		" attempt #%d", dur, retryCount+1)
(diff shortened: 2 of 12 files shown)


e2e/e2e_test.go
```diff
@@ -100,6 +102,53 @@ func TestE2ERestartSync(t *testing.T) {
...
+
+	// stop Consul
+	consul.Stop()
+	time.Sleep(2 * time.Second)
+
+	// restart Consul
+	consul = testutils.NewTestConsulServer(t, testutils.TestConsulServerConfig{
+		HTTPSRelPath: "../testutils",
+		PortHTTPS:    parsePort(t, consul.HTTPSAddr),
+	})
```
findkim, line 136: Interesting to find that the SDK doesn't have a `Start()` or restart method for the Consul test server!

I don't like the relative path being a parameter, it's awkward and clunky. But that aside. Would it be possible to get the port from the config?

```suggestion
	consul = testutils.NewTestConsulServer(t, testutils.TestConsulServerConfig{
		HTTPSRelPath: "../testutils",
		PortHTTPS:   consul.Config.Ports.HTTPS),
	})
```

controller/watcher.go
```diff
@@ -46,11 +50,33 @@ func newWatcher(conf *config.Config) (*hcat.Watcher, error) {
...
+func retryConsul(retryCount int) (bool, time.Duration) {
+	max := 8 // 8+1 retries. total wait of 8.5-12.8 min
+	if retryCount > max {
+		log.Printf("[ERR] (hcat) error connecting with Consul even after retries")
+		return false, 0 * time.Second
+	}
+
+	r := retry.NewRetry(0, time.Now().UnixNano()) // 0 is not used
+	wait := r.WaitTime(uint(retryCount))
+	dur := time.Duration(wait)
```
findkim, line 74: This is a round-about way to get the exponential backoff duration from the retry count. What do you think about refactoring `WaitTime` to be a standalone function in addition to the retry object's method? Something like:


```go
func WaitTime(attempt uint, random) int {
  // move current r.waitTime logic here
}

func (r Retry) waitTime(attempt uint) int {
	if r.testMode {
		return 1
	}
	return WaitTime(attempt, r.random.Float64())
}
```

controller/readwrite.go
```diff
@@ -26,7 +26,7 @@ type ReadWrite struct {
 	store *event.Store
 	retry retry.Retry
 
-	testMode bool
+	testMode   bool
```
findkim, line 29: Thanks for linting this file. My editor has been inconsistent missed these :(

findkim (approved):
Minor suggestions that aren't blocking.

Great job fixing a handful of things along the way to get CTS behave in a predictable manner

eikenb (approved):
Looks good to me too.

e2e/e2e_test.go
lornasong: :nod: a `Start()` or `Restart()` would be great. I agree, I don't like the look of the relative path as a config field as much. It seemed like we might add more customizations to the test server though, so I thought the config would be the best option. The other options I thought of seemed even more clunky. If you think of any alternatives you like, I’m very open to it!

Re: getting port from config. Yes!! Thanks for finding this. This is much better. I will update to use this and remove `parsePort`

controller/watcher.go
lornasong: Good idea! I like this a lot. It's a lot cleaner. I will make this refactor!

e2e/e2e_test.go
findkim: No alternatives come to mind for me just yet.

lornasong: 👍 happy to open a new PR if anyone comes up with something better. will continue to mull on this myself as well

PR #243, opened 2021-04-03
move bulk consul service loading to testutils

eikenb:
Move the functions used to load consul with test servers/instances into
the testutils module for reuse in other tests.

Also moved the ShowMe* functions as they are also generally handy.

Cleaned up the loading functions a bit and refactored the existing tests
to use the shared functions instead of the old, local versions.

### DIFF
--- a/testutils/consul.go
+++ b/testutils/consul.go
@@ -1,8 +1,11 @@
 package testutils
 
 import (
+	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"log"
+	"net/http"
 	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


testutils/consul.go
```diff
@@ -50,3 +53,95 @@ func NewTestConsulServer(tb testing.TB, config TestConsulServerConfig) *testutil
...
+// when you need to check on Consul's data .
+
+// All registered services
+func ShowMeServices(t testing.TB, srv *testutil.TestServer) {
+	u := fmt.Sprintf("http://%s/v1/agent/services", srv.HTTPAddr)
+	resp := RequestHTTP(t, http.MethodGet, u, "")
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		panic(err)
+	}
```
lornasong, line 118: For the checking errors in test functions, we've often been using `require.NoError` instead of the if-statement. I've been liking it since its a little cleaner to my eyes. Suggestion to use `require.NoError` if you'd like! If you do like, there are two more instances of error checks below (line 129 + 141)

testutils/consul.go
```diff
@@ -50,3 +53,95 @@ func NewTestConsulServer(tb testing.TB, config TestConsulServerConfig) *testutil
...
+	fmt.Println(string(b))
+	defer resp.Body.Close()
+}
+
+// Health status for all services
+func ShowMeHealth(t testing.TB, srv *testutil.TestServer) {
+	// get the node
+	u := fmt.Sprintf("http://%s/v1/health/service/%s", srv.HTTPAddr, "svc_0")
```
lornasong, line 126: It looks like "svc_0" is a service name that the engineer would manually modify. If I'm understanding this right, I think it could be useful to make this value a variable so that engineers won't need to come into the method to modify it.

lornasong (approved):
Looks good to me! I left some non-blocking suggestions

testutils/consul.go
```diff
@@ -50,3 +53,95 @@ func NewTestConsulServer(tb testing.TB, config TestConsulServerConfig) *testutil
...
+		func(i int) string {
+			return fmt.Sprintf("svc_id_%d", i)
+		})
+}
+
+// shorter name for the formatting functions
+type fmtFunc func(i int) string
+
+// does the actual work of generating the TestService objects
+func generateServices(n int, namefmt, idfmt fmtFunc) []testutil.TestService {
```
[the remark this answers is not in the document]
eikenb: Thanks. First class functions are awesome.

testutils/consul.go
eikenb: Thanks for the suggestion. I wasn't sure about other use cases so I just left them as I had them, but I'm happy to make this change.

testutils/consul.go
eikenb: Good call. That is more compressed and fits better with the other code. I'll update those panics to NoError checks.

PR #244, opened 2021-04-03
Services render race test

### LINKED ISSUE #236: Add test that mimics 100S benchmark issue
See this PR thread for more...

https://github.com/hashicorp/consul-terraform-sync/pull/232#pullrequestreview-617807218

Need to test the contents of `sync-tasks/test-services-task/terraform.tfvars` for all relevant entries when running a test similar to the `Benchmark100Service` benchmark. That is the former should have 100 entries when running the latter.

eikenb:
Note this is branched off the bulk-loader-util branch and depends on it being merged first.

Fixes #236

### DIFF
--- a/e2e/benchmarks/services_test.go
+++ b/e2e/benchmarks/services_test.go
@@ -6,10 +6,7 @@
 package benchmarks
 
 import (
-	"encoding/json"
 	"fmt"
-	"io/ioutil"
-	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
@@ -102,44 +99,24 @@ func runSyncOnce(configPath string) error {
 	}
 }
 
-// Sets up consul and config to test setup with N service instances
+// Sets up consul and config with N service instances of the same service
 func setupServiceInstances(t testing.TB, srv *testutil.TestServer, N int) (string, func() error) {
-	services := make([]string, 1)
-	for _, s := range generateServices(N, true) {
-		testutils.RegisterConsulService(t, srv, s, testutil.HealthPassing)
-		services[0] = s.Name
-	}
+	testInstances := testutils.TestInstances(N)
+	testutils.AddServices(t, srv, testInstances)
+	services := []string{testInstances[0].Name} // shared name
 	return makeConfig(t, services, srv.HTTPAddr, false)
 }
 
-// add generated services as separte services
+// Sets up consul and config with N unique services
 func setupMultiServices(t testing.TB, srv *testutil.TestServer, N int) (string, func() error) {
+	testServices := testutils.TestServices(N)
+	testutils.AddServices(t, srv, testServices)
+
 	services := make([]string, N)
-	for i, s := range generateServices(N, false) {
-		testutils.RegisterConsulService(t, srv, s, testutil.HealthPassing)
+	for i, s := range testServices {
 		services[i] = s.Name
 	}
 	return makeConfig(t, services, srv.HTTPSAddr, true)
-}
-
-// generate service instance entries
-func generateServices(n int, instances bool) []testutil.TestService {
-	baseport := 30000
-	services := make([]testutil.TestService, n)
-	for i := 0; i < n; i++ {
-		name, id := "test_service", fmt.Sprintf("svc_%d", i)
-		if !instances {
-			name = id
-		}
-		services[i] = testutil.TestService{
-			Name:    name,
-			ID:      id,
-			Address: "127.0.0.2",
-			Port:    int(baseport + i),
-			Tags:    []string{},
-		}
-	}
-	return services
 }
 
 // writes the config files out to disk, returns path to them and cleanup func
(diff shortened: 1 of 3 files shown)


e2e/issues_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+// issues GH-236 and GH-239
+func TestServicesRenderRace(t *testing.T) {
+	const NumberOfServices = 100
+
+	srv := testutils.NewTestConsulServer(t,
+		testutils.TestConsulServerConfig{HTTPSRelPath: "../testutils"})
+	defer srv.Stop()
+
+	tempDir := t.Name()
```
lornasong, line 29: I like the `t.Name()` a lot more than the current pattern I've been using to name the temporary directory. Will have to steal... :)

e2e/issues_test.go
```diff
@@ -0,0 +1,81 @@
...
+	err = rwCtrl.Once(ctx)
+	require.NoError(t, err)
+
+	// veryify results
+	tfvarsFile := filepath.Join(tempDir, "serv_rend_race_task", "terraform.tfvars")
+	data, err := os.ReadFile(tfvarsFile)
+	require.NoError(t, err)
+	// 'svc_name_' is unique per service and much easier to count than
+	// parsing the file and counting the real entries
+	require.Equal(t, NumberOfServices, bytes.Count(data, []byte("svc_name_")))
```
lornasong, line 80: nice! like the bytes.Count!

e2e/issues_test.go
```diff
@@ -0,0 +1,81 @@
...
+	require.NoError(t, err)
+
+	// run controller
+	ctrl, err := controller.NewReadWrite(conf, event.NewStore())
+	rwCtrl := ctrl.(*controller.ReadWrite)
+	require.NoError(t, err)
+	_, err = rwCtrl.Init(ctx)
+	require.NoError(t, err)
+	err = rwCtrl.Once(ctx)
+	require.NoError(t, err)
```
lornasong, line 72: I'm curious about choosing to run CTS by calling the controller rather than running the binary! I know we have the runSyncOnce() option which could be useful. I was wondering if you could share your thought process?

e2e/issues_test.go
```diff
@@ -0,0 +1,81 @@
...
+	for i, s := range testServices {
+		serviceNames[i] = s.Name
+	}
+	conf.Tasks = &config.TaskConfigs{
+		&config.TaskConfig{
+			Name:     config.String("serv_rend_race_task"),
+			Source:   config.String("../../test_modules/null_resource"),
+			Services: serviceNames,
+		}}
+	conf.Consul.Address = config.String(srv.HTTPAddr)
```
findkim, line 54: I'm wondering why HTTP is used for this e2e test instead of HTTPS, can you share some thought around this?

e2e/issues_test.go
```diff
@@ -0,0 +1,81 @@
+// +build e2e
+
+//
```
findkim, line 3: Can this be cleaned up or modified?

Can the filename be renamed to better capture at a high level what it tests? Something like `render_test.go` -- `issues_test.go` reads to me more related to GitHub than the content of the test itself.

e2e/issues_test.go
eikenb: I stole that from Kim. I like it is as well. :)

e2e/issues_test.go
eikenb: Couple reasons for not calling cli.Run() directly like runSyncOnce()..
- This is a more targeted test of the one problem. I wanted to constrain it more to the problem area.
- I wanted to try it. Kim's been doing this and it seemed nice and explicit as you don't use intermediate config files, but create the config object directly.

Why I didn't run the binary.. I was thinking of proposing we switch from the binary over to using cli.Run() instead. It would mean no more worrying about rebuilding the binary each time you make some change and are running individual tests. It would also eliminate the binary call overhead which should shave a bit of time off the suite for no real loss as all that really tests is the underlying system exec call and Go's os.Args parsing.

e2e/issues_test.go
eikenb: It wasn't relevant to the specific test and the test was a bit simpler and ran a bit faster without.
We have a test for the HTTPS setup and so, IMHO, should otherwise use whichever protocol is faster or more appropriate for that specific test.

e2e/issues_test.go
eikenb: I wasn't 100% on the name but wanted to call out that this was a bug related test, targeting a specific issue, and not a general/feature/golden-path e2e test. I'd be happy to change the name and use a comment to call out the issue, but I thought maybe a bit more info (filename, common subdir) might be useful.

e2e/issues_test.go
lornasong: Thanks for sharing! That makes sense. I also prefer how config building works for cli.Run(). I'm up for switching tests that target constrained problems/use-cases to using cli.Run()! We can use running the full binary for specific tests that actually benefit from it and maybe some 'basic' tests (for peace of mind for a complete e2e).

lornasong (approved):
approving for things we chatted about!

e2e/issues_test.go
findkim: 👍 thanks fine with me then! Seeing how the server setup above used `testutils.NewTestConsulServer()` which adds the overhead of HTTPS stuff, I wanted to double check on the intent. 

Non-blocking comment: In this case you could optionally configure the server using the Consul SDK directly or make a helper function that only does HTTP.

```
srv, err := testutil.NewTestServerConfigT(tb,
	func(c *testutil.TestServerConfig) {
		c.LogLevel = "warn"
		c.Stdout = ioutil.Discard
		c.Stderr = ioutil.Discard
	})
```
https://github.com/hashicorp/consul-terraform-sync/blob/master/testutils/consul.go#L32-L36

e2e/issues_test.go
findkim: Gotcha, yeah I saw the comment on [#L21](https://github.com/hashicorp/consul-terraform-sync/pull/244/files/c7c65a7ad81f0d11fa88f6cc28c7b670245b2911..13c493ef5855cdf7830a8738fe6eab80152a1a2c#diff-986d4a3319e0f499de5e5970cb965cde1c2571d7d9502c113c082b0321bb8e8cR21).

If you imagine this test file persisting for a while, I'd prefer it to be renamed since there are many github issues and we'll likely write more tests targeting bugs captured in gh issues.

e2e/issues_test.go
eikenb: Using `testutils.NewTestConsulServer()` and having it support HTTPS doesn't really add much overhead to the startup. HTTPS mainly adds overhead to the connections from the TLS negotiation.
I think the common function works fine even if we don't use the TLS connection.

e2e/issues_test.go
eikenb: Cool. I'll rename it. :+1:

eikenb: Oh... and yeah, I'll get rid of that empty comment.

findkim (approved):
Excited about more e2e tests!

Issue #74, opened 2020-09-24
Invalid expression error when no instances of a service


lornasong:
When running the daemon, if there are no services, daemon will throw with the following error:

```
could not apply: Error tf-apply for 'print': 
Error: Invalid expression
  on terraform.tfvars line 8:
   8: services = {,
Expected the start of an expression, but found an invalid expression token.
```

This happens when there are no instances of a service. Some reasons: 
 - the consul agent has not yet started up - hence the daemon is unable to retrieve registered instances
 -  there just are no instances of a service registered (yet)

If no instances are registered of any service, this will cause the `terraform.tfvars` file to contain something like: 
```
services = {,
}
```

If some services have instances while others don't, there can be a preceding comma. This happens when the service that is alphabetically first does not have an instances. For example, if there is the `api` and `web` service declared for a task but `web` service has a registered instance and `api` does not, the `tfvars` file will contain something like:
```
services = {,
  "web" : {
    id              = "web"
    name            = "web"
    address         = "127.0.0.1"
    ....
  }
}
```

PR #247, opened 2021-04-06
Always set trailing comma in tfvars.tmpl

findkim:
* Updates the examples folder to include `providers.tfvars` file
* Always include trailing comma in `var.services` instead of omitting last comma. This makes the template more straightforward to read but still resolves the bug https://github.com/hashicorp/consul-terraform-sync/issues/74 when no services are found.
* Fix `TestRenderTFVarsTmpl` hanging indefinitely on `Wait` when the content did not match expected value


templates/tftmpl/tfvars.go
```diff
@@ -162,21 +158,13 @@ func nonNullMap(m map[string]string) map[string]string {
...
-const baseAddressStr = `
+const serviceBaseTmpl = `
 {{- with $srv := service %s }}
-  {{- $last := len $srv | subtract 1}}
-  {{- range $i, $s := $srv}}
+  {{- range $s := $srv}}
   "{{ joinStrings "." .ID .Node .Namespace .NodeDatacenter }}" : {
 {{ HCLService $s | indent 4 }}
-  } {{- if (ne $i $last)}},{{end}}
+  },
```
lornasong, line 168: nice! much cleaner to have the dangling comma and remove this bit of template + baseCommaStr  ✨

templates/tftmpl/integration_test.go
```diff
@@ -229,21 +229,26 @@ func TestRenderTFVarsTmpl(t *testing.T) {
...
 					// there may be a race with the consul services registering
 					// let's retry once.
-					if (string(gld) != string(re.Contents)) && retry < 1 {
-						retry++
+					contents := string(re.Contents)
+					if (string(gld) != contents) && cachedContents == "" {
+						cachedContents = contents
 						continue
 					}
 					assert.Equal(t, string(gld), string(re.Contents))
```
lornasong, line 245: Suggestion to reuse `contents` here instead of `string(re.Contents)`

templates/tftmpl/integration_test.go
```diff
@@ -229,21 +229,26 @@ func TestRenderTFVarsTmpl(t *testing.T) {
...
 			gld := testutils.CheckFile(t, true, tc.goldenFile, "")
-			retry := 0
+			var cachedContents string
 			for {
 				re, err := r.Run(tmpl, w)
 				require.NoError(t, err)
 
 				if re.Complete {
 					// there may be a race with the consul services registering
 					// let's retry once.
```
lornasong, line 239: Looks like the 'retry once' comment no longer relevant. Remove?

templates/tftmpl/integration_test.go
```diff
@@ -229,21 +229,26 @@ func TestRenderTFVarsTmpl(t *testing.T) {
...
+						cachedContents = contents
 						continue
 					}
 					assert.Equal(t, string(gld), string(re.Contents))
 					break
 				}
+				if cachedContents != "" {
+					assert.Equal(t, string(gld), cachedContents)
+					t.Fail()
+				}
```
lornasong, line 251: It looks like this assert.Equal will always fail so long as `string(gld) != contents` on the first `re.Complete`. I might be missing something. I'm not sure that I'm following the intention of this if-block.

Also was wondering about using `t.Fail()`. I'm not familiar with it! From initial reading I see that it marks the test as failed. Is it doing some in addition to assert.Equal, which also marks failure?

templates/tftmpl/integration_test.go
findkim: thank you! missed this

templates/tftmpl/integration_test.go
findkim: idk if the race condition still exists, so i kept the retry in there but perhaps since the word retry is not used it's not as clear. I updated it to be more clear.

There's a bug in the test where `re.Complete` is true but the rendered content does not match the expected. In this case, it'll attempt a retry and `r.Run()` again but the second time `re.Complete` returns false and then block indefinitely on `w.WaitCh` because there is no new data.

---

Assert marks that test as failed only when it returns. However assert does not stop the test from continuing, and in this case it'll continue and proceed on the wait + for loop.

templates/tftmpl/integration_test.go
lornasong: Thanks for clarifying this in the comment below! 👍 the retry is still happening once.

templates/tftmpl/integration_test.go
lornasong: Ah! Gotcha. Thank you for all the details you shared around the `re.Complete` and `t.Fail`. I did a poor job of following the changes 🤦 . The 'retry' var you added in this iteration was helpful for me. Thanks!

lornasong (approved):
looks good! thanks for chatting with me about my questions!

findkim:
@lornasong thank you for the review and the questions! They helped identify areas that were not clear to be iterated on. 🙏

PR #248, opened 2021-04-06
Add e2e for testing theoretical Consul compatibility

lornasong:
Initial work for running Consul compatibility testing for OSS GA releases for the highest patch version given a major + minor version. 

Note:
 - Does not test datacenter-related compatibility
 - Is not run automatically in CircleCI with e2e tests (separate compatibility directory) since this test takes a long time. Definitely room for improvement on selecting specific Consul versions to test and parallelizing tests within a version.

### DIFF
--- a/e2e/compatibility/consul_test.go
+++ b/e2e/compatibility/consul_test.go
@@ -1,0 +1,549 @@
+// +build e2e
+
+// $ go test -run TestCompatibilty_Consul ./e2e/compatibility -tags=e2e -v
+package compatibility
+
+import (
+	"fmt"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"testing"
+	"time"
+
+	"github.com/hashicorp/consul-terraform-sync/api"
+	"github.com/hashicorp/consul-terraform-sync/templates/tftmpl"
+	"github.com/hashicorp/consul-terraform-sync/testutils"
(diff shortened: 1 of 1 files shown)


e2e/compatibility/consul_test.go
```diff
@@ -0,0 +1,549 @@
...
+	configFile    = "config.hcl"
+
+	nullTaskName = "null_task"
+)
+
+func TestCompatibilty_Consul(t *testing.T) {
+	// Tested only OSS GA releases for the highest patch version given a
+	// major minor version. v1.4.5 starts losing compatibility, details in
+	// comments. Theoretical compatible versions 0.1.0 GA:
+	consulVersions := []string{"1.9.4", "1.8.9", "1.7.13", "1.6.10", "1.5.3"}
```
eikenb, line 37: We should be able to programmatically determine this list for the most part, but a hard coded list is a good starting point.

findkim (approved):
This is an amazing set of tests ✨ 

Appreciate all the comments that helped make reading through it fast.

eikenb (approved):
I also really like the comments. They improve readability while keeping the code simple and contained to the test.

e2e/compatibility/consul_test.go
lornasong: Yes! Love the idea of a programatic way to determine this list! The hard-coded list isn't great for automatically running this in Circle and would need some up-keep

PR #249, opened 2021-04-13
Nightly Benchmarks

shore:
This PR adds a job to run the benchmark suite nightly.  Benchmark results are stored as a CircleCI artifact, `benchmarks.json`.  This is the standard JSON output from `go test`, which should be a good starting point for future work to automatically collect, parse, and plot the metrics.

Sample run of the benchmark suite: https://app.circleci.com/pipelines/github/hashicorp/consul-terraform-sync/853/workflows/f54edc08-c57b-439c-a8a2-489fbe0ebd7e/jobs/1660


.circleci/config.yml
```diff
@@ -47,9 +47,67 @@ jobs:
...
   build-test:
     jobs:
       - unit_integration_tests
       - e2e_tests
+  nightly-benchmarks:
+    jobs:
+      - benchmarks
+    triggers:
+      - schedule:
+          cron: 10 2 * * *
```
alvin-huang, line 109: Having a comment of what this cron time is would be helpful 😄

.circleci/config.yml
```diff
@@ -47,9 +47,67 @@ jobs:
...
+          command: /bin/cp /usr/local/bin/terraform ./e2e/
+      - run:
+          name: Version Check
+          command: |
+            go version
+            consul version
+            terraform version
+      - run:
+          name: Fetch Go modules
+          command: go mod download
```
alvin-huang, line 88: Running go tests should do this inherently so you probably don't need to split this out

.circleci/config.yml
```diff
@@ -47,9 +47,67 @@ jobs:
...
+      - run:
+          name: Run benchmark suite
+          no_output_timeout: << parameters.no_output_timeout >>
+          command:
+            go test -json ./e2e/benchmarks -timeout 2h -bench=. -tags e2e | /usr/bin/tee /tmp/benchmarks.json
```
alvin-huang, line 90: Can this use `gotestsum` and upload to CircleCI instead to have the tests be consumable by CircleCI? This will allow using test insights and will highlight failures in the UI quicker than scanning a json doc.   

ex:
```
gotestsum --format=short-verbose --junitfile $TEST_RESULTS_DIR/gotestsum-report.xml -- ./e2e/benchmarks -timeout 2h -bench=. -tags e2e
``` 
(untested)

and
```
      - store_test_results:
          path: *TEST_RESULTS_DIR
```

shore: I'm not a gotestsum guru, but I didn't think it handled benchmark data.  Have you seen it used to process benchmarks?

shore: I just tried this:  `$ gotestsum --format=short-verbose --junitfile moo.xml -- ./e2e/benchmarks/ -timeout 2h -bench=BenchmarkTasksConcurrent_t01_s50 -tags e2e` (limited to this one test because the whole suite takes a while).
I see this in the output: `BenchmarkTasksConcurrent_t01_s50/once_mode-4                   1        31854419398 ns/op`
But the results file doesn't have any benchmark data:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
        <testsuite tests="0" failures="0" time="0.000000" name="github.com/hashicorp/consul-terraform-sync/e2e/benchmarks">
                <properties>
                        <property name="go.version" value="go1.16.2 darwin/amd64"></property>
                </properties>
        </testsuite>
</testsuites>
```

alvin-huang: Is that related to https://github.com/gotestyourself/gotestsum/issues/62? Does `--format=standard-verbose` like what was suggested in the issue, change anything? If not then it may just not work. I haven't tested it myself but thought it might work so I suggested it 😄

shore: Using `standard-verbose` doesn't include the benchmark data either.  So it's probably the test2json issue you linked.

alvin-huang (approved):

.circleci/config.yml
```diff
@@ -47,9 +47,65 @@ jobs:
...
+          path: /tmp/benchmarks.json
+          destination: benchmarks.json
+
 workflows:
   version: 2
   build-test:
     jobs:
       - unit_integration_tests
       - e2e_tests
+  nightly-benchmarks:
```
lornasong, line 101: @hashicorp/api-tooling-core we had considered weekly at one point. Wanted to double-check if nightly works for you all?

go.mod
```diff
@@ -67,8 +56,13 @@ replace (
...
 	// `vault/api` nested go module.
-	github.com/hashicorp/vault => github.com/hashicorp/vault v1.5.5
+	github.com/hashicorp/vault => github.com/hashicorp/vault v1.7.0
 	github.com/hashicorp/vault/api => github.com/hashicorp/vault/api v1.0.5-0.20200805123347-1ef507638af6
-	github.com/hashicorp/vault/http => github.com/hashicorp/vault/http v1.5.5
-	github.com/hashicorp/vault/vault => github.com/hashicorp/vault/vault v1.5.5
+	github.com/hashicorp/vault/http => github.com/hashicorp/vault/http v1.7.0
+	github.com/hashicorp/vault/vault => github.com/hashicorp/vault/vault v1.7.0
+
+	// pin this version to avoid later versions that depend on a v1alpha1 branch that's no longer available
```
lornasong, line 64: suggestion to include in the comment that vault is project depending on the pinned k8s version (maybe that's obvious though) and that v0.18.2 is sourced from their go.mod

lornasong (commented):
Thanks so much for adding this benchmarking job and working through the vault dependency issue! We're excited to have this ✨ 

I'm not very familiar with circleci configuration but this makes sense to me. I left a question for the team on the specifics of the frequency of running the benchmarks job. I also left a very minor suggestion to update a comment.

.circleci/config.yml
findkim: I lean towards weekly, but could be something we update after the CI setup is merged. No preference there.

go.mod
shore: Done!

.circleci/config.yml
lornasong: Good idea, we'll adjust as needed later!

lornasong (approved):
Thanks @shore! This looks good to me!

PR #250, opened 2021-04-15
Fix api-serve unit test race condition retry

lornasong:
The retry was no longer working because testutils.RequestHTTP no longer returns
an error and instead checks & handles errors within. The err that was attempted
to be handled for the retry was actually an err from above FreePort().

Change: wait a little after serving the API before testing endpoints. 3s is
used in other places in our code for serving the API and it so has so far
worked well for those tests e.g. api/client_test.go


findkim (approved):
Cleanup!

PR #251, opened 2021-04-15
Enable backport/* label automation

kpenfound:
This adds [backport-assistant](https://github.com/hashicorp/backport-assistant) as a Github Action to enable backport/* label automation

TODO: When this is merged, add `ELEVATED_GITHUB_TOKEN` secret

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,22 @@
+---
+  name: Backport Assistant Runner
+
+  on:
+    pull_request_target:
+      types:
+        - closed
+        - labeled
+      
+  jobs:
+    backport:
+      if: github.event.pull_request.merged
+      runs-on: ubuntu-latest
+      container: hashicorpdev/backport-assistant:0.2.2
+      steps:
+        - name: Run Backport Assistant
+          run: |
+            backport-assistant backport
+          env:
+            BACKPORT_LABEL_REGEXP: "backport/(?P<target>\\d+\\.\\d+\\.\\w+)"
+            BACKPORT_TARGET_TEMPLATE: "releases/{{.target}}"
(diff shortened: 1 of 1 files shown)


findkim (approved):
Looks good thanks @kpenfound for adding the GH action!

.github/workflows/backport.yml
```diff
@@ -0,0 +1,22 @@
...
+    backport:
+      if: github.event.pull_request.merged
+      runs-on: ubuntu-latest
+      container: hashicorpdev/backport-assistant:0.2.2
+      steps:
+        - name: Run Backport Assistant
+          run: |
+            backport-assistant backport
+          env:
+            BACKPORT_LABEL_REGEXP: "backport/(?P<target>\\d+\\.\\d+\\.\\w+)"
```
alvin-huang, line 20: @findkim what are the expected `backport/*` labels and their corresponding `release` branches? 

In consul we use `backport/a.b` -> `release/a.b.x `. This capture group as written might be too wide  since it seems to match `backport/1.2.foobarbaz` when there is probably not a target branch called `releases/1.2.foobarbaz`

To be consistent with consul it should be something like:
```
BACKPORT_LABEL_REGEXP: "backport/(?P<target>\\d+\\.\\d+)"
BACKPORT_TARGET_TEMPLATE: "release/{{.target}}.x"
```
or changing it to something else could work too 😄

findkim: It would be the same as consul, so `backport/a.b` -> `release/a.b.x`

alvin-huang: Cool! @kpenfound would you mind updating the PR with this slight change to the regex config and I can have another look

kpenfound: @alvin-huang updated!

alvin-huang (approved):
LGTM!

PR #254, opened 2021-04-16
Support condition config in task blocks

lornasong:
Currently tasks run depending on service changes. We want to start supporting a
new feature where tasks can be configure to run depending on different
conditions. For the feature, we will add a condition configuration that allows
different condition types. The default type is the current run condition,
service changes.

Commits (see message for more details)
 - Add ConditionConfig interface and ServicesConditionConfig{} (default)
 - Add CatalogServicesConditionConfig{}
 - Update TaskConfig with new Condition field
 - Update configuration decoding to support ConditionConfig

### DIFF
--- a/config/condition.go
+++ b/config/condition.go
@@ -31,7 +31,17 @@ func DefaultConditionConfig() ConditionConfig {
 // both the type and value. Not needed for checking a ConditionConfig
 // implementation i.e. isNil(ConditionConfig), servicesConditionConfig == nil
 func isNil(c ConditionConfig) bool {
-	return c == nil || reflect.ValueOf(c).IsNil()
+	var result bool
+	// switching on type is a performance enhancement
+	switch v := c.(type) {
+	case *ServicesConditionConfig:
+		result = v == nil
+	case *CatalogServicesConditionConfig:
+		result = v == nil
+	default:
+		return c == nil || reflect.ValueOf(c).IsNil()
+	}
+	return c == nil || result
 }
 
 // conditionToTypeFunc is a decode hook function to decode a ConditionConfig
--- a/config/condition.go
+++ b/config/condition.go
@@ -1,0 +1,28 @@
+package config
+
+import (
+	"reflect"
+)
+
+// ConditionConfig configures a condition on a task to define the condition on
+// which to execute a task.
+type ConditionConfig interface {
+	Copy() ConditionConfig
+	Merge(ConditionConfig) ConditionConfig
+	Finalize([]string)
+	Validate() error
+	GoString() string
+}
+
+// DefaultConditionConfig returns the default conditions which is an unconfigured
+// 'services' type condition.
+func DefaultConditionConfig() ConditionConfig {
+	return &ServicesConditionConfig{}
+}
+
+// isNil can be used to check if a ConditionConfig interface is nil by checking
+// both the type and value. Not needed for checking a ConditionConfig
+// implementation i.e. isNil(ConditionConfig), servicesConditionConfig == nil
+func isNil(c ConditionConfig) bool {
+	return c == nil || reflect.ValueOf(c).IsNil()
+}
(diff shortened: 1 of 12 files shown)


config/condition.go
```diff
@@ -0,0 +1,28 @@
...
+// 'services' type condition.
+func DefaultConditionConfig() ConditionConfig {
+	return &ServicesConditionConfig{}
+}
+
+// isNil can be used to check if a ConditionConfig interface is nil by checking
+// both the type and value. Not needed for checking a ConditionConfig
+// implementation i.e. isNil(ConditionConfig), servicesConditionConfig == nil
+func isNil(c ConditionConfig) bool {
+	return c == nil || reflect.ValueOf(c).IsNil()
```
eikenb, line 34: No changes requested, just felt like passing along an alternative to IsNil() that can be handy when performance is important (this is about 10x faster). Apologies if already known.

Instead of reflect.ValueOf(c).IsNil() you can use the knowledge of potential types to speed this up with a switch over the types. Something like...

```
switch v := c.(type) {
    case *ServicesConditionConfig:
        result := v == nil
    case  *SomeOtherType:
        result := v == nil
}
```

eikenb (approved):
Looks good.

config/condition.go
lornasong: I didn't know this! Thanks for sharing! I think this is a great idea. I pushed a change with your suggestion. I kept the reflect.ValueOf() as the default as a back up... (not expecting to hit that code).

eikenb: Ohh... good idea using the reflect as the backup/default. :+1:

config/condition_catalog_services.go
```diff
@@ -0,0 +1,170 @@
...
+		for ix, s := range services {
+			regex[ix] = fmt.Sprintf("^%s$", s) // exact match on service's name
+		}
+		c.Regexp = String(strings.Join(regex, "|"))
+
+		if len(services) == 0 {
+			// for testing. can't occur when running cts due to config validation
+			c.Regexp = String("")
+		}
+	}
```
findkim, line 119: This bit in `Finalize()` and relevant part in `Validate()` might be unnecessary depending on how you plan on working the template usage of this.

For example when regexp == nil, the template could use `/health/services/:service` endpoint instead of the workflow for using regex:
1. `/catalog/services`
2. Filter results based on regex match
3. nested template for `/health/services/:service`

config/condition.go
```diff
@@ -0,0 +1,123 @@
...
+// DefaultConditionConfig returns the default conditions which is an unconfigured
+// 'services' type condition.
+func DefaultConditionConfig() ConditionConfig {
+	return &ServicesConditionConfig{}
+}
+
+// isNil can be used to check if a ConditionConfig interface is nil by checking
+// both the type and value. Not needed for checking a ConditionConfig
+// implementation i.e. isNil(ConditionConfig), servicesConditionConfig == nil
+func isNil(c ConditionConfig) bool {
```
findkim, line 33: Since this is in the `config` package with many types, can you change the function name to be more descriptive to the type it's checking for?

findkim (approved):
Config changes look good overall!

config/condition.go
lornasong: Updated 👍

config/condition_catalog_services.go
lornasong: Interesting! My understanding of switching to the `/health/service/:service` endpoint on `regexp==nil` is that it would no longer be able to support new service de/registrations. It would result to switch to supporting changes in services instances. I might be misunderstanding you!

findkim: Oh great point, I had overlooked that! You are right that'll defeat the purpose of the new services trigger. Sorry for the noise 🙇‍♂️

lornasong: No worries! I appreciate you taking the time to think of improvements ❤️

PR #256, opened 2021-04-19
Update e2e tests to poll API availability

findkim:
Flakey tests due to sleep times had us increasing them every so often. It seems sensitive to the load of the machine and changes to the underlying CTS logic. The new CTS client method allows the test to configure the max timeout
waiting for the API to start up and not have to wait the full duration. The API availability determines that CTS has completed executing the tasks once through successfully.

Reduces the e2e CI job from 2m to 1m32s

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -44,6 +44,43 @@ func NewClient(c *ClientConfig, httpClient httpClient) *Client {
 	}
 }
 
+func (c *Client) Port() int {
+	return c.port
+}
+
+// WaitForAPI polls the /v1/status endpoint to check when the CTS API is
+// available. The API is started when CTS is run as a daemon and completes
+// all tasks once.
+func (c *Client) WaitForAPI(timeout time.Duration) error {
+	polling := make(chan struct{})
+	stopPolling := make(chan struct{})
+	statusAPI := c.Status()
+
+	go func() {
+		var err error
+		for {
+			select {
+			case <-stopPolling:
+				return
+			default:
+				_, err = statusAPI.Overall()
+				if err == nil {
+					polling <- struct{}{}
+					return
+				}
+			}
+		}
+	}()
+
+	select {
+	case <-polling:
+		return nil
+	case <-time.After(timeout):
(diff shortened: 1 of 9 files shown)


api/client.go
```diff
@@ -44,6 +44,43 @@ func NewClient(c *ClientConfig, httpClient httpClient) *Client {
 	}
 }
 
+func (c *Client) Port() int {
```
eikenb, line 47: Added bonus cleanup here. Nice.

eikenb (approved):
I'd approve this twice if I could. Love seeing sleeps removed from tests.

api/client.go
findkim: 🎉 thanks john! It was a fun work-around to think through. The testutil function `StartCTS` runs using a config file written to disk by the caller, but to create the client within the helper it requires the port.

e2e/api_test.go
```diff
@@ -14,6 +14,7 @@ import (
 
 	"github.com/hashicorp/consul-terraform-sync/api"
 	"github.com/hashicorp/consul-terraform-sync/testutils"
+	ctsTestClient "github.com/hashicorp/consul-terraform-sync/testutils/cts"
```
findkim, line 17: I'm not very satisfied with the package name, lmk if any others have thoughts here.

lornasong: My only thought to throw out would be to name the package something like `ctsTest` and rename `StartCTS()` to `StartClient()`. This sort of leans on the idea that we might have more cts-related utils that aren't test-client related. I also was thinking the `StartCTS()`'s usage of "cts" could be perceived as redundant.

Just threw it out in case! ctsTestClient works for me and we can rename down the line as it evolves.

lornasong (commented):
Though post-merge, just wanted to add that I'm excited about this change 🎉

api/client.go
findkim: it worked out nicely to do this refactor after @lornasong had built out the client for easy usage 💯

PR #257, opened 2021-04-19
update hcat version for render-complete race fix

### LINKED ISSUE #239: Services template generation bug
### Describe the bug
When generating the terraform configuration templates for a setup with many (tested 100) services the templates get generated with (many) missing entries.

### Versions
**Consul Terraform Sync**
```
v0.1.0
```

### Configuration File(s)
See the  [100Services](https://github.com/hashicorp/consul-terraform-sync/blob/354ce7a8aa7e434e0c74e3d9d0ecc1e33235b443/e2e/benchmarks/services_test.go#L57) benchmark for a reproduce-able case. Just disable the cleanup to leave the sync-tasks/ and you'll see the bad entry in the generate file `./sync-tasks/test-services-task/terraform.tfvars`.

### Expected Behavior
services entry in terraform.tfvars should list all the services (all 100 in benchmark example).

### Actual Behavior
services entry has 1-3 entries.

### Additional Context
The above benchmark seems to work as there was a [change to hcat](https://github.com/hashicorp/hcat/pull/41) that mitigated the issue to some extent. Doesn't really fix the underlying problem though which I'm filing against hcat next and will link back here.

This is related to #236, an issue about a missing test for this issue.

eikenb:
Fixes #239

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/hashicorp/go-syslog v1.0.0
 	github.com/hashicorp/go-uuid v1.0.2
 	github.com/hashicorp/go-version v1.2.1
-	github.com/hashicorp/hcat v0.0.0-20210401143330-8f813cb572a8
+	github.com/hashicorp/hcat v0.0.0-20210414201846-1b51a2a2651f
 	github.com/hashicorp/hcl v1.0.1-vault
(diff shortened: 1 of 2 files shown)


findkim (approved):
the test from #244 passes 🙌

PR #261, opened 2021-04-21
all but last consul retry log set to DEBUG level

### LINKED ISSUE #260: Reduce Consul reconnect logging
#### Description
Reduce the logging for each Consul retry on service monitoring when an agent is unavailable.

When CTS is monitoring 100 services and the Consul agent is unavailable, CTS could log a total of 100 lines * 8 retries.

#### Additional context
A task with 2 services
```
2021/04/21 17:45:01.462862 [INFO] (ctrl) all tasks completed once
2021/04/21 17:45:01.462869 [INFO] (cli) running controller in daemon mode
2021/04/21 17:45:01.463423 [INFO] (api) starting server at '2000'
2021/04/21 17:45:04.523482 [WARN] (hcat) error connecting with Consul. Waiting 1.364767007s to retry attempt #1
2021/04/21 17:45:04.523496 [WARN] (hcat) error connecting with Consul. Waiting 1.346008816s to retry attempt #1
2021/04/21 17:45:05.875653 [WARN] (hcat) error connecting with Consul. Waiting 2.256786706s to retry attempt #2
2021/04/21 17:45:05.893154 [WARN] (hcat) error connecting with Consul. Waiting 2.687462695s to retry attempt #2
...
```

eikenb:
Consul's retries were logging every retry for every watched field. It
was very verbose. This quiets it down by changing all the retry logs to
DEBUG level except for the last one which is (unchanged from) ERR level.

Fixes #260

### DIFF
--- a/controller/watcher.go
+++ b/controller/watcher.go
@@ -73,7 +73,7 @@ func retryConsul(retryCount int) (bool, time.Duration) {
 	random := rand.New(rand.NewSource(time.Now().UnixNano()))
 	wait := retry.WaitTime(uint(retryCount), random)
 	dur := time.Duration(wait)
-	log.Printf("[WARN] (hcat) error connecting with Consul. Waiting %v to retry"+
+	log.Printf("[DEBUG] (hcat) error connecting with Consul. Waiting %v to retry"+
 		" attempt #%d", dur, retryCount+1)
 	return true, dur
 }


controller/watcher.go
```diff
@@ -73,7 +73,7 @@ func retryConsul(retryCount int) (bool, time.Duration) {
 	random := rand.New(rand.NewSource(time.Now().UnixNano()))
 	wait := retry.WaitTime(uint(retryCount), random)
 	dur := time.Duration(wait)
-	log.Printf("[WARN] (hcat) error connecting with Consul. Waiting %v to retry"+
+	log.Printf("[DEBUG] (hcat) error connecting with Consul. Waiting %v to retry"+
```
lornasong, line 76: For operators who have above debug-level logs configured, there will be some silence in the logs while this retries. I wonder if this could be confusing to operators? Or maybe this is expected experience for those configure above debug-level? Just curious what your thoughts are!

lornasong (approved):
Makes sense to me! Left a non-blocking question to get more context on the operator experience 🙏

controller/watcher.go
eikenb: Most logs are reviewed after the fact where they'd see the initialization and the final failed retry. If they're watching the logs more actively they'd probably have enabled more verbose logging.

We could bump it up to INFO level or down to TRACE level... DEBUG level seemed the best at the time. IMO INFO is one of those in-between values that never seems that useful and TRACE is usually very verbose.

lornasong: Ah, good point about operators setting the log-level depending on if they're looking after the fact or actively! Thanks for distinguishing those two and sharing your thoughts. It was helpful for me to think through. I agree with you that DEBUG seems to be the best level.

PR #265, opened 2021-04-23
Support catalog-services condition in templating

lornasong:
This PR is the initial change to consume condition configuration (added in https://github.com/hashicorp/consul-terraform-sync/pull/254) in the terraform template file.

When the default services condition is configured, no changes are made to the template file. When the new catalog-services condition is configured, an additional template block is added to the template file. This new template block currently does not render anything and will be used internally to identify when configured catalog-services have been updated.

Commits (see messages for more details):
 - Pull in latest hcat changes needed to template catalog-services
 - Add condition structures to templating package + new catalog-services condition
 - Consume conditions in templating package
 - Pass condition configuration through controller, driver, to templating

Future work: enable_tf_vars; filter catalog-services by datacenter, namespaces, node_meta


templates/tftmpl/golden_test.go
```diff
@@ -87,18 +87,33 @@ func TestNewFiles(t *testing.T) {
...
 			Golden: "testdata/terraform.tfvars.tmpl",
 			Input: RootModuleInputData{
-				Providers: []hcltmpl.NamedBlock{hcltmpl.NewNamedBlock(
-					map[string]interface{}{
-						"testProvider": map[string]interface{}{
-							"alias": "tp",
-							"attr":  "value",
-							"count": 10,
-						},
-					})},
```
lornasong, line 101: provider configs weren't necessary for the test-case so I removed it just to trim down the test-case a bit

templates/tftmpl/condition.go
```diff
@@ -0,0 +1,54 @@
...
+// This is the default run condition
+type ServicesCondition struct{}
+
+func (c *ServicesCondition) appendTemplate(w io.Writer) error {
+	// no-op: services conditon currently requires no additional condition
+	// templating. it relies on the monitoring template as the run condition
+	return nil
+}
+
+var _ Condition = (*CatalogServicesCondition)(nil)
```
findkim, line 27: Can these be grouped at the top of the file near the interface? It helps me quickly identify what classes are in file that satisfy `Condition`

```go
var (
  _ Condition = (*...)
  _ Condition = (*...)
)
```

go.sum
```diff
@@ -758,6 +758,8 @@ github.com/hashicorp/golang-lru v0.5.4 h1:YDjusn29QI/Das2iO9M0BHnIbxPeyuCHsjMW+l
 github.com/hashicorp/golang-lru v0.5.4/go.mod h1:iADmTwqILo4mZ8BN3D2Q6+9jd8WM5uGBxy+E8yxSoD4=
 github.com/hashicorp/hcat v0.0.0-20210414201846-1b51a2a2651f h1:L/fJ24aVu/TbYV4AyoMfzir8HCbDDfynXhoZLsLUzrM=
 github.com/hashicorp/hcat v0.0.0-20210414201846-1b51a2a2651f/go.mod h1:AJ2fdGudt/27fXjQBl9srE7mOV9fGZnjVxua/CpjT3I=
```
findkim, line 760: Would running `go mod tidy` remove these and use the added ones below?

driver/terraform.go
```diff
@@ -361,6 +362,25 @@ func (tf *Terraform) initTask(force bool) error {
...
+			Regexp:      *v.Regexp,
+			EnableTfVar: *v.EnableTfVar,
+			Datacenter:  *v.Datacenter,
+			Namespace:   *v.Namespace,
+			NodeMeta:    v.NodeMeta,
+		}
+	case *config.ServicesConditionConfig:
+		condition = &tftmpl.ServicesCondition{}
+	default:
+		// expected only for test scenarios
```
findkim, line 378: Nice!! this beats updating all of the tests to anticipate this addition.

templates/tftmpl/testdata/terraform_cs.tfvars.tmpl
```diff
@@ -0,0 +1,30 @@
...
+# The HCL blocks, arguments, variables, and values are derived from the
+# operator configuration for Sync. Any manual changes to this file
+# may not be preserved and could be overwritten by a subsequent update.
+#
+# Task: test
+# Description: user description for task named 'test'
+
+{{- with $allCatalogServices := services }}
+  {{- range $cs := $allCatalogServices }}
+    {{- if regexMatch ".*" $cs.Name }}{{/* Empty template. Detects changes in catalog-services */}}
```
findkim, line 12: The empty template is an interesting work-around, TIL!

Would this be replaced with the `enable_tf_vars` functionality later or would it using a different template block?

driver/task.go
```diff
@@ -55,6 +56,7 @@ type Task struct {
 	Version         string
 	UserDefinedMeta map[string]map[string]string
 	BufferPeriod    *BufferPeriod // nil when disabled
+	Condition       config.ConditionConfig
```
findkim, line 59: No-op, but wanted to note my thoughts. When we start seeing this become more of a pattern to import the config package for using the underlying objects, we may need to look into a dedicated object package so the config package doesn't end up so pervasive in every core code paths.

findkim (approved):
It's neat to see how minimal the template addition is to support the condition feature

templates/tftmpl/condition.go
lornasong: Sure thing!

go.sum
lornasong: Good call! It cleaned them up :)

templates/tftmpl/testdata/terraform_cs.tfvars.tmpl
lornasong: Good question! As of now, I was thinking it would be used when `enable_tf_vars=false` and then a different template block for when `enable_tf_vars=true`. Also open to changing this especially after seeing how the hcat changes work to detect catalog-services updates. Let me know if you have any thoughts!

driver/task.go
lornasong: Agree! Thanks for mentioning this. 🤔 I spent awhile debating whether or not to reuse the config object here. On one hand it didn't seem worth it to build out a Condition interface with Services and CatalogServices implementations just to thread condition information through. On the other hand, to your point, importing config around isn't great. I think having a dedicated object package is a good idea that would solve this dilemma!

findkim: Agreed on the trade offs you mentioned! We can continue keeping tabs on this to see how it plays out later

templates/tftmpl/testdata/terraform_cs.tfvars.tmpl
findkim: Thanks for sharing your plans. I don't know exactly how nesting conditional templating works and if the the `enable_tf_vars` evaluation was a part of this inner portion would add unnecessary overhead for the false case. I'm mostly interested in following along to see how it works and don't have thoughts that lean one way over another.

lornasong: Sounds good! If you have thoughts in the future, don't hesitate to share!

PR #268, opened 2021-04-23
Update tfvars map to use "key = value"

findkim:
This updates the CTS generated Terraform variables to use the suggested format from the documentation instead of `key : value`. No breaking change, but all generated templates and tfvars will be re-rendered with `=` now.

https://www.terraform.io/docs/language/expressions/types.html#maps-objects

> Maps/objects are represented by a pair of curly braces containing a series of `<KEY> = <VALUE>` pairs:
```hcl
services = {
  "api.worker-01.dc1" = { }
}
```


lornasong (approved):
looks good to me!

PR #274, opened 2021-04-29
Rename catalog-services condition 'enable_tf_var' to 'source_includes_var'

lornasong:
The `enable_tf_var` field is a boolean that should be set based on whether or
not the task's configured source module includes the catalog_services variable
in the terraform `variables.tf` file.

"enable" suggests that this field is optional. However, the field value isn't
truly optional since it's dependent on the source module. "tf_var" language
seems more module-author friendly, but the person writing the config may be a
different person.

Considerations for `source_includes_var`:
 - "source" hints that this field should be configured based on the task's
 configured source
 - "var" abbreviates the "variable" terminology in the source module
 - "includes" describes the module to variable relationship
 - can be interpreted as "does the source include var?" which hopefully
 increases intuition in configuring this field

Note: "enable_tf_var" is not yet consumed and will be in subsequent pr


findkim (approved):
I appreciate the rationale breakdown for the renaming. The new `source_includes_var` is more descriptive than `enable_tf_far` in terms of the purpose it toggles -- that the module makes use of the variable and for CTS to generate a terraform input variable and set it as a module argument

PR #275, opened 2021-04-30
Support for catalog-services condition 'source_includes_var'

lornasong:
Consumes catalog-services condition `source_includes_var` configuration field added in https://github.com/hashicorp/consul-terraform-sync/pull/254. For example:

```
task {
 	name = "catalog_task"
 	services = ["api"]
 	source = "../some/module"
 		source_includes_var = true // here!
 	}
}
```

`source_includes_var` must be set true when a task's source module includes usage of the catalog-services variable. This requires CTS to pass along the catalog-services variable/information in the auto-generated files: main.tf, variables.tf, and terraform.tfvars.tmpl

Commits (see messages for more details):
 - Update terraform.tfvars.tmpl to use different templating which adds the catalog_services variable
 - Add templating unit/integration tests
 - Update main.tf to pass catalog_services variable
 - Update variables.tf to add catalog_services variable definition
 - e2e test


e2e/condition_test.go
```diff
@@ -0,0 +1,55 @@
...
+	tempDir := fmt.Sprintf("%s%s", tempDirPrefix, "cs_condition_include")
+	delete := testutils.MakeTempDir(t, tempDir)
+
+	conditionTask := `task {
+	name = "catalog_task"
+	services = ["api"]
+	source = "../../test_modules/local_tags_file"
+	condition "catalog-services" {
+		regexp = "db|web"
+		source_includes_var = true
```
eikenb, line 32: Aha! I missed this on my first reading. The source_includes_var is set in the section for the conditional variable, not a more global setting as I imagined. This makes more sense to me know.

(sorry community member, we talked about this in slack but basically I was confused about how this was used)

lornasong: Gotcha! Ok, I'm glad to hear that this cleared it up! I considered adding an example config in the PR description but decided against it, darn. Seems like it would have been helpful for contextualizing 'source_includes_var'. I'll add it anyways after the fact. Thanks for the feedback!

templates/tftmpl/template_funcs.go
```diff
@@ -42,6 +45,22 @@ func joinStringsFunc(sep string, values ...string) string {
 	return strings.Join(cleaned, sep)
 }
 
+// hclServicesFunc is a wrapper of the template function to marshal Consul
+// catalog service tag information into HCL. It returns the list of tags with
+// formatted like: "["tag1", "tag2"]". It returns an empty array string "[]"
+// when no tags
+func hclServicesFunc() func(tags *dep.ServiceTags) string {
```
findkim, line 52: `HCLServices` and `hclServicesFunc()` function names are strongly tied to the `services` template function name, but it was not intuitive for me to realize this formats just the tags for a single service until reading the signature. Could you rename this to reflect the input/output of the function and relate it to "tags"? The "services" portion of the name makes it seem like the parameter should be tags for multiple services, perhaps something like `hclServiceTagsFunc()`

findkim (approved):
Love all the tests you added for the condition feature. The changes to the `tftmpl` package are straight forward and concise!

templates/tftmpl/template_funcs.go
lornasong: Good point, the name matches the template function name but isn't intuitive. I like hclServiceTagsFunc()!  Changed!

PR #276, opened 2021-05-03
Support for catalog-services condition query parameter

lornasong:
Currently, CTS supports catalog-services condition with no filtering on the Catalog-Services API. This PR adds query filtering by datacenter, namespace, and node-meta.

Commits:
 - Pull latest hcat changes to support querying by namespace and node-meta (datacenter already supported)
 - Update templating to include query filtering


findkim (approved):
Nice, everything looks threaded together well.

PR #278, opened 2021-05-11
Support catalog-services condition name regex filtering

lornasong:
For catalog-service condition, we want to be able to filter catalog services by
regex such that only [de]registration of catalog services with names that match
the configured regex trigger the task. However, the current state of this feature
will trigger the task when a service name does not match the regex (though the
task will make no changes when triggered).

This PR adds a custom hcat template function to do this custom regex filtering
on the catalog service api results. In the following PR, this custom template func
will also handle tag trigger suppression, so changes for this future feature are
considered in this current pr (particularly around naming).

Custom template function usage:
`{{ catalogServicesRegistration "regexp=.*" "dc=dc1" }}`

Commits:
 - Add custom template function to do regex filtering (and will handle future tag
 tag trigger suppression) in a new directory
 - Refactor existing template function code to new directory
 - Consume new custom template function in place of `regexMatch`

Custom template function naming thoughts below. Feedback welcome!
 - "catalogServices": wanted to differentiate from `services` since this template
 function has its own custom additions. But still wanted to have naming link back
 to the Catalog Services API
 - "Registration": to emphasize that this template function differs by specifically
 triggering on registration information (and not tag info)


templates/tftmpl/tmplfunc/hcat.go
```diff
@@ -0,0 +1,119 @@
...
+	"sort"
+	"time"
+
+	consulapi "github.com/hashicorp/consul/api"
+)
+
+// QueryOptions is a list of options to send with the query. These options are
+// client-agnostic, and the dependency determines which, if any, of the options
+// to use.
+type QueryOptions struct {
```
findkim, line 18: A lot of this is replicating the additional hcat layer of query options. Do you know if there is a way it work to use the `consulapi.QueryOptions` directly? would it simplify the changes?

e2e/condition_test.go
```diff
@@ -53,3 +63,212 @@ func TestCondition_CatalogServices_Include(t *testing.T) {
...
+
+	// Test that node-meta filter is filtering service registration information
+	// and task triggers
+	// 0. Confirm baseline: check current number of events and that that no
+	//    catalog_service variable contains no service information
+	// 1. Register service with server with no node-meta. Confirm that the task
+	//    was not triggered (no new event) and data is filtered out of
+	//    catalog_service.
+	// 2. Register service with server with configured node-meta. Confirm that
+	//    task was triggered (one new event) and its data exists in catalog_service.
```
findkim, line 220: This is a good e2e test, and really appreciate the human-readable summary.

e2e/condition_test.go
```diff
@@ -53,3 +63,212 @@ func TestCondition_CatalogServices_Include(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Test that regex filter is filtering service registration information and
+	// task triggers
+	// 0. Confirm baseline: check current number of events and that that no
+	//    catalog_service variable contains no service information
+	// 1. Register db service instance. Confirm that the task was not triggered
+	//    (no new event) and its data is filtered out of catalog_service.
+	// 2. Register api-web service instance. Confirm that task was triggered
+	//    (one new event) and its data exists in catalog_service.
```
findkim, line 115: This is a good e2e test, and really appreciate the human-readable summary.

e2e/condition_test.go
```diff
@@ -53,3 +63,212 @@ func TestCondition_CatalogServices_Include(t *testing.T) {
...
+}
+
+// TestCondition_CatalogServices_NodeMeta runs the CTS binary. It specifically
+// tests a task configured with a catalog service condition with the node-meta
+// configuration set. This test confirms that when a service is registered with
+// a Consul server that does not have the configured node-meta:
+//  1) the task is not triggered (determined by confirming from the Task Status
+//     API that an event had not been added)
+//  2) the service information does not exist in the tfvars file
+func TestCondition_CatalogServices_NodeMeta(t *testing.T) {
```
findkim, line 159: I like how you are testing the filtering functionality. It reads to me that the core portion being tested is the query params and Consul API response is filtered as expected. A (not required) improvement could be to scope down this test and change it from an e2e test to an integration test within the `templates/tftmpl/tmplfunc` package to test the query parameter `node-meta` specifically.

findkim (approved):
The context you had provided in the summary was really helpful for me to digest the changes. The approach and code look good!

Approving and feel free to address the other comments I added as you see fit.

templates/tftmpl/tmplfunc/tmpl_func.go
```diff
@@ -9,20 +9,17 @@ import (
...
-	"github.com/hashicorp/hcl/v2/hclwrite"
 )
 
-// HCLTmplFuncMap is the map of template functions for rendering HCL
+// HCLMap is the map of template functions for rendering HCL
 // to their respective implementations
-func HCLTmplFuncMap(meta map[string]map[string]string) template.FuncMap {
+func HCLMap(meta map[string]map[string]string) template.FuncMap {
 	tmplFuncs := hcat.FuncMapConsulV1()
+	tmplFuncs["catalogServicesRegistration"] = catalogServicesRegistrationFunc
```
findkim, line 18: I like the template function name!

templates/tftmpl/tmplfunc/hcat.go
lornasong: Great idea!! I tried it out and was able to replace this `QueryOptions` with the `consulapi.QueryOptions` and that cleaned up all this query option code I copied over 🎉. Thanks for the suggestion!

One thing that I debated for awhile is that `consulapi.QueryOption` doesn't have a merge-type function. I tried writing a merge function initially, but because `consulapi.QueryOption` doesn't support nil fields, it couldn't handle some empty values accurately. For now, I opted to write a custom merge within Fetch() since it was less code and a little more accurate.

🤔 I know we typically follow the pattern of having a merge function. I was thinking we could revisit this if we add more template funcs in the future that need to share a merge-type func. Going to go ahead and merge but happy to make changes if you have concerns after-the-fact.

e2e/condition_test.go
lornasong: I thought of writing this as an integration test but didn't want to set up the test client and went with the lazy route of writing an e2e 😅. Thanks for calling this out! I completely agree. This shouldn't be an e2e test. I'll add this on my todo list for revisit this.

templates/tftmpl/tmplfunc/hcat.go
eikenb: Using the direct API query options is probably the best way, but if using the internal QueryOptions ends up being the best solution here please let me know and we can look into making it public in hashicat so you wouldn't need all the copied code.

lornasong: Thanks @eikenb, will do!

e2e/condition_test.go
lornasong: Revisited here https://github.com/hashicorp/consul-terraform-sync/pull/310!

Issue #264, opened 2021-04-23
Confusing task source for local modules


findkim:
### Describe the bug
Configuring a task to use a local module is not intuitive. It takes some trial and error to figure out that the expected local path is from the working directory, e.g. the default path `./sync-task/<task-name>`

### Versions
**Consul Terraform Sync**

This issue is present in all versions so far
```
consul-terraform-sync v0.1.1 (4dc13b5)
Compatible with Terraform >= 0.13.0, < 0.15
```

**Consul Version**
any

**Terraform Version**
any

### Configuration File(s)

```hcl
task {
  name = "example"
  description = "this task prints output and creates files with content"
  source = "../../my-local-module"
  services = ["api", "web"]
}
```

### Expected Behavior

Source supports local path to where CTS is run

```hcl
task {
  name = "example-task"
  description = "this task prints output and creates files with content"
  source = "./my-local-module"
  services = ["api", "web"]
}
```

### Actual Behavior
The module directory cannot be found by Terraform when run by CTS in the working directory `sync-tasks/<task-name>`

### Steps to Reproduce
1. In the current working directory, run CTS with one task configured from the current working directory to the local path of a module, `source = "./my-local-module"`
2. The `./sync-tasks/<task-name>` working directory is created on startup
3. CTS exits with a Terraform error that the module cannot be found

```
2021/04/23 14:21:34.920139 [TRACE] (driver.terraform) initializing workspace 'example-task'
2021/04/23 14:21:35.009096 [INFO] running Terraform command: /Users/kngo/dev/hashicorp/consul-terraform-sync/terraform init -no-color -force-copy -input=false -lock-timeout=0s -backend=true -get=true -upgrade=false -lock=true -get-plugins=true -verify-plugins=true
Initializing modules...
- example-task

Error: Unreadable module directory

Unable to evaluate directory symlink: lstat my-local-module: no such file or directory


Error: Failed to read module directory

Module directory  does not exist or cannot be read.


Error: Unreadable module directory

Unable to evaluate directory symlink: lstat my-local-module: no such file or directory


Error: Failed to read module directory

Module directory  does not exist or cannot be read.

2021/04/23 14:21:35.097094 [ERR] (driver.terraform) error initializing workspace, skipping apply for 'example-task'
```

### Additional Context
CTS could expand the local path for a task to the full path when generating the module block in `main.tf`. This would introduce a breaking change to the current behavior but it was buggy to begin with

PR #283, opened 2021-05-19
Update expected relative path of task source for local modules

mkam:
**Breaking Change:**
When using a local module for a task, the relative path provided as the source is now based on the working directory of CTS. Previously, the path was based on the task directory (e.g., `sync-tasks/<task-name>`).

Operators with tasks that use local modules will need to update the task's source variable. For example, when running CTS with the following directory structure:
```
.
| -- config.hcl
| -- my-local-module/
| -- sync-tasks/example-task  
```
Previous to this change, the source would have been specified as follows:
```hcl
task {
  name = "example-task"
  description = "this task prints output and creates files with content"
  source = "../../my-local-module"
  services = ["api", "web"]
}
```
With this change, the source needs to be updated to:
```hcl
task {
  name = "example-task"
  description = "this task prints output and creates files with content"
  source = "./my-local-module"
  services = ["api", "web"]
}
```

Closes https://github.com/hashicorp/consul-terraform-sync/issues/264

### DIFF
--- a/templates/tftmpl/root_test.go
+++ b/templates/tftmpl/root_test.go
@@ -1,6 +1,9 @@
 package tftmpl
 
 import (
+	"fmt"
+	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/hashicorp/consul-terraform-sync/config"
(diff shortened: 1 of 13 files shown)


templates/tftmpl/root_test.go
```diff
@@ -170,6 +173,125 @@ func TestAppendRootProviderBlocks(t *testing.T) {
 	}
 }
 
+func TestAppendRootModuleBlocks(t *testing.T) {
+	workingDir, err := os.Getwd()
+	assert.Nil(t, err, "Error determining current working directory")
```
findkim, line 178:
```suggestion
	require.Nil(t, err, "Error determining current working directory")
```

One minor nit is to change this assertion to require since the rest of the test would never pass if this had errored. `assert` continues running through the rest of the test even if it fails.

findkim (approved):
I like the location of where the path change happens. Good callout on the breaking change too.

PR #284, opened 2021-05-20
Relax task.services config requirement when catalog-service condition regexp configured

lornasong:
Tasks with a catalog-service condition configured may not need to monitor
service instances, which is configured in task.services. When a catalog-service
condition regexp field is configured, the task monitors and triggers on matches
to the regexp. When catalog-service condition regexp field is not configured,
the regexp value defaults to matching on the task.services values.

Therefore, we want to relax the task.services config requirement only when
catalog-service regexp is configured.

Commits:
 - Changes to task config and catalog-service condition config to loosen
 requirement
 - Update templating to handle when no services are configured
 - Update e2e tests that don't require task.services


config/task_test.go
```diff
@@ -374,10 +387,37 @@ func TestTaskConfig_Validate(t *testing.T) {
...
+			false,
+		},
+		{
+			"invalid catalog-service condition (bad regexp)",
+			&TaskConfig{
+				Name:      String("task"),
+				Source:    String("source"),
+				Condition: &CatalogServicesConditionConfig{Regexp: String("*")},
+			},
+			false,
```
lornasong, line 419: this test case is unrelated to the feature. i noticed it missing when looking at coverage so snuck it in 🙏

config/condition_catalog_services.go
```diff
@@ -96,13 +96,15 @@ func (c *CatalogServicesConditionConfig) Merge(o ConditionConfig) ConditionConfi
 	return r2
 }
 
-// Finalize ensures there no nil pointers.
+// Finalize ensures there no nil pointers with the _exception_ of Regexp. There
```
lornasong, line 99: I didn't really like making this exception here but couldn't figure out another way. Ideas welcome!

lornasong:
^ resolved conflict!

findkim (approved):
Changes look good!

config/task_test.go
findkim: out of curiosity, what error is surfaced when `*` is set?

config/condition_catalog_services.go
findkim: This exception strikes me as fine!

config/task_test.go
lornasong: Yea, it surprised me that `*` errors. `regexp.Compile("*")` returns the error "error parsing regexp: missing argument to repetition operator: `*`"

findkim: oh cool ok that seems informative enough that it should be `.*` for ppl somewhat familiar with regex. Thanks!

lornasong: Yea, agree the message is pretty helpful. Thanks for checking on this! @ eikenb also suggested, which I thought would be really helpful for users, that I add a link to a regex intro or something in our documentation. There've been a couple other regex surprises for me personally

PR #285, opened 2021-05-20
Resolve vault dep conflict

findkim:
This removes the conflicting dependency import of Vault (with go mod and pre-go mod) and only imports the sub modules `vault/api` and `vault/sdk`.
* Removed usage of `vault/vault` and `vault/http` packages that are not intended to be imported.
* Added `testutils.NewTestVaultServer` function to startup Vault in dev mode for integration tests
* Moved vault integration tests to a separate build tag `-tags 'integration vault` so installing vault is not always a requirement to run integration tests.
* Updated CircleCI to run Vault integration tests separately and weekly since there is not many tests for it and is not in the critical path to install vault each CI run.
* Moves `FreePort` from api to `testutils` package since it's only used in tests
* Moves `StartCTS` test helper to the `api` package to resolve import cycles
  * Previously, testutils/cts -> `testutils.FreePort()` and `api.Client`


findkim:
The vault integration job passes [here](https://circleci.com/gh/hashicorp/consul-terraform-sync/2016?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) before I had changed it to run weekly

testutils/vault.go
```diff
@@ -0,0 +1,64 @@
+// +build integration
+// +build vault
```
findkim, line 2: TIL to require both tags go expects separate lines for logical AND, and if on the same line it uses the logical OR operator.

lornasong (approved):
🎉 getting rid of the vault dependency! the new vault test and client look great. we chatted offline about seeing if it was potentially possible to now also get rid of the k8s block in the go.mod replace block. so double-checking on that in case it's possible. either way, this looks good, approving!

findkim:
@lornasong woohoo removing the k8 replace block works. I'm happy to finally resolve this pain point.

lornasong:
@findkim so beautiful ✨✨✨ thanks for sitting down and taking the time to resolve this

PR #286, opened 2021-05-20
Copy HCL2ValueFromConfigValue and remove TF dep

findkim:
[`HCL2ValueFromConfigValue`](https://github.com/hashicorp/terraform/blob/d35bc0531255b496beb5d932f185cbcdb2d61a99/internal/configs/hcl2shim/values.go) is handy in parsing HCL config into generating Terraform syntax. This removes the nasty dependency on importing the whole terraform project for one helper function.


lornasong (commented):
looks good. deleted 150 lines from go.sum 🎉

lornasong (approved):

Issue #279, opened 2021-05-11
Relative paths not supported for variable_files 


mkam:
### Describe the bug
The `variable_files` attribute in the `task` block only works with absolute paths. This is not stated as a requirement in the documentation and also differs from the behavior of the `source` attribute in the `task` block, which does support relative paths.

### Versions
**Consul Terraform Sync**
```
consul-terraform-sync v0.1.0-97-g388259b (388259b)
Compatible with Terraform >= 0.13.0, < 0.16
```

**Consul Version**
```
Consul v1.9.5
```

**Terraform Version**
```
Terraform v0.15.3
```

### Configuration File(s)
```hcl
task {
  name = "example-task"
  description = "Writes the service name, id, and IP address to a file"
  source = "../../example-module"
  providers = ["local"]
  services = ["web", "api"]
  variable_files = ["example.module.tfvars"]
}
```

### Task Variable Files
The directory structure is as follows:
```
.
|-- main.tf
|-- config.hcl
|-- example.module.tfvars
|-- example-module/
```

### Expected Behavior
CTS loads the variable file using the relative path.

### Actual Behavior
CTS fails saying that the file does not exist.
```
$ consul-terraform-sync -config-file=config.hcl -once
...
2021/05/11 22:27:32.523797 [INFO] (driver.terraform) retrieved 0 Terraform handlers for task 'example-task'
2021/05/11 22:27:32.524149 [INFO] (templates.tftmpl) overwriting providers.tfvars in root module for task "example-task"
2021/05/11 22:27:32.544510 [INFO] (templates.tftmpl) overwriting main.tf in root module for task "example-task"
2021/05/11 22:27:32.563662 [INFO] (templates.tftmpl) overwriting variables.tf in root module for task "example-task"
2021/05/11 22:27:32.582683 [INFO] (templates.tftmpl) overwriting variables.module.tf in root module for task "example-task"
2021/05/11 22:27:32.602675 [INFO] (templates.tftmpl) overwriting terraform.tfvars.tmpl in root module for task "example-task"
2021/05/11 22:27:32.622840 [INFO] (ctrl) driver initialized
2021/05/11 22:27:32.622866 [INFO] (ctrl) executing all tasks once through
2021/05/11 22:27:32.732026 [INFO] (ctrl) executing task example-task
2021/05/11 22:27:33.678696 [ERR] (cli) error running controller in Once mode: could not apply changes for task example-task: error tf-apply for 'example-task': exit status 1

[...]

PR #288, opened 2021-05-21
Support relative paths for task variable files

mkam:
Expands relative paths to full paths for the task variable files 
to support both relative and full paths.

Closes https://github.com/hashicorp/consul-terraform-sync/issues/279

### DIFF
--- a/client/terraform_cli.go
+++ b/client/terraform_cli.go
@@ -9,7 +9,6 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
-	"strings"
 
 	"github.com/hashicorp/consul-terraform-sync/templates/tftmpl"
 	"github.com/hashicorp/terraform-exec/tfexec"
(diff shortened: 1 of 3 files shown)


client/terraform_cli.go
```diff
@@ -77,11 +79,28 @@ func NewTerraformCLI(config *TerraformCLIConfig) (*TerraformCLI, error) {
 		log.Printf("[INFO] (client.terraformcli) persiting Terraform logs on disk: %s", logPath)
 	}
 
+	// Expand any relative paths for variable files to absolute paths
+	var varFiles []string
+	for _, vf := range config.VarFiles {
+		if !strings.HasPrefix(vf, "/") {
```
eikenb, line 85: Consider using `path.IsAbs()` for this. A dedicated method makes the purpose more obvious.

eikenb: Actually `filepath.IsAbs()` would probably be better. Already imported and better cross platform support.

client/terraform_cli.go
```diff
@@ -77,11 +79,28 @@ func NewTerraformCLI(config *TerraformCLIConfig) (*TerraformCLI, error) {
...
+			wd, err := os.Getwd()
+			if err != nil {
+				log.Println("[ERR] (client.terraformcli) unable to retrieve current " +
+					"working directory to determine path to variable files")
+				log.Panic(err)
+			}
+			vfAbs := filepath.Join(wd, vf)
+			varFiles = append(varFiles, vfAbs)
+		} else {
+			varFiles = append(varFiles, vf)
```
eikenb, line 95: `filepath.Join()` calls `filepath.Clean()` internally. Might be a good idea to Clean() these as well for consistency?

eikenb (approved):
LGTM.

findkim (approved):
Changes look good! I'm excited for consistency to support absolute and relative paths across the different configuration options 🎉

PR #289, opened 2021-05-25
Refactor driver.Task to be thread safe

findkim:
Task attributes are now only readable by the API and controller.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -47,7 +47,7 @@ type unit struct {
 
 type baseController struct {
 	conf      *config.Config
-	newDriver func(*config.Config, driver.Task, templates.Watcher) (driver.Driver, error)
+	newDriver func(*config.Config, *driver.Task, templates.Watcher) (driver.Driver, error)
 	units     []unit
 	watcher   templates.Watcher
 	resolver  templates.Resolver
@@ -88,7 +88,10 @@ func (ctrl *baseController) init(ctx context.Context) (*driver.Drivers, error) {
 
 	// Future: improve by combining tasks into workflows.
 	log.Printf("[INFO] (ctrl) initializing all tasks")
-	tasks := newDriverTasks(ctrl.conf, providerConfigs)
+	tasks, err := newDriverTasks(ctrl.conf, providerConfigs)
+	if err != nil {
+		return nil, err
+	}
 	units := make([]unit, 0, len(tasks))
 	drivers := driver.NewDrivers()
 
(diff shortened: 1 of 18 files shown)


controller/readonly.go
```diff
@@ -87,10 +87,14 @@ func (ctrl *ReadOnly) Run(ctx context.Context) error {
 
 func (ctrl *ReadOnly) checkInspect(ctx context.Context, u unit) (bool, error) {
 	taskName := u.taskName
+	d := u.driver
+	if !d.Task().IsEnabled() {
```
lornasong, line 91: Similar to a previous comment. Just wanted to make sure I understood if this change is it related to the thread-safe refactor or fixing something separately maybe?

driver/terraform_test.go
```diff
@@ -510,56 +511,53 @@ func TestGetTerraformHandlers(t *testing.T) {
 		name        string
 		expectError bool
 		nilHandler  bool
-		task        Task
+		providers   TerraformProviderBlocks
```
lornasong, line 514: +1 thank you for simplifying this!

driver/terraform.go
```diff
@@ -550,10 +494,10 @@ func (tf *Terraform) initTaskTemplate() error {
 //
 // Returned handler may be nil even if returned err is nil. This happens when
 // no providers have a handler.
-func getTerraformHandlers(task Task) (handler.Handler, error) {
+func getTerraformHandlers(name string, providers TerraformProviderBlocks) (handler.Handler, error) {
```
lornasong, line 497: I like that you changed this to only pass in what's needed! Optional thought - someone using this function might not know what `name` refers to without looking into the function's details (e.g. it might be handler name) - would something like taskName work ok?

driver/task.go
```diff
@@ -44,54 +47,295 @@ type BufferPeriod struct {
...
+		input.Variables[k] = v
+	}
+}
+
 // clientConfig configures a driver client for a task
 type clientConfig struct {
-	task       Task
 	clientType string
 	log        bool
+	name       string
```
lornasong, line 327: optional thought - I could see myself not being sure if `name` meant client name. Would maybe taskName work ok to clarify 'whose' name should be configured here?

lornasong (commented):
so much great refactoring!

controller/controller.go
```diff
@@ -100,27 +103,30 @@ func (ctrl *baseController) init(ctx context.Context) (*driver.Drivers, error) {
...
 		d, err := ctrl.newDriver(ctrl.conf, task, ctrl.watcher)
 		if err != nil {
 			return nil, err
 		}
 
-		err = d.InitTask(true)
-		if err != nil {
-			log.Printf("[ERR] (ctrl) error initializing task %q: %s", task.Name, err)
-			return nil, err
+		if task.IsEnabled() {
```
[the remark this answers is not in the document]
findkim: ah sorry this was a half-baked idea that I should have removed since I wasn't planning to make any enable changes. Thanks for calling it out. I'll remove it!

controller/readonly.go
findkim: This one I had added because I thought it was missing after looking at the [`ReadWrite.checkApply` method](https://github.com/hashicorp/consul-terraform-sync/blob/master/controller/readwrite.go#L165-L178). After discussion it was added there to avoid event tracking. So it seems this is not needed here, will remove!

controller/controller.go
```diff
@@ -252,10 +258,14 @@ func newDriverTasks(conf *config.Config, providerConfigs driver.TerraformProvide
 			UserDefinedMeta: conf.Services.CTSUserDefinedMeta(t.Services),
 			BufferPeriod:    getTemplateBufferPeriod(conf, t),
 			Condition:       t.Condition,
+		})
+		if err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
findkim: 🕵️‍♀️ great idea, will add task name context to the error

driver/task.go
findkim: sure thing, thanks for always looking out for maintainability of code

driver/terraform.go
findkim: done! thanks again

driver/task.go
```diff
@@ -44,54 +47,295 @@ type BufferPeriod struct {
...
 }
 
+// Source returns the module source for the task
+func (t *Task) Source() string {
+	t.mu.RLock()
+	defer t.mu.RUnlock()
+	return t.source
+}
+
+// UserDefinedMeta returns metadata grouped by service
```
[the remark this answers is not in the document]
findkim: This was honestly me not being thorough and consistent. Sorry to add extra work for you to comment on it. I added copy logic for both `Providers()` and `UserDefinedMeta()` 🙇‍♂️

controller/readonly.go
lornasong: Gotcha! Sorry for the confusion!

driver/task.go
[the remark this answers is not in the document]
lornasong: No worries :) thanks for sharing 💞

lornasong (approved):
thanks for the discussion! this looks good to me!

PR #290, opened 2021-05-25
Update driver interface

findkim:
The force parameter for `driver.InitTask` is only used for testing and was prematurely added. Updated the signature to remove `force` and be consistent to accept `context`.


findkim:
I have this PR temporarily pointing to the branch for https://github.com/hashicorp/consul-terraform-sync/pull/289. This would be slated to merge to master after it.

lornasong (approved):
changes make sense to me!

PR #291, opened 2021-05-27
Fix catalog-service race condition during once-mode

lornasong:
Once-mode requires all dependencies to be retrieved before sending a
notification to CTS to render the complete template and execute the task.

A task has a template which can have multiple types of dependencies. A task configured with
catalog-service condition will only notify on a catalog-service type dependency changes and
not on other types of dependency changes.

During once-mode, the template dependencies are received asynchronously and therefore
can cause a race:
 - If catalog-service dependency is received last: this is fine with once-mode i.e. all
 dependencies are received, notification is sent, task is executed
 - If a different dependency is received last: no notification is sent because only catalog-service
 dependencies send notification. However, once-mode is waiting for all the dependencies to
 be received. Therefore, once-mode hangs even though all dependencies are received and task
 is ready to execute for once-mode.

Change:
 - Remove existing code to handle once-mode. This only partially handles the once-mode complexity.
 It only resolves the case when catalog-service dependency has no registration change during once-mode
 - Add new code to use a dependency counter to determine when all dependencies have been received.
 Once received, trigger notification regardless of what type of dependency is received last
 - Alternative ideas welcome!

Sample logs showing the hang:

```
2021/05/26 19:41:05.979458 [INFO] (ctrl) driver initialized
2021/05/26 19:41:05.979477 [INFO] (ctrl) executing all tasks once through
2021/05/26 19:41:05.979917 [DEBUG] (ctrl) rendered template false
2021/05/26 19:41:05.979926 [DEBUG] (ctrl) watching 3 dependencies                                 // notice: waiting for 3 dependency changes
2021/05/26 19:41:06.082522 [DEBUG] (notifier) dependency change type []*dep.HealthService
2021/05/26 19:41:06.086875 [DEBUG] (notifier) dependency change type []*dep.CatalogSnippet
2021/05/26 19:41:06.086891 [DEBUG] (notifier) notify registration change                          // catalog-service dep (log above) sends notification as expected
2021/05/26 19:41:06.087272 [DEBUG] (ctrl) rendered template false                                 // because of once-mode, template is not rendered because still missing 3rd dependency
2021/05/26 19:41:06.088918 [DEBUG] (notifier) dependency change type []*dep.HealthService         // last dependency received. but is not catalog-service type. therefore no notification sent.
2021/05/26 19:41:20.769355 [INFO] (cli) signal received to initiate graceful shutdown: interrupt  // note the timstamp. cts hangs for 14s
2021/05/26 19:41:30.769673 [INFO] (cli) graceful shutdown timed out, exiting
```

### DIFF
--- a/driver/terraform.go
+++ b/driver/terraform.go
@@ -482,7 +482,8 @@ func (tf *Terraform) initTaskTemplate() error {
 	})
 	switch tf.task.Condition().(type) {
 	case *config.CatalogServicesConditionConfig:
-		tf.template = &notifier.CatalogServicesRegistration{Template: tmpl}
+		tf.template = notifier.NewCatalogServicesRegistration(tmpl,
+			len(tf.task.Services()))
 	default:
 		tf.template = tmpl
 	}
--- a/templates/tftmpl/notifier/notifier.go
+++ b/templates/tftmpl/notifier/notifier.go
@@ -1,6 +1,8 @@
 package notifier
 
 import (
+	"log"
+
 	"github.com/hashicorp/consul-terraform-sync/templates"
 	"github.com/hashicorp/hcat/dep"
 )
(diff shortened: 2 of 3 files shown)


findkim (approved):
This took me a bit to understand where the race condition lies. Thanks for syncing with me on this! The fix makes sense to track the deps that go through `Notify()` now. If I am understanding your approach, it ensures that the catalog notification only occurs after the muted services ones go through, preventing the edge case that the muted dependencies come after causing the hang.

lornasong:
Thanks for the review and discussion @findkim! It really helped me identify where I was missing important context and how to explain the race condition for someone who isn't in my head. Updated the PR description with more details for posterity!

lornasong:
Sorry - to clarify on the approach, I would describe it as:
 - Send a special notification the first time (once-mode) that all dependencies have been received (regardless of the type of the last dependency that 'completes'). This is what this PR primarily adds
 - Send a notification when there is a catalog-service registration dependency change. And suppress notifications for other types of dependencies and catalog-service tag dependency changes. This is mostly existing code

I'll update the code comments to try to better document this. Let me know if you have any questions!

findkim:
> Send a special notification the first time ... that all dependencies have been received (regardless of the type of the last dependency that 'completes'

Gotcha thanks for the clarification

PR #294, opened 2021-06-01
Update e2e test to poll for Consul service registration and CTS events

lornasong:
In our e2e tests, there are a lot of `time.Sleep()`s used to wait for a Consul service to register and for CTS to trigger and finish in response to a change. These sleeps can sometimes cause tests to flake and are an arbitrarily long value that cause our tests to sometimes be longer than needed.

Currently, this has a minor improvement on the e2e CI job time. It reduces from ~1min 30-50s to ~1min 10s. However, this change's original intention was for an upcoming e2e test that causes the e2e test suite to timeout. This change reduces this upcoming test's length and prevents it from causing the e2e suite timeout.

Commits:
 - Add new WaitForEvent which polls CTS for triggering
 - Update RegisterConsulService to poll for the service to be registered
 - Update e2e test sleeps with latest WaitForEvent and RegisterConsulService.
 - Update non-e2e tests with latest RegisterConsulService
 - Update existing WaitForApi() polling method with a similar default timeout variable
 - Fix a small extra call to MakeTempDir that I noticed

### DIFF
--- a/api/test.go
+++ b/api/test.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"testing"
+	"time"
 
 	"github.com/hashicorp/consul-terraform-sync/testutils"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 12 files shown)


eikenb (approved):
So much goodness in such a small PR.
Replacing the Sleep calls with polling is brilliant. Makes it so the tests can complete as fast as possible while keeping timeouts. Awesome. :100: 
Other great things like removing cut-n-paste defaults with real constants, constant blocks and some general cleanup.
Thanks!

lornasong:
Thanks for the review @eikenb! Replacing sleep calls with polling is a pattern I stole from @ findkim when she added WaitForAPI() :).

PR #295, opened 2021-06-01
Add option to provide filter expressions on services

mkam:
New service config value allows the user to pass a [Consul filter expression](https://www.consul.io/api-docs/features/filtering) to filter nodes. Because tags can be used in this filter expression, the existing `tag` attribute will be deprecated in favor of the more general `filter`. Additionally, this will allow for filtering by multiple tags.

**Example Usage:**
Replacing existing `tag`:
Before:
```hcl
service {
  name = "web",
  tag = "example"
}
```
After:
``` hcl
service {
  name = "web",
  filter = "\"example\" in Service.Tags"
}
```
Filtering on multiple tags:
``` hcl
service {
  name = "web",
  filter = "\"example\" in Service.Tags and \"test\" in Service.Tags"
}
```
Filtering on service metadata:
``` hcl
service {
  name = "web",
  filter = "Service.Meta.example contains \"test\""
}
```
``` hcl
service {
  name = "web",
  filter = "Service.Meta[\"example\"] contains \"test\""
}
```
See the [filtering options for the health service](https://www.consul.io/api-docs/health#filtering-2) for an extensive list of filter selectors.


config/service.go
```diff
@@ -144,6 +154,16 @@ func (c *ServiceConfig) Finalize() {
 
 	if c.Tag == nil {
 		c.Tag = String("")
+	} else {
+		log.Println("[WARN] (config) The 'tag' attribute was marked for " +
+			"deprecation in v0.2.0 and will be removed in a future version " +
```
mkam, line 159: I can update this to a specific version if we know when we would want to remove the deprecated field.

lornasong: Great warning log for the deprecated tag field!

lornasong (approved):
looks good to me! thanks for chatting offline about how parsing stripped the `/` from the configuration and cts needing it add it back. example usage in the pr description was really helpful!

eikenb:
Does HCL support multiple quote styles that would remote all those escaped quotes, like how you can use single ticks (python) or back ticks (go) to define string values with double quotes in it? 
If so I think it would help the readability of those examples.

mkam:
> Does HCL support multiple quote styles that would remove all those escaped quotes, like how you can use single ticks (python) or back ticks (go) to define string values with double quotes in it?
> If so I think it would help the readability of those examples.


Based on [this documentation](https://www.terraform.io/docs/language/expressions/strings.html), it doesn't seem so.

eikenb:
Maybe using the indented heredoc syntax?

https://www.terraform.io/docs/language/expressions/strings.html#indented-heredocs

eikenb:
Hmm... they also mention single-tick quotes in the HCL 2 docs.

> Double and single quotes (" and ') are used to mark literal character sequences, which may be either punctuation markers or keywords.

https://github.com/hashicorp/hcl/blob/v2.10.0/hclsyntax/spec.md#syntax-notation

Seems like something should work. The docs are very vague though as they only mention the single tick at the top, but skip mentioning it in the expressions section where it covers quoting, heredocs and escaping.

Note that this obviously **isn't a blocker** as the examples we're discussing are in the commit message, not the PR itself. Just something that, if possible, I think we should use in our docs as it would make them easier to read.

mkam:
Tested some different cases and found that we don't currently support the single quotes since we're using HCL V1 for parsing. I also found that with the heredoc style strings, there's a newline at the end, which caused issues with the templating. I added a trim for newlines around the filter to account for this.

Examples with [heredoc](https://www.terraform.io/docs/language/expressions/strings.html#heredoc-strings):
```hcl
service {
  name = "web",
  filter = <<EOT
"example" in Service.Tags
EOT
}
```
```hcl
service {
  name = "web",
  filter = <<-EOT
  "example" in Service.Tags
  EOT
}
```

PR #296, opened 2021-06-02
Additional e2e tests for catalog-services condition

lornasong:
Add test case for `source_includes_var=false` to related tests. Add test for CTS configured with multiple tasks.

Commits:
 - Rename existing 'e2e_basic_task' module to 'local_instances_file' to align with other module naming patterns
 - Refactor some existing catalog-service condition e2e test with case for `source_includes_var=false` in addition to `source_includes_var=true`
 - Add new test for multiple tasks
 - Disable buffer_period for all e2e tests. The default buffer_period led to a wait period that caused timeouts in tests that triggered CTS multiple times.


e2e/e2e_test.go
```diff
@@ -61,17 +61,19 @@ func TestE2EBasic(t *testing.T) {
 
 	runSyncStop(t, configPath, 20*time.Second)
 
-	resourcesPath := fmt.Sprintf("%s/%s", tempDir, resourcesDir)
-	files := testutils.CheckDir(t, true, resourcesPath)
-	require.Equal(t, 3, len(files))
```
mkam, line 66: Could we check for the required number of files in each task's resource folder since this line has been removed?

lornasong: Good point! I added back the check for files in each of the new resource paths. Thanks!

lornasong:
@mkam thanks for reviewing! I added back the check for number of files in TestE2EBasic as you suggested. I also made a couple typo fixes and improvements that we chatted about offline. Ready for another look when you have a chance!

mkam (approved):
Looks good to me! Really liked the refactoring done to make the reusable test methods.

Issue #212, opened 2021-02-23
Issue on Terraform installation with `path=""`


danielehc:
### Describe the bug
Running Consul-Terraform-Sync with the following configuration for the driver block

```
# Driver "terraform" block
driver "terraform" {
 # version = "0.14.0"
  path = ""
  log = false
  persist_log = false
  working_dir = ""
}
```

causes the following output:

```
2021/02/18 16:02:34.980659 [INFO] (driver.terraform) installing terraform to path ''
2021/02/18 16:02:37.424658 [DEBUG] (driver.terraform) successfully installed terraform 0.14.7: /tmp/tfexec272869085/terraform
2021/02/18 16:02:37.424716 [INFO] (driver.terraform) successfully installed terraform
...
[ERR] (cli) error running controller in Once mode: could not apply changes for task learn-cts-example: error tf-init for 'learn-cts-example': exec: "terraform": executable file not found in $PATH
```

ad the execution will stop with a failure.

### Versions
**Consul Terraform Sync**
```
$ consul-terraform-sync -version

 0.1.0-techpreview2 (d844653)
Compatible with Terraform >= 0.13.0, < 0.15
```

### Configuration File(s)

The relevant section of the configuration is `path = ""` in the `driver "terraform"` block

```hcl
# Driver "terraform" block
driver "terraform" {
 # version = "0.14.0"
  path = ""
  log = false
  persist_log = false
  working_dir = ""
}
```

### Expected Behavior

One of the following:

1. CTS defaults `path = ""` as `./` (same as not having the value defined in the config) and installs TF binary in the current directory
2. CTS uses a random path to install TF  (i.e. `/tmp/tfexec272869085/terraform`) but then remembers that choice and is able to find the TF binary

I'd say the option 1 is preferable to avoid possible multiple downloads of the TF binary.

### Actual Behavior

A mix of the two, the TF binary gets downloaded on a random path `/tmp/tfexec272869085/terraform` but then CTS looks for the binary in `./`.

PR #297, opened 2021-06-02
Set Terraform path to working directory if empty string

mkam:
Fixes the issue where if the path was an empty string, Terraform would be installed in
a temporary directory, but CTS expected the binary to be in the current working directory.
This aligns now with the behavior for when no path is provided.

Closes https://github.com/hashicorp/consul-terraform-sync/issues/212


eikenb (approved):
Nice simple change w/ a test. :+1:

PR #303, opened 2021-06-08
Set namespace to empty string instead of null

findkim:
The namespace default value for Consul OSS will be an empty string instead of null, and "default" for Consul Enterprise


eikenb (approved):
A pretty straight forward :+1:.

Nice improvement. I also prefer the semantics of default/empty values over pointers when possible.

PR #304, opened 2021-06-08
Refactor user-defined meta to be tracked on the service instead of task

findkim:
This omits the need for the `UserDefinedMeta()` getter and reduces an attribute from the `driver.Task` object


controller/controller.go
```diff
@@ -271,16 +271,17 @@ func newDriverTasks(conf *config.Config, providerConfigs driver.TerraformProvide
 // configuration by ID to a driver service type. If a service is not
 // explicitly configured, it assumes the service is a logical service name
 // in the default namespace.
-func getService(services *config.ServiceConfigs, id string) driver.Service {
+func getService(services *config.ServiceConfigs, id string, meta map[string]map[string]string) driver.Service {
```
eikenb, line 274: Those map of maps data types can really get verbose. I tend to want to make them into local types to help with readability. Definitely not a big deal though.

eikenb (approved):
Nice simplification.

driver/task.go
```diff
@@ -226,21 +224,6 @@ func (t *Task) Source() string {
...
-	t.mu.RLock()
-	defer t.mu.RUnlock()
-
-	meta := make(map[string]map[string]string)
-	for k, nestedMap := range t.userDefinedMeta {
-		meta[k] = make(map[string]string)
-		for nestedK, nestedV := range nestedMap {
-			meta[k][nestedK] = nestedV
-		}
-	}
```
lornasong, line 240: Now that user-defined-meta is a field on services, does this type of 'copying' logic have to be added to the Services() getter, or maybe it's ok as is?

lornasong (approved):
looks good to me as well! left a non-blocking question

driver/task.go
findkim: Believe so since maps are by reference, good catch.

findkim: Love me some go playground https://play.golang.org/p/L-KljbPqTqZ

controller/controller.go
findkim: :nod: i do not like either. Thanks for the nudge, I looked around and saw this nested map in a number of places. Time to typify this! https://github.com/hashicorp/consul-terraform-sync/pull/304/commits/1e686da0eb8c13fa8a9b67d03839d16a0db6f3de

PR #305, opened 2021-06-08
Template newlines and expand tftmpl.Condition interface

findkim:
The newline changes don't effect the end result, but changes the extra newlines to be added by the caller instead of included in the template bits inconsistently.

The second commit around Condition interface is more of an opinionated change to group all condition related stuff to the Condition types instead of each file template managing the different condition template/generated code independently. Lmk if you prefer for me to leave out the second commit.

### DIFF
--- a/testutils/consul.go
+++ b/testutils/consul.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
-	"log"
 	"net/http"
 	"path/filepath"
 	"testing"
@@ -28,8 +27,6 @@ type TestConsulServerConfig struct {
 
 // NewTestConsulServer starts a test Consul server as configured
 func NewTestConsulServer(tb testing.TB, config TestConsulServerConfig) *testutil.TestServer {
-	log.SetOutput(ioutil.Discard)
-
 	var certFile string
 	var keyFile string
 	if config.HTTPSRelPath != "" {
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ test-integration:
 # test-e2e runs e2e tests
 test-e2e: dev
 	@echo "==> Testing ${NAME} (e2e)"
-	@go test ./e2e -count=1 -timeout=100s -tags=e2e -v ./... ${TESTARGS}
+	@go test ./e2e -count=1 -timeout=100s -tags=e2e ./... ${TESTARGS}
 .PHONY: test-e2e
 
 # test-setup-e2e sets up the sync binary and permissions to run in circle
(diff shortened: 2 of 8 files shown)


testutils/consul.go
```diff
@@ -28,8 +27,6 @@ type TestConsulServerConfig struct {
 
 // NewTestConsulServer starts a test Consul server as configured
 func NewTestConsulServer(tb testing.TB, config TestConsulServerConfig) *testutil.TestServer {
-	log.SetOutput(ioutil.Discard)
```
findkim, line 31: The caller can set this instead of forcing all tests utilizing the helper to be muted.

findkim: I changed it because it took me forever trying to debug something by running test and the verbose flag, and didn't realize it was not logging because I called test Consul server.

Just double checked the e2e CI job though, and definitely see a wall of logs 😂, so if this change is ok with everyone, I'll also look to remove `-v` from the job https://github.com/hashicorp/consul-terraform-sync/blob/master/Makefile#L53

eikenb: I found the common case was that I didn't want the logs and it was easier to keep them off and turn them back on when needed. The opposite could be true for CTS or other people just have different workflows that don't trigger that particular annoyance.

I do think some easy way to flip logging output on/off in testing would be the best long term solution. Maybe a testing CLI flag or a io.Writer var in testutils that is passed to SetOutput for tests... I'm not sure of the best answer but this does seem like something that would be very handy.

eikenb (approved):
I'm not as familiar with some of the things changed here, so I'm not 100% on semantics. But in general this looks good.

testutils/consul.go
lornasong: I'm good with going with this change. It sounds like this was impacting seeing CTS and/or Consul logs for integration tests maybe? As far as e2e tests, I tested out this change locally with a failing test and all looked readable - I didn't see any Consul logs likely since they're still set to be discarded (~12 lines down) and CTS logs were handled as expected.

lornasong (approved):
I like the refactor! I like that it keeps all the logic contained within the conditions and removes the need for each caller to check for condition type.

One thought that you may have already considered that I think could go well with the philosophy of your changes - not necessarily a change for this PR - is it looks like logic around sourceIncludesVariable() could even be handled within the appendVariable() and appendModuleAttribute() rather than by the caller of those methods. But I may be unaware of some design intention and future use of sourceIncludesVariable()

lornasong:
Thanks for chatting about this async @findkim! It sounds like there's need for sourceIncludesVariable() down the line and that handling sourceIncludesVar within appendVariable() and appendModuleAttribute() would not be very user-friendly since they could be a no-op which is unintuitive with the 'append' naming.

findkim:
Thanks for the discussion @lornasong ! I changed the SourceVariable method to public since it's not directly tied to templating.

PR #306, opened 2021-06-09
Run Terraform validate after initializing tasks

mkam:
Executes Terraform validate after task is initialized but before apply is run so that CTS fails faster when there are configuration issues.

- Returns solution suggestion when erroring with known CTS-specific issues (e.g., missing service variable)
- Returns diagnostics for all other issues formatted similarly to the human readable output of `terraform validate`
- If only warnings and not errors present, logs the warnings and continue

Example outputs:
```
2021/06/09 19:31:20.370041 [ERR] (cli) error initializing controller: 
module "example-task" is missing the "services" variable

module "example-task" is missing the "catalog_service" variable, add to module or set source_includes_var to false
```

```
2021/06/09 15:39:05.896247 [ERR] (cli) error initializing controller: 
warning: Version constraints inside provider configuration blocks are deprecated

on main.tf line 26, in provider "local"
26:  version = var.local.version

Terraform 0.13 and earlier allowed provider version constraints inside the provider configuration block, but that is now deprecated and will be removed in a future version of Terraform. To silence this warning, move the provider version constraint into the required_providers block.


error: Variables not allowed

on main.tf line 26, in provider "local"
26:  version = var.local.version

Variables may not be used here.
```


driver/driver.go
```diff
@@ -29,6 +29,9 @@ type Driver interface {
 	// UpdateTask supports updating certain fields of a task
 	UpdateTask(ctx context.Context, task PatchTask) (InspectPlan, error)
 
+	// ValidateTask validates the configurations of the task
+	ValidateTask(ctx context.Context) error
+
```
lornasong, line 34: I was wondering what your thoughts were if validation was instead happened within InitTask() instead of in its own method? No opinion, just want to understand better! Also curious @findkim of your thoughts since you've been doing some refactoring of the driver interface lately.

client/terraform_cli.go
```diff
@@ -194,6 +195,45 @@ func (t *TerraformCLI) Plan(ctx context.Context) (bool, error) {
...
+		return err
+	}
+
+	var sb strings.Builder
+	for _, d := range output.Diagnostics {
+		sb.WriteByte('\n')
+		switch d.Detail {
+		case `An argument named "services" is not expected here.`:
+			fmt.Fprintf(&sb, "%s is missing the \"services\" variable\n", *d.Snippet.Context)
+		case `An argument named "catalog_services" is not expected here.`:
```
findkim, line 211: I don't think the error verbiage around this has changed much at all across Terraform versions, so this would be fine. Just wanted to it call-out in case we stop seeing this useful new error, we could suspect that the error string has changed.

lornasong (commented):
this looks really good! terraform-validate is threaded through nicely and the tests are ✨. thanks for including samples of the error message in the pr description. i left a couple questions that i was curious about

controller/controller.go
```diff
@@ -117,6 +117,11 @@ func (ctrl *baseController) init(ctx context.Context) (*driver.Drivers, error) {
 			return nil, err
 		}
 
+		err = d.ValidateTask(ctx)
```
findkim, line 120: Looks like a good location to validate! Right after all of the files are generated 👍

client/terraform_cli_test.go
```diff
@@ -266,6 +268,99 @@ func TestTerraformCLIPlan(t *testing.T) {
 	}
 }
 
+func TestTerraformCLIValidate(t *testing.T) {
+	t.Parallel()
+
+	errJson := `{
+		"valid": false,
+		"diagnostics": [
```
findkim, line 276: Strings in backticks preserve the exact string and its spacing, so here it would have the raw JSON include the go-linter tabs that align it wonky. For example:

```json
{
        "valid": false,
        "diagnostics": [ ]
    }
```

To get the pretty formatted json
```json
{
  "valid": false,
  "diagnostics": [ ]
}
```

The go backtick'ed string would need to resemble
```go
	errJson := `{
  "valid": false,
  "diagnostics": [ ]
}`
```

Since this is a test and the spacing within the raw string itself doesn't matter to the JSON unmarshaller, this is a nit and you don't have to change! it's pretty readable as-is.

findkim (approved):
The PR summary is 💯 and clean code! Thanks for including examples of that CTS would output, it made reviewing the changes very clear. I'm excited to have more informative messages for users. The current errors you'd see is confusing and misleading in the CTS use-case for Terraform

driver/terraform.go
```diff
@@ -491,6 +504,22 @@ func (tf *Terraform) initTaskTemplate() error {
 	return nil
 }
 
+func (tf *Terraform) validateTask(ctx context.Context) error {
+	taskName := tf.task.Name()
+
+	if err := tf.init(ctx); err != nil {
```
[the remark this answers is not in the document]
mkam: I was thinking about that too! I ended up leaving them in because the init method first checks if the workspace is already initialized, and I thought it would be better to not have applyTask() and inspectTask() depend on validateTask() running first. Maybe another point for moving the validate to be part of InitTask, though!

client/terraform_cli.go
mkam: I could add an integration or e2e tests for these two cases specifically so that when we test new versions of Terraform, we'd be able to tell if an error message changed.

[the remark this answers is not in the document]
mkam: Yeah, I would worry that we could potentially return a wrong message if it's too broad of a check. @findkim had the same concern so I've already mentioned this in response to her comment, but I think I can mitigate this by adding an integration or e2e tests for these two errors.

findkim: Good idea about an integration/e2e test! Doesn't have to block this PR and a follow-up would work too.

Another thought came to mind, do you know how this edge-case handling would catch if the `var.services` type declared in the module is not compatible with the type CTS expects? not sure how to clearly say that, so here's an example:

[CTS expects this type](https://www.consul.io/docs/nia/terraform-modules#services-variable)
```hcl
variable "services" {
  type = map(
    object({
      id          = string
      name      = string
      ...
    })
  )
}
```

Terraform module used has a different type for declared in its `variables.tf` file
```hcl
variable "services" {
  type = list(string)
}
```

mkam: Here's the current output:

```
2021/06/10 16:23:13.736550 [ERR] (cli) error initializing controller: 
error: Invalid value for module argument

on main.tf line 31, in module "example-task"
31:  services         = var.services

The given value is not suitable for child module variable "services" defined at .terraform/modules/example-task/variables.tf:37,1-20: list of string required.
```

I could also check for this error with services and catalog_services.

driver/terraform.go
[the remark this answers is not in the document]
lornasong: Mmmm good point about not having a applyTask() and inspectTask() not depending on validateTask() running first. Agree that if validate gets moved in to InitTask, then it would be worth considering removing the terraform-init from apply and inspect!

client/terraform_cli.go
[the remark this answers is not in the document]
lornasong: :nod: I like the idea of having a test for the two errors. Will second what Kim mentioned separately that it's non-blocking and can be done after this PR.

lornasong (approved):
Thanks for chatting about my questions! Looks like any potential adjustments to the driver interface can be done after this PR. These changes look great!

client/terraform_cli.go
findkim: Thanks for checking, the error seems informative enough for me!

mkam:
Thanks for the reviews! I'll be adding the tests for error messages in a follow up PR.

PR #310, opened 2021-06-11
Test for CatalogServicesRegistrationQuery query parameter filtering in Fetch()

lornasong:
Changes:
 - Add testing for various query parameters for Fetch()
 - Catch a regexp nil-pointer edge-case from testing (regexp can't be nil when running
 CTS e2e)
 - Remove e2e test that was only able to test node-meta query parameter
 filtering and was over-scoped (also unnecessarily tested triggering)

Context: addresses's @ findkim comment that testing node-meta as an e2e test was
overscoped and that the main test case is around query filtering (thanks!) https://github.com/hashicorp/consul-terraform-sync/pull/278#discussion_r631158825

### DIFF
--- a/e2e/condition_test.go
+++ b/e2e/condition_test.go
@@ -5,7 +5,6 @@ package e2e
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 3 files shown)


findkim (approved):
Nice job sliming the test around Fetch

PR #311, opened 2021-06-14
E2E test for parsing Terraform validate

mkam:
Also fixes the error message which did not work for TF v0.13 and v0.14 because they did not return the context. I also realized the context was showing the task name and not the module name anyway, so this new message is more informative either way.

- Add e2e test so that if the error messages changes in newer versions of terraform, we will know to adjust accordingly
- Updated error message to determine task name from current workspace and improve some formatting
- Assert that specific error message is returned in unit test
- Include entire validate output in unit test (previously was just the info we cared about)
- Add test cases to unit test for both v0.13/v0.14 and v0.15 outputs

### DIFF
--- a/client/terraform_cli.go
+++ b/client/terraform_cli.go
@@ -207,20 +207,21 @@ func (t *TerraformCLI) Validate(ctx context.Context) error {
 		sb.WriteByte('\n')
 		switch d.Detail {
 		case `An argument named "services" is not expected here.`:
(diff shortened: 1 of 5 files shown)


findkim (approved):
Good call to add test cases for v0.13 and v0.14. I like the task name context that's now included too

lornasong:
Nice solution for checking the error across versions!

Issue #314, opened 2021-06-15
Move working_dir option from driver to task-level


findkim:
#### Description
The working directory is more associated with each task and its Terraform workspace than the driver. Suggest to move the `driver.working_dir` option to the task configuration block for more granular control.

- Absolute path: `"/custom/path/for/my-task"`
- Local path: `"./subdir/my-task"`
- (default) Child path from a new global `working_dir` option: `"my-task"` which results in `./sync-tasks/my-task`

```hcl
task {
  name = "my-task"
  working_dir = "/custom/path/for/my-task"
}
```

Meanwhile, still offering the top-level `working_dir` option that is the parent directory for all tasks, currently defaults to `./sync-tasks`

```hcl
working_dir = "./sync-tasks"
```

#### Note
The plan is for the current [`driver.working_dir`](https://www.consul.io/docs/nia/configuration#working_dir) to be marked as deprecated in v0.3.0, and then removed in v0.5.0
```hcl
driver "terraform" {
  working_dir = "./sync-tasks"
}
```

findkim:
Reminder to self: make config doc updates for v0.3.0

PR #315, opened 2021-06-15
Deprecate driver.working_dir in preference for global working_dir

findkim:
The working directory is more associated with each task and its Terraform workspace than the driver.

Related #314


lornasong (approved):
makes sense to me! nice deprecation warn message!

findkim:
Thanks! I referenced @mkam 's deprecation message as prior art ✨

PR #316, opened 2021-06-16
Move validate call to be part of task initialization

mkam:
We determined that not all drivers may have a validate method, so removed validate
task from the interface and made it part of the Terraform driver InitTask method.
Also moves workspace initialization to InitTask and remove it from apply/inspect/validate.

New output:
```
2021/06/16 17:14:43.618472 [ERR] (ctrl) error initializing task "test-task"
2021/06/16 17:14:43.618496 [ERR] (cli) error initializing controller: 
module for task "test-task" is missing the "services" variable

module for task "test-task" is missing the "catalog_services" variable, add to module or set "source_includes_var" to false
```

### DIFF
--- a/driver/terraform.go
+++ b/driver/terraform.go
@@ -361,8 +361,8 @@ func (tf *Terraform) initTask(ctx context.Context) error {
 	// initialize workspace
 	taskName := tf.task.Name()
 	if err := tf.init(ctx); err != nil {
-		log.Printf("[ERR] (driver.terraform) error initializing workspace, "+
-			"skipping validate for '%s'", taskName)
+		log.Printf("[ERR] (driver.terraform) error initializing workspace "+
+			"for task '%s'", taskName)
 		return err
 	}
 
--- a/driver/terraform.go
+++ b/driver/terraform.go
@@ -148,7 +148,7 @@ func (tf *Terraform) Task() *Task {
 
 // InitTask initializes the task by creating the Terraform root module and related
 // files to execute on.
-func (tf *Terraform) InitTask(context.Context) error {
+func (tf *Terraform) InitTask(ctx context.Context) error {
 	tf.mu.Lock()
 	defer tf.mu.Unlock()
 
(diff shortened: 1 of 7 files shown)


driver/terraform.go
```diff
@@ -350,7 +337,7 @@ func (tf *Terraform) init(ctx context.Context) error {
 }
 
 // initTask initializes the task
-func (tf *Terraform) initTask() error {
+func (tf *Terraform) initTask(ctx context.Context) error {
```
findkim, line 340: It makes sense to group the workspace init and validate here. The removal of redundancy looks good! But I can't recall if there was prior intention for this -- gonna loop in @lornasong

driver/terraform.go
```diff
@@ -148,7 +148,7 @@ func (tf *Terraform) Task() *Task {
 
 // InitTask initializes the task by creating the Terraform root module and related
 // files to execute on.
-func (tf *Terraform) InitTask(context.Context) error {
+func (tf *Terraform) InitTask(ctx context.Context) error {
```
findkim, line 151: yay good to see that ctx did end up being useful for this signature.

findkim (approved):

driver/terraform.go
```diff
@@ -371,6 +358,19 @@ func (tf *Terraform) initTask() error {
 	// will reinit
 	tf.inited = false
 
+	// initialize workspace
+	taskName := tf.task.Name()
+	if err := tf.init(ctx); err != nil {
+		log.Printf("[ERR] (driver.terraform) error initializing workspace, "+
+			"skipping validate for '%s'", taskName)
```
lornasong, line 365: Minor nit: the "skipping..." wording comes from the fact that (for example) the terraform-apply didn't happen but will be retried on the next change. Since I think we're exiting if this init fails, the "skipping" language may not be necessary anymore.

driver/terraform.go
lornasong: I can't recall any context. I also think adding init and validate here makes sense!

lornasong (approved):
also looks good to me! left a non-blocking suggestion

driver/terraform.go
mkam: Makes sense, I'll remove it!

Issue #318, opened 2021-06-18
Enabling a task executes the task but does not store an event


lornasong:
### Describe the bug
When a disabled task is enabled, the task is executed. Since the task is executed, a task should be stored indicating the time of the task execution, success/failure, and other information. However, no event is stored. Because no event is stored for the latest task execution, the task's status can be potentially out-of-date.

It looks like this bug is not newly introduced. Event was actually not stored when first adding the enable task CLI.

### Versions
**Consul Terraform Sync**
```
consul-terraform-sync v0.1.0-175-gc5016b5 (c5016b5)
Compatible with Terraform >= 0.13.0, < 0.16
```

### Configuration File(s)
<details>
<summary>Click to toggle contents of config file </summary>

```hcl
log_level = "INFO"

consul {
  address = "localhost:8500"
}

driver "terraform" {
  log = true
  working_dir = "build"
}

task {
  name = "task_b"
  services = ["api"]
  version = "0.0.1"
  source = "lornasong/cts_file/local"
}
```
</details>

### Expected Behavior
When a task is enabled and executes, a new event is stored for that execution

### Actual Behavior
When a task is enabled and executes, no event is stored for that execution

### Steps to Reproduce
1. Run Consul: `consul agent -dev`
2. Run CTS with config above: `consul-terraform-sync -config-file=config.hcl`
3. Disable task: `consul-terraform-sync task disable task_b`
4. Check task's events: `curl localhost:8558/v1/status/tasks/task_b?include=events | jq`
Sample response
```
{
  "task_b": {
    "task_name": "task_b",
    "status": "successful",
    "enabled": false,
    "providers": [],
    "services": [
      "api"
    ],
    "events_url": "/v1/status/tasks/task_b?include=events",
    "events": [
      {
        "id": "e96f1c65-fd03-97bc-2d94-6205f49b84db",
        "success": true,
        "start_time": "2021-06-18T10:15:17.73669-04:00",
        "end_time": "2021-06-18T10:15:18.324435-04:00",
        "task_name": "task_b",
        "error": null,
        "config": {
          "providers": [],
          "services": [
            "api"
          ],
          "source": "lornasong/cts_file/local"
        }
      }
    ]
  }
}
```
5. Observe in response that there is one event
6. Register a service: `consul services register -id=api-1 -name=api`
7. Enable task: `consul-terraform-sync task enable task_b` (walk through flow to confirm enabling task)
8. See task execute in CTS logs
[...]

PR #319, opened 2021-06-18
Store an event when a task becomes enabled and executes

lornasong:
When a task is disabled and then enabled via CLI, the task may execute if there
are changes. Whenever a task executes, an event should be stored.

When enable CLI was implemented, storing an event on task execution was
unintentionally not implemented. This change fixes this and stores an event.

Changes:
 - Identifies when a task is executed: when ?run=now i.e. runOp == driver.RunOptionNow
 - Stores an event
 - Retrieves an error that may result from driver.UpdateTask to store in the event

Fixes: https://github.com/hashicorp/consul-terraform-sync/issues/318

### DIFF
--- a/api/task.go
+++ b/api/task.go
@@ -14,6 +14,7 @@ import (
 	"github.com/hashicorp/consul-terraform-sync/driver"
 	"github.com/hashicorp/consul-terraform-sync/event"
 	"github.com/mitchellh/mapstructure"
+	"github.com/pkg/errors"
 )
 
 const taskPath = "tasks"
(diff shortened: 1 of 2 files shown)


api/task.go
```diff
@@ -122,11 +122,33 @@ func (h *taskHandler) updateTask(w http.ResponseWriter, r *http.Request) {
...
+			Source:    task.Source(),
+		})
+		if err != nil {
+			log.Printf("[ERROR] (api.task) error creating event for task %q: %s",
+				taskName, err)
+		}
+		defer func() {
+			ev.End(storedErr)
+			log.Printf("[TRACE] (api.task) adding event %s", ev.GoString())
+			if err := h.store.Add(*ev); err != nil {
```
findkim, line 140: Minor: The `defer func` seems to be initialized even if there is an error, where `ev` may be nil, then it would panic here. It's quite unlikely `NewEvent` will ever error, but can you update this where panics can't be a part of a possible path? Thanks!

findkim (approved):
Good catch that the update API was missing the event tracking. Changes look good!

api/task.go
lornasong: Good point! Thanks for raising this! Some above code checks for the edge-case that would cause `NewEvent` to return error i.e. when taskName is empty. But I'm realizing that this dependency isn't clear at all and that code may evolve in a way that opens up for more panic. I decided to add a return when there's an error creating an event. Thank you!

Issue #320, opened 2021-06-18
Disabled tasks do not trigger after re-enabling


lornasong:
### Describe the bug
After a task that was disabled is re-enabled using the cli, the task should trigger when its condition is met. However, the task is not triggered even though the dependency change is identified by CTS.

### Versions
**Consul Terraform Sync**
```
consul-terraform-sync v0.1.0-175-gc5016b5 (c5016b5)
Compatible with Terraform >= 0.13.0, < 0.16
```

### Configuration File(s)
_Reminder to redact any sensitive information that may be present in this file_

<details>
<summary>Click to toggle contents of config file </summary>

```hcl
log_level = "TRACE"

consul {
  address = "localhost:8500"
}

driver "terraform" {
  log = true
  working_dir = "build/testing"
}

task {
  name = "task_c"
  source      = "./build/modules/catalog-services"
  condition "catalog-services" {
      regexp = "api"
      source_includes_var = true
  }
}
```
</details>

### Terraform Module
<details>
<summary>Click to toggle main.tf</summary>

```hcl
resource "local_file" "tags" {
    for_each = var.catalog_services
    content = join(", ", each.value)
    filename = "resources/tags_${each.key}.txt"
}
```
</details>

<details>
<summary>Click to toggle variables.tf</summary>

```hcl
variable "services" {
  description = "Consul services monitored by Consul Terraform Sync"
  type = map(
    object({
      id        = string
    })
  )
}

variable "catalog_services" {
  description = "Consul catalog service names and list of all known tags for a given service"
  type = map(list(string))
}
```
</details>

### Expected Behavior
When the task is re-enabled and a new service is registered, the task is triggered and creates the resource.

### Actual Behavior
When the task is re-enabled and a new service is registered, the task is not triggered.

Example logs:
```
// ... task was enabled
2021/06/18 21:17:06.432995 [DEBUG] (notifier.cs) received dependency change type []*dep.CatalogSnippet
2021/06/18 21:17:06.433031 [DEBUG] (notifier.cs) notify registration change
2021/06/18 21:17:06.433172 [TRACE] (driver.terraform) checking dependency changes for task task_c
2021/06/18 21:17:11.435207 [TRACE] (driver.terraform) checking dependency changes for task task_c
```
In the above logs, I registered a service that meets the condition of task_c. The notifier sent a notification for the registration. The driver checks for the change but does not render and execute the template.

### Steps to Reproduce
1. Run Consul: `consul agent -dev`
[...]

PR #321, opened 2021-06-19
Fix re-enabled task to trigger

lornasong:
Fixes: https://github.com/hashicorp/consul-terraform-sync/issues/320

Draft PR: 
 - Depends on related hcat proposal https://github.com/hashicorp/hcat/pull/57 to be okayed
 - Failing e2e test requires this hcat update
 - Depends on a bug fix in branch `enable-event`. Once that branch is merged into v0.2.0, I'll adjust this PR to merge into v0.2.0 as well.

Fixes an issue where tasks that were originally enabled, then disabled, then re-enabled :) would hang and not trigger even when condition satisfied.

Commits (more details in commit msg):
 - Fix the original issue which occurs when a task is re-enabled, it creates another hcat template with the same template content (hcat template content must be unique https://github.com/hashicorp/consul-terraform-sync/pull/167 - thanks @ findkim!). Fix this by only creating a new hcat template only if content is different.
 - Fix a new issue that the fix above introduced. New issue occurs when existing (vs. new) hcat templates are forced to re-render when there are no dependency changes (as can happen when a task is enabled), the re-rendering will hang. Fix this by proposing an [hcat change](https://github.com/hashicorp/hcat/pull/57) to return result.Complete == true (which will prevent hanging) when there are no dependency changes

### DIFF
--- a/driver/terraform.go
+++ b/driver/terraform.go
@@ -496,12 +496,18 @@ func (tf *Terraform) initTaskTemplate() error {
 		FuncMapMerge: tmplfunc.HCLMap(metaMap),
 	})
 
-	if tf.template != nil && tf.template.ID() == tmpl.ID() {
-		// if the new template ID is the same as an existing one (e.g. during a
-		// task update), then the template content is the same. Template
-		// content must be unique.
-		// See: https://github.com/hashicorp/consul-terraform-sync/pull/167
-		return nil
+	if tf.template != nil {
+		if tf.template.ID() == tmpl.ID() {
+			// if the new template ID is the same as an existing one (e.g.
+			// during a task update), then the template content is the same.
+			// Template content must be unique.
+			// See: https://github.com/hashicorp/consul-terraform-sync/pull/167
+			return nil
+		}
+
+		// cleanup old template from watcher
+		tf.watcher.Mark(tf.template)
+		tf.watcher.Sweep(tf.template)
 	}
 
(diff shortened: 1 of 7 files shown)


driver/terraform.go
```diff
@@ -487,6 +487,15 @@ func (tf *Terraform) initTaskTemplate() error {
 		Renderer:     renderer,
 		FuncMapMerge: tmplfunc.HCLMap(metaMap),
 	})
+
+	if tf.template != nil && tf.template.ID() == tmpl.ID() {
```
findkim, line 491: It may be the case where the existing template lingers indefinitely when the new template content differs for this task, wrt to the hcat.Watcher.

Is there a way to clear the older, un-used template from the Watcher? Is [`Watcher.Sweep`](https://pkg.go.dev/github.com/hashicorp/hcat#Watcher.Sweep) appropriate here when `tf.template != nil && tf.template.ID() != tmpl.ID()`

lornasong: Thanks for the feedback! Yes, it makes sense to add a cleanup! I went ahead and added it.

Noting also that we chatted that there are other issues with this PR to work through to resolve the original issue.

lornasong:
We deleted branch `v0.2.0` as part of the 0.2.0 release which auto-closed this PR. Leaving it closed as fixing #320 will take more development.

PR #322, opened 2021-06-22
Fix re-enabled task to trigger

### LINKED ISSUE #320: Disabled tasks do not trigger after re-enabling

findkim:
This PR is a superset of #321 from @lornasong, it has more details around the bug for #320

Fix UpdateTask to enable a disabled task that has never run

Fixes the failing tests that were waiting on the wrong
event, and prematurely continuing the test before the catalog
changes were realized as Terraform changes.

Closes #320

### DIFF
--- a/api/test.go
+++ b/api/test.go
@@ -103,7 +103,9 @@ func WaitForEvent(t *testing.T, client *Client, taskName string, start time.Time
 					continue
 				}
 				mostRecent := task.Events[0]
-				if mostRecent.EndTime.After(start) {
+				if mostRecent.StartTime.After(start) && mostRecent.EndTime.After(start) {
+					// start is the time before a trigger occurs, so this checks if the
+					// most recent event started and completed after the trigger
 					polling <- struct{}{}
 					return
 				}
--- a/driver/terraform.go
+++ b/driver/terraform.go
@@ -496,12 +496,18 @@ func (tf *Terraform) initTaskTemplate() error {
 		FuncMapMerge: tmplfunc.HCLMap(metaMap),
 	})
 
-	if tf.template != nil && tf.template.ID() == tmpl.ID() {
-		// if the new template ID is the same as an existing one (e.g. during a
-		// task update), then the template content is the same. Template
-		// content must be unique.
-		// See: https://github.com/hashicorp/consul-terraform-sync/pull/167
-		return nil
+	if tf.template != nil {
+		if tf.template.ID() == tmpl.ID() {
+			// if the new template ID is the same as an existing one (e.g.
+			// during a task update), then the template content is the same.
+			// Template content must be unique.
+			// See: https://github.com/hashicorp/consul-terraform-sync/pull/167
+			return nil
+		}
+
+		// cleanup old template from watcher
+		tf.watcher.Mark(tf.template)
+		tf.watcher.Sweep(tf.template)
 	}
 
 	switch tf.task.Condition().(type) {
--- a/driver/terraform.go
+++ b/driver/terraform.go
@@ -148,7 +148,7 @@ func (tf *Terraform) Task() *Task {
 
 // InitTask initializes the task by creating the Terraform root module and related
 // files to execute on.
-func (tf *Terraform) InitTask(context.Context) error {
+func (tf *Terraform) InitTask(ctx context.Context) error {
 	tf.mu.Lock()
 	defer tf.mu.Unlock()
 
@@ -158,7 +158,7 @@ func (tf *Terraform) InitTask(context.Context) error {
 		return nil
 	}
 
-	return tf.initTask()
+	return tf.initTask(ctx)
 }
 
 // SetBufferPeriod sets the buffer period for the task. Do not set this when
@@ -280,7 +280,7 @@ func (tf *Terraform) UpdateTask(ctx context.Context, patch PatchTask) (InspectPl
 	}
 
 	if reinit {
-		if err := tf.initTask(); err != nil {
+		if err := tf.initTask(ctx); err != nil {
 			return InspectPlan{}, fmt.Errorf("Error updating task '%s'. Unable to init "+
 				"task: %s", taskName, err)
 		}
@@ -318,19 +318,6 @@ func (tf *Terraform) UpdateTask(ctx context.Context, patch PatchTask) (InspectPl
 	return InspectPlan{}, nil
 }
 
-// ValidateTask validates the generated configurations
-func (tf *Terraform) ValidateTask(ctx context.Context) error {
-	tf.mu.Lock()
-	defer tf.mu.Unlock()
-
-	if !tf.task.IsEnabled() {
-		log.Printf("[TRACE] (driver.terraform) task '%s' disabled. skip"+
-			"validating", tf.task.Name())
-		return nil
-	}
-	return tf.validateTask(ctx)
-}
-
 // init initializes the Terraform workspace if needed
 func (tf *Terraform) init(ctx context.Context) error {
 	taskName := tf.task.Name()
@@ -350,7 +337,7 @@ func (tf *Terraform) init(ctx context.Context) error {
 }
 
 // initTask initializes the task
-func (tf *Terraform) initTask() error {
+func (tf *Terraform) initTask(ctx context.Context) error {
 	input := tftmpl.RootModuleInputData{
 		TerraformVersion: TerraformVersion,
 		Backend:          tf.backend,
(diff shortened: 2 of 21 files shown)


driver/terraform.go
```diff
@@ -60,7 +60,8 @@ type Terraform struct {
 	logClient  bool
 	postApply  handler.Handler
 
-	inited bool
+	inited       bool
+	renderedOnce bool
```
eikenb, line 64: I was thinking about hashicat's API in regard to the ResolveEvent and how this sort of internal state field really shouldn't be necessary.

Right not the `Complete` field only returns as true if *that run* resulted in a completed template. That is Complete is false even when the template has been completely rendered (all fields filled in), if there was no new values. 

With the addition of `NoChange`, we now have a bit of a redundancy. Previously, as long as you ignored the initial runs that don't have the values, every run after the initial Complete=true indicated both that there was a change and that the template is complete. Now NoChange and Complete duplicate this information.

Basically I'm thinking we change Complete to always return true unless it is missing values. That way once it had completed once it would stay complete (unless it had hierarchical dependencies, like looping over all services and looking up service info on each). Then NoChange could be used to know if an update had occurred.

Sorry about this being a bit much for a PR comment, I started with a short comment and it sort of grew.

findkim: No worries about it being long, the context is useful! And I agree with your sentiment that this is a hashicat detail that the caller doesn't need to manage. I wanted to work through a solution from the CTS side and it helped reveal some limitations fo the current hashicat API. While working on this, I was also tinkering on the hashicat side to see how a different API would work.

Yeah `Complete` ended up being a challenge to use only because now there are 2 threads attempting to render the same template. They end up racing each other -- UpdateTask and the background go routine waiting on the `watcher.Wait`.

+1 to your suggestions, adding a bit to it too to make this comment thread longer 😉 

1. Always return `Complete = true` unless it's missing values
2. `NoChange` to determine if an update had occurred
3. The [template's dirty state](https://github.com/hashicorp/hcat/blob/master/template.go#L139-L155) doesn't handle multiple calls to `resolver.Run` well. This works well with one caller, but CTS probably introduced a new use-case that caused a problem here.
    a. The first caller would receive that it's dirty and run the template, but the second concurrent call would not know that it's dirty and receive `NoChange`.
    b. I saw this happen with the main go routing waiting on `watcher.Wait` and the second caller ended up being `UpdateTask`. In this case, since UpdateTask is a CLI/API call, it is intended to be synchronous, but the end-user would see that there are no changes for the task to execute.
4. It might be worthwhile to cache the rendered template content and always return `Content` for the above two cases, `Complete = true` or `NoChange = true`. This may make it friendlier for multiple callers.

The `renderedOnce` tracking here was a weird to figure out due to the race from #3b. Since we want UpdateTask to execute synchronously, it exposed an edge case where hcat returns `NoChange` since the other caller received the `Complete` since only one could consume the `dirty` state. The `UpdateTask` thread would continue processing and attempt to run the task, but turns out the template file hasn't rendered yet (racing with the other go routine) and errors when using the expected file for Terraform.

We don't have merge this PR in preference for hcat changes, it was a useful exercise for me to work through to better understand the bug. Though I would like to merge the changes around the `WaitForEvent` test helper function.

api/test.go
```diff
@@ -103,7 +103,9 @@ func WaitForEvent(t *testing.T, client *Client, taskName string, start time.Time
 					continue
 				}
 				mostRecent := task.Events[0]
-				if mostRecent.EndTime.After(start) {
+				if mostRecent.StartTime.After(start) && mostRecent.EndTime.After(start) {
+					// start is the time before a trigger occurs, so this checks if the
+					// most recent event started and completed after the trigger
```
lornasong, line 108: Thanks for looking into this! Just want to make sure I understand the cases that this fixes. Is the 'waiting on the wrong event' typically the WaitForEvent after a cli? I can imagine that the cli returns only after the event is stored so the WaitForEvent isn't waiting for the right event. Or is it maybe a different scenario?

I was also curious what your thoughts on if it would be possible to use it in a pattern like:
```
api.WaitForEvent(t, cts, taskName, time.Now(), defaultWaitForEvent) // call Wait ahead
service := testutil.TestService{ID: "api-1", Name: "api"}
testutils.RegisterConsulService(t, srv, service, testutil.HealthPassing,
 	defaultWaitForRegistration)
```
Asking mostly because I've found the e2e tests visually busy and have been thinking about how to reduce lines and make the tests easier on the eyes to read

findkim: From the uses, it seems that `WaitForEvent` was a helper to wait for the event status to update when a Consul catalog change occurs that triggers a task and finish executing.

Yeah, the scenario for using `WaitForEvent` after a CLI update.

What I found happening was that `WaitForEvent` would return early because it's possible that the main go-routine would run a task at the same time when the `UpdateTask` would run the same task. Both could write an event for the task. When testing for `UpdateTask`'s event, it could catch the event created by the main go-routine because it may have started prior to trigger and complete after the `time.Now()` when `WaitForEvent` was called. Most noticeable for the test `e2e.TestTasksUpdate`.

I'd be all for reducing lines of code! Though I don't know how the specific code example would work out since `WaitForEvent` is a blocking call. I can play around to see what's up

lornasong: Thanks for the details and example of how the WaitForEvent wasn't working! Really helpful! That makes a lot sense.

🤦 completely forgot that WaitForEvent is blocking. No need to spend time playing around with it unless it's something you're very interested in. I'm good with the changes related to fixing WaitForEvent. I know there's an ongoing thread on the changes for the re-enable bug. So I'm planning to hold-off until we decide what changes y'all want to merge.

driver/terraform.go
eikenb: The only tricky bit with implementing the above is the Content caching. I don't think it would be good to just add it to tmpl.Execute(), so I'm giving it some time to percolate.

This could be merged as is if it will move things forward. Depends on whether it is better to have this working now so it isn't a blocker or wait until the hcat changes land to do it. I didn't mean for this to be a blocking review, more that it prompted the design thought and I wanted to communicate that.

findkim: :nod: no strong feelings about the cache, it was an idea that came to mind around consistency. I think CTS would function just fine without a cache for `NoChange` since the template is written to disk here, and so if the file exists that would be the same as the cache.

I see. Thanks for clarifying it's non-blocking. We're lucky to be in close proximity so once hashicat changes are in movement, we can easily update CTS as well. Thanks @eikenb for challenging the current API and looking to improve it 🙂

lornasong (approved):
🎉 🎉🎉 I like how clean and simple your solution is! It'll make it easy to update CTS with any hcat changes down the line as you and @ eikenb discussed. Also nice catch with the WaitForEvent issue!

findkim:
I see a new PR https://github.com/hashicorp/hcat/pull/58 that changes the `resolver.Run` semantics. I'll go ahead and merge this PR and pull in the changes in a follow-up PR in case it effects other areas.

PR #326, opened 2021-06-28
Temporarily disable benchmarks

lornasong:
Benchmarks are currently failing weekly. Disabling until we can prioritize
looking into why it is failing


mkam (approved):
Thanks for taking care of this!

Issue #327, opened 2021-06-28
Support remote operations for CTS on Terraform Enterprise 


devarshishah3:
#### Description
Allow Terraform Enterprise to support remote operations for CTS Tasks

#### Use Cases
Terraform Enterprise provides infrastructure operators the ability to support audit logging, secure state store, secure variable management, consistent environments to drive automation, approval workflows and notifications. These capabilities allow enterprises to streamline internal processes and demonstrate ongoing accountability to regulators for the changes made to the networking infrastructure using Terraform Enterprise. 

Supporting remote operations for CTS on Terraform Enterprise will allow operators to integrate their CTS driven workflows for automating infrastructure into their organization's existing operational process.

findkim:
released 😁 https://discuss.hashicorp.com/t/consul-terraform-sync-v0-3-0-oss-and-enterprise/29006

PR #331, opened 2021-07-06
Add error when setting terraform_version config option per task in OSS

findkim:
The Terraform Cloud API allows configuring the Terraform CLI version for each workspace. A new option is being added to CTS enterprise to independently configure the TF version per task. The PR adds a user-friendly error to CTS OSS when trying to use an enterprise configuration option.

```
2021/07/06 14:56:09 [ERR] (cli) error validating configuration: unsupported configuration 'terraform_version' for task "test". This option is available for Consul-Terraform-Sync enterprise when using the Terraform Cloud driver, or configure the Terraform client version within the Terraform driver block
```

Related #327


mkam (approved):
Looks good to me!

PR #333, opened 2021-07-09
Add support for Terraform v1.0

mkam:
- Sets Terraform upperbound to 1.1.0
- Updates terraform-exec to v0.14.0


version/terraform_test.go
```diff
@@ -31,7 +31,7 @@ func TestTerraformConstraint(t *testing.T) {
 			false,
 		}, {
 			"invalid upper bound",
-			"0.16.2",
+			"1.1.0",
```
lornasong, line 34: I saw the other versions have a "valid <version>" test case. Wondering what you think about including a "valid 1.0" test case as well?

lornasong (approved):
looks good to me! left a non-blocking thought

version/terraform_test.go
mkam: Sure thing, forgot about that one!

PR #335, opened 2021-07-12
Handle local module source relative paths in Terraform driver

mkam:
The logic for converting the relative path to an absolute path is specific to
the Terraform driver, so moving it from the template code.


lornasong (approved):
looks good!

PR #336, opened 2021-07-12
Fix UpdateTask race

findkim:
An API request to update a task would race against the controller's main go routine and eventually caused a deadlock for the specific edge case when a task with multiple services starts off disabled and then is enabled via CLI/API.

My understanding of the root cause: On enabling, this would be the first time the task is run. The API request would initialize the template and wait for the template to fetch all dependencies. Meanwhile the controller gets notified on those dependency changes and then also attempt to render that same template. The API request is waiting for the controller/watcher to process the dependencies and mark that template as dirty. But that controller routine waits indefinitely for the lock that the API request had obtained and cannot proceed to call `watcher.Wait` in order for the API request to render the template and complete.

Fix: prevent both the controller and API server to run the same task at a given time by monitoring the active status of a task.
* This monitors the drivers that are actively running a task by the API server
* When a driver is active, the controller will skip running it and continue resuming to wait on the Watcher for changes and process dependency updates

Overview of changes
* Connect the context used for `UpdateTask` with the client request context so the API server can exit early if the client connection closes.
* Remove intermediate struct `unit` from the controller since all of that information resides in the driver
* Add active map to `driver.Drivers` and thread safe methods
* Simplify controller inits to only return error and not the driver
* Relate the API server to the controller, this allows the code to simplify what layers need information about `event.Store` and drivers.

### DIFF
--- a/controller/readonly.go
+++ b/controller/readonly.go
@@ -2,9 +2,9 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
-	"sort"
 
 	"github.com/hashicorp/consul-terraform-sync/config"
 	"github.com/hashicorp/consul-terraform-sync/driver"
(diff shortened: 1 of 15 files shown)


controller/readonly.go
```diff
@@ -33,35 +33,26 @@ func NewReadOnly(conf *config.Config) (Controller, error) {
...
-func (ctrl *ReadOnly) Init(ctx context.Context) (*driver.Drivers, error) {
-	drivers, err := ctrl.init(ctx)
-	if err != nil {
-		return nil, err
-	}
-
-	// Sort units for consistent ordering when inspecting tasks
-	sort.Slice(ctrl.units, func(i, j int) bool {
-		return ctrl.units[i].taskName < ctrl.units[j].taskName
-	})
```
lornasong, line 45: It looks like the inspect task sorting might be removed (vs. refactored). Just wanted to double-check if that was intentional?

lornasong (approved):
Changes look good! I like the 'active' flag solution you used. Left two non-blocking questions that are just things I'm curious about.

PR #337, opened 2021-07-13
Fix handler to use http.Request context to react to client connection

findkim:
When client connection closes, the server would still attempt process the request to completion. Changes allow server to error and exit early

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -32,9 +32,7 @@ type ClientConfig struct {
 // NewClient returns a client to make api requests
 func NewClient(c *ClientConfig, httpClient httpClient) *Client {
 	if httpClient == nil {
-		httpClient = &http.Client{
-			Timeout: time.Second * 5,
-		}
+		httpClient = &http.Client{}
 	}
 	return &Client{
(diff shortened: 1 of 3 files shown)


api/client.go
```diff
@@ -32,9 +32,7 @@ type ClientConfig struct {
 // NewClient returns a client to make api requests
 func NewClient(c *ClientConfig, httpClient httpClient) *Client {
 	if httpClient == nil {
-		httpClient = &http.Client{
-			Timeout: time.Second * 5,
-		}
+		httpClient = &http.Client{}
```
lornasong, line 35: I was just curious if you were seeing that the 5 second timeout wasn't enough for this driver?

lornasong (approved):
Looks good! I like the test you included! Left a non-blocking question that I was curious about.

api/client.go
findkim: Yeah it wasn't enough to run through a full init and plan when the task had started off disabled.

lornasong: Good to know! Thank you

PR #338, opened 2021-07-13
Refactor API server to stem from controller

findkim:
The API contains identical objects as the controller and is
dependent on it. This proposes to move the API server to the
controller to keep controller details out of the main function:
drivers and event.Store

### DIFF
--- a/controller/readonly.go
+++ b/controller/readonly.go
@@ -5,10 +5,8 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"sort"
 
 	"github.com/hashicorp/consul-terraform-sync/config"
-	"github.com/hashicorp/consul-terraform-sync/driver"
 )
 
 var _ Controller = (*ReadOnly)(nil)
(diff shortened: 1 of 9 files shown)


controller/readonly.go
```diff
@@ -34,18 +32,8 @@ func NewReadOnly(conf *config.Config) (Controller, error) {
...
-func (ctrl *ReadOnly) Init(ctx context.Context) (*driver.Drivers, error) {
-	drivers, err := ctrl.init(ctx)
-	if err != nil {
-		return nil, err
-	}
-
-	// Sort units for consistent ordering when inspecting tasks
-	sort.Slice(ctrl.units, func(i, j int) bool {
-		return ctrl.units[i].taskName < ctrl.units[j].taskName
-	})
```
lornasong, line 45: It looks like the inspect task sorting might be removed (vs. refactored). Just wanted to double-check if that was intentional?

lornasong (approved):
Looks good! I like how you split out the commits - the way you grouped changes together is really readable and easy to follow along! (PR split also very nice ❤️ ). Left a non-blocking question

controller/readonly.go
findkim: sorting is removed since `units` are no longer needed, `driver.Drivers` is a map. This probably would have been better with PR https://github.com/hashicorp/consul-terraform-sync/pull/339

lornasong: Ah, no worries. I can see it easier to remove here when splitting apart changes. Makes sense why it's removed!

PR #339, opened 2021-07-13
Remove intermediate struct units

findkim:
Units became a redundant abstraction with the usage of drivers


lornasong (approved):
Nice cleanup! Agree the unit struct hasn't been very meaningful as drivers has evolved. Thanks for removing it 🎉

PR #344, opened 2021-07-13
Fix update race bug

findkim:
An API request to update a task would race against the controller's main go routine and eventually caused a deadlock for the specific edge case when a task with multiple services starts off disabled and then is enabled via CLI/API.

My understanding of the root cause: On enabling, this would be the first time the task is run. The API request would initialize the template and wait for the template to fetch all dependencies. Meanwhile the controller gets notified on those dependency changes and then also attempt to render that same template. The API request is waiting for the controller/watcher to process the dependencies and mark that template as dirty. But that controller routine waits indefinitely for the lock that the API request had obtained and cannot proceed to call watcher.Wait in order for the API request to render the template and complete.

Fix: prevent both the controller and API server to run the same task at a given time by monitoring the active status of a task.
* This monitors the drivers that are actively running a task by the API server
* When a driver is active, the controller will skip running it and continue resuming to wait on the Watcher for changes and process dependency updates

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -32,9 +32,7 @@ type ClientConfig struct {
 // NewClient returns a client to make api requests
 func NewClient(c *ClientConfig, httpClient httpClient) *Client {
 	if httpClient == nil {
-		httpClient = &http.Client{
-			Timeout: time.Second * 5,
-		}
+		httpClient = &http.Client{}
 	}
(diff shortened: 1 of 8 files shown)


lornasong (approved):
Exciting to see this bug fixed! I like the 'active' flag solution you used.

PR #349, opened 2021-07-27
Add config decode error wrapper

findkim:
The mapstructure error is not intuitive to interpret when a user configures repeated blocks for a configuration that is not supported.

With the same config file `test.hcl`
```
log_level = "DEBUG"

consul {}

consul {}
```

Current error
```
$ consul-terraform-sync -inspect -config-file test.hcl
2021/07/27 16:50:02 [DEBUG] (config) mapstructure decode failed
2021/07/27 16:50:02 [ERR] (config) failed decoding content from file: test.hcl
2021/07/27 16:50:02 [ERR] (cli) error building configuration: 1 error(s) decoding:

* 'consul' expected a map, got 'slice'
```

New error
```
$ consul-terraform-sync -inspect -config-file test.hcl
2021/07/27 16:50:44 [DEBUG] (config) mapstructure decode failed
2021/07/27 16:50:44 [ERR] (config) failed decoding content from file: test.hcl
2021/07/27 16:50:44 [ERR] (cli) error building configuration: only one 'consul' block can be configured
```

### DIFF
--- a/config/decode.go
+++ b/config/decode.go
@@ -1,0 +1,46 @@
+package config
+
+import (
+	"fmt"
+	"regexp"
+	"sort"
+	"strings"
+
+	"github.com/mitchellh/mapstructure"
+)
+
+var reRepeatedBlock = regexp.MustCompile(`'([^\']+)' expected a map, got 'slice'`)
+
+func processUnusedConfigKeys(md mapstructure.Metadata, file string) error {
+	if len(md.Unused) == 0 {
+		return nil
+	}
+
+	sort.Strings(md.Unused)
+	err := fmt.Errorf("'%s' has invalid keys: %s", file, strings.Join(md.Unused, ", "))
+
+	for _, key := range md.Unused {
+		if key == "provider" {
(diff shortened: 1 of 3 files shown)


config/decode.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// decodeError is a middleware for mapstructure.Decoder.Decode() errors
+// with decode hooks specifically for CTS.
+func decodeError(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	r := regexp.MustCompile(`'([^\']+)' expected a map, got 'slice'`)
```
lornasong, line 39: I usually see regex as a global variable because it's safer since `regex.MustCompile` can panic. Is there maybe a use-case for declaring the regex locally?

lornasong (approved):
the new error message is much more helpful and actionable!

config/decode.go
findkim: oh that's true about the panic, vs global it'll panic earlier. It was an oversight 🙂

PR #350, opened 2021-07-28
Fix and enable benchmarks

mkam:
- Fixes issues for when parts of tests are run multiple times
  - Add a dependency change before a run
  - Reinitialize the controller context per run
  - Remove the multiple calls of the Once method
- Enables benchmarks and adds a CircleCI step to print out just the results of the benchmarks without other logs
- Fixes some instances where the temporary directory wasn't being used for the tasks 
- Updates Consul services names to prevent warning log

For context, benchmarks were failing because of issues when parts of the tests were run multiple times. For example, a run might pass the first time because previously there was a change in a dependency, but on the second run, there would have been no change. The tests would then hang indefinitely on that second run, waiting for a dependency change.

These fixes take into account that there are not just `for n := 0; n < b.N; n++` loops that will run multiple times but also that the entire sub-benchmarks will potentially run multiple times with different values of `b.N`.

### DIFF
--- a/e2e/benchmarks/tasks_concurrent_test.go
+++ b/e2e/benchmarks/tasks_concurrent_test.go
@@ -67,14 +67,8 @@ func benchmarkTasksConcurrent(b *testing.B, numTasks, numServices int) {
 		}
 	})
 
-	b.Run("once mode", func(b *testing.B) {
-		// Run through once to initialize Terraform workspaces to isolate the
-		// concurrent execution to only Terraform applies.
-		for n := 0; n < b.N; n++ {
-			err = rwCtrl.Once(ctx)
-			require.NoError(b, err)
-		}
-	})
+	err = rwCtrl.Once(ctx)
+	require.NoError(b, err)
 
 	b.Run("task concurrent execution", func(b *testing.B) {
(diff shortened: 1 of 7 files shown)


e2e/benchmarks/tasks_concurrent_test.go
```diff
@@ -67,19 +68,14 @@ func benchmarkTasksConcurrent(b *testing.B, numTasks, numServices int) {
...
-		}
-	})
+	err = rwCtrl.Once(ctx)
+	require.NoError(b, err)
 
 	b.Run("task concurrent execution", func(b *testing.B) {
 		// This is the crux of the benchmark which evaluates the performance of
 		// tasks triggered and executing concurrently.
 		for n := 0; n < b.N; n++ {
+			ctx, ctxCancel = context.WithCancel(context.Background())
```
findkim, line 78: the `ctxCancel` in L52 is no longer used and is overwritten by this change. Can you remove L52 and simplify the init in L66 to `err = rwCtrl.Init(context.Background())`?

findkim (approved):
🎉🎉🚀

excited to have benchmarks back!

PR #351, opened 2021-08-05
Docker docs

findkim:
Adds run instructions for docker container and build instructions for the docker image.

PR includes other small quality of life improvements I stumbled upon:
* Bugfix: Consul address env was never read because a default was set.
* Improve Dockerfile required argument to have a default name since `consul-terraform-sync` is expected in the entry script.
* Changes the order of installing Terraform only after the Consul connection has been tested to fail earlier and faster.

### DIFF
--- a/command/cli.go
+++ b/command/cli.go
@@ -181,11 +181,6 @@ func (cli *CLI) runBinary(configFiles, inspectTasks config.FlagAppendSliceValue,
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	if err := controller.InstallDriver(ctx, conf); err != nil {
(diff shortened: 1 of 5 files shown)


command/cli.go
```diff
@@ -221,6 +216,12 @@ func (cli *CLI) runBinary(configFiles, inspectTasks config.FlagAppendSliceValue,
 	}
 	defer ctrl.Stop()
 
+	// Install the driver after controller has tested Consul connection
```
lornasong, line 219: Good idea to fail fast and install driver later!

lornasong (approved):
docker docs ✨ i like all of the different sections on how to accomplish different things - they're very readable and actionable. nice quality of life improvements! the separate commits with descriptions very helpful for me ❤️

Issue #23, opened 2020-08-12
Setup TF printer client log to filter log levels


findkim:
The printer client is used for testing purposes to circumvent executing the TF CLI when not needed. It uses a different logger from the global logger to isolate for unit tests. It would be ideal to have this client's logger to also be configured to filter the same log levels as the global one does on startup.

wilkermichael:
`INFO` is a good default for learning a tool, `WARN` is a good production default. Our tools generally are fairly simple to understand and pick up and so having a production level default is nice to keep the configurations smaller and avoid log spam. However, as CTS evolves with more features, and given the general consensus from the team that our `INFO` logs don't over-flood the console, we have decided to go ahead an change the default log level to `INFO`

PR #352, opened 2021-08-06
Issue23 printer log level support

wilkermichael:
This PR is to implement the enhancement as specified in https://github.com/hashicorp/consul-terraform-sync/issues/23

### DIFF
--- a/client/printer_test.go
+++ b/client/printer_test.go
@@ -3,7 +3,6 @@ package client
 import (
 	"bytes"
 	"context"
-	"fmt"
 	"testing"
 
 	"github.com/hashicorp/consul-terraform-sync/logging"
@@ -99,7 +98,6 @@ func TestPrinterSetEnv(t *testing.T) {
 	assert.Contains(t, buf.String(), "client.printer")
 	assert.Contains(t, buf.String(), "set")
 	assert.Contains(t, buf.String(), "env")
-	fmt.Println(buf.String())
 }
 
 func TestPrinterSetStdout(t *testing.T) {
--- a/client/printer_test.go
+++ b/client/printer_test.go
@@ -4,14 +4,11 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"github.com/hashicorp/consul-terraform-sync/logging"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
-)
+	"github.com/hashicorp/consul-terraform-sync/logging"
 
-const (
-	testDefaultLogLevel = "INFO"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestNewPrinter(t *testing.T) {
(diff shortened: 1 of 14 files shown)


wilkermichael:
One thing of note is that the old printer printed local time, with the new printer it will print UTC time, same as the global logger. Let me know if this is ok, but I think it's clearer.

wilkermichael:
@hashicorp/api-tooling-core this is now ready for review

client/printer_test.go
```diff
@@ -77,6 +99,7 @@ func TestPrinterSetEnv(t *testing.T) {
 	assert.Contains(t, buf.String(), "client.printer")
 	assert.Contains(t, buf.String(), "set")
 	assert.Contains(t, buf.String(), "env")
+	fmt.Println(buf.String())
```
findkim, line 102: Can you remove the print? 🙂

logging/logging.go
```diff
@@ -5,15 +5,24 @@ import (
...
+	gsyslog "github.com/hashicorp/go-syslog"
 	"github.com/hashicorp/logutils"
 )
 
 // Levels are the log levels we respond too.
 var Levels = []logutils.LogLevel{"TRACE", "DEBUG", "INFO", "WARN", "ERR"}
 
+// LogLevel is the log level of all logs in this package
+// it is set by the Setup() function and is shared with all other setup variations (i.e. SetupLocal())
+var LogLevel = ""
```
findkim, line 20: Appreciate the context provided as a comment!

logging/logging.go
```diff
@@ -5,15 +5,24 @@ import (
...
+
+const (
+	defaultLogLevel = "INFO"
```
findkim, line 23: The default log level in the config package is using `WARN` ([code](https://github.com/hashicorp/consul-terraform-sync/blob/master/config/config.go#L21)). I'm not necessarily advocating that warn is the right level to default to, but wanted to open up the discussion since it is different. What log level would you suggest for consistency?

findkim (approved):
The changes look good 🙌 , nice job on keeping the changes small in scope + added tests

logging/logging.go
wilkermichael: Thanks!

logging/logging.go
wilkermichael: I think WARN is better you're right, we should receive warnings if things aren't directly setup by default.

I've also moved the setting of the default log level globally to where the default logger is created in NewLogFilter for consistency. Doesn't really make

wilkermichael:
`INFO` is a good default for learning a tool, `WARN` is a good production default. Our tools generally are fairly simple to understand and pick up and so having a production level default is nice to keep the configurations smaller and avoid log spam. However, as CTS evolves into a larger tool, and given the general consensus from the team that our `INFO` logs don't over-flood the console, we have decided to go ahead an change the default log level to `INFO`

Issue #299, opened 2021-06-07
Regex support for service trigger


findkim:
#### Description
Allow user to specify a regex to match the service name in Consul catalog within the task configuration block. Task should be triggered for any changes to the service matching the regex in the Consul catalog. Current implementation of CTS for a service trigger needs the entire service name to be specified in the `task.services` list.

```hcl
task {
  name = "my-task"
  services = ["web", "web-sidecar-proxy", "web-microservice1", "web-microservice2"]
}
```

#### Use Cases
CTS to automate Terraform for discovered services registered with the Consul catalog without requiring to explicitly update CTS configuration (which requires restarting CTS).

#### Alternative Solutions
```hcl
task {
  name = "my-task"
  services = []
  conditions {
    regexp = "^web.*"
  }
}
```

#### Additional context
How would a user know the names of service that match the regex?

sofixa:
That would be a really useful feature, especially when Consul services are automatically registered by an orchestrator ( like Nomad).

mkam:
The regex support has been added, and support for query parameters in addition to the regex will be handled in a follow up feature (https://github.com/hashicorp/consul-terraform-sync/issues/357).

PR #353, opened 2021-08-09
Regex support for service trigger

mkam:
Adds support for specifying  a regex to match the service name in Consul catalog within the task configuration block. The configuration would look like the following:

```hcl
task {
  name = "condition-services-task"
  description = "task that triggers only for services that match the regex"
  source = "mkam/hello/cts"
  condition "services" {
      regexp = "^web.*"
  }
}
```
Note:
- If a non-empty services list is provided, CTS errors. The expectation is that if both an explicit list and a regex is needed, then the list should be included in the regex. My thoughts were that this was a cleaner solution for the code and removed any ambiguity on how the task is triggered.
- Currently does not support datacenter/namespace/node_meta/filter query parameters. The custom regex template  method supports these query options, but adding the configuration options and using them will be done in a future PR. This also means that we can only support regex in the default datacenter and default namespace.
- Similarly, these changes do not take in account any [service blocks](https://www.consul.io/docs/nia/configuration#service). Even if the id of the service matches the regex, it will not be used.

Code Changes:
- Adds the `regexp` configuration option to the services condition
- Adds a custom template method that finds all services, determines which services names match the regex, and then gets the health data for each of the matching services
- Sets the services variable with the custom template method and not the default template if the condition is specified

Related GitHub Issue: https://github.com/hashicorp/consul-terraform-sync/issues/299

### DIFF
--- a/config/task.go
+++ b/config/task.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"log"
 	"path/filepath"
 	"strings"
 
@@ -270,7 +271,7 @@ func (c *TaskConfig) Validate() error {
 		}
 		switch cond := c.Condition.(type) {
 		case *ServicesConditionConfig:
-			if cond.Regexp == nil ||  *cond.Regexp == ""  {
+			if cond.Regexp == nil || *cond.Regexp == "" {
 				return fmt.Errorf("at least one service is required in task.services " +
 					"or task.condition.regexp must be configured")
 			}
@@ -285,7 +286,11 @@ func (c *TaskConfig) Validate() error {
 		switch cond := c.Condition.(type) {
 		case *ServicesConditionConfig:
 			if cond.Regexp != nil && *cond.Regexp != "" {
-				return fmt.Errorf("task.services is not allowed if task.condition.regexp is configured")
+				log.Printf("[ERR] (config.task) list of services and service condition regex " +
+					"both provided. If both are needed, consider including the list in the regex " +
+					"or creating separate tasks.")
+				return fmt.Errorf("task.services is not allowed if task.condition.regexp " +
+					"is configured for a services condition")
 			}
 		}
 	}
--- a/config/task.go
+++ b/config/task.go
@@ -259,7 +259,7 @@ func (c *TaskConfig) Validate() error {
 	// We'll validate early resembling Terraform restrictions to surface any errors
 	// before a task is ran.
 	if !hclsyntax.ValidIdentifier(*c.Name) {
-		return fmt.Errorf("A task name must start with a letter or underscore and "+
+		return fmt.Errorf("a task name must start with a letter or underscore and "+
 			"may contain only letters, digits, underscores, and dashes: %q", *c.Name)
 	}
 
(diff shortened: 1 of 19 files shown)


lornasong (approved):
looks good to me! really thorough test coverage. exciting to have this feature

config/task.go
```diff
@@ -270,15 +270,24 @@ func (c *TaskConfig) Validate() error {
...
 				return fmt.Errorf("catalog-services condition requires either" +
 					"task.condition.regexp or at least one service in " +
 					"task.services to be configured")
 			}
 		}
+	} else {
+		switch cond := c.Condition.(type) {
+		case *ServicesConditionConfig:
+			if cond.Regexp != nil && *cond.Regexp != "" {
+				return fmt.Errorf("task.services is not allowed if task.condition.regexp is configured")
```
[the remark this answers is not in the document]
mkam: That makes sense, I'll clarify the exact situation! I was thinking of also adding some suggestions if both are needed, like combining the list into the regex or creating a separate task.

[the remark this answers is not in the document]
lornasong: I think providing suggestions on how to address an error would be a nice user experience!

templates/tftmpl/condition.go
```diff
@@ -40,24 +44,61 @@ type Condition interface {
...
+}
+
 func (c ServicesCondition) appendModuleAttribute(body *hclwrite.Body) {}
 
-func (c ServicesCondition) appendTemplate(io.Writer) error {
-	// no-op: services condition currently requires no additional condition
-	// templating. it relies on the monitoring template as the run condition
+func (c ServicesCondition) appendTemplate(w io.Writer) error {
+	q := c.hcatQuery()
+	if c.Regexp != "" {
```
findkim, line 63: Minor go nit that would reduce the white space by returning early

```go
  if c.Regexp == "" {
    return nil
  }
  // Proceed with c.Regexp != "" logic
```

templates/tftmpl/condition.go
```diff
@@ -40,24 +44,61 @@ type Condition interface {
...
+		_, err := fmt.Fprintf(w, serviceRegexTmpl, q)
+		if err != nil {
+			log.Printf("[WARN] (templates.tftmpl) unable to write service condition template")
```
findkim: I am wondering if there's more context for the log level `WARN`.

This log line corresponds to an error, do you know if this error would terminate CTS during startup? If so, changing the log level to `ERR` would be more representative.

templates/tftmpl/tmplfunc/service_regex.go
```diff
@@ -0,0 +1,258 @@
...
+	regexp *regexp.Regexp
+
+	filter   string
+	dc       string
+	ns       string
+	nodeMeta map[string]string
+	opts     consulapi.QueryOptions
+}
+
+// newCatalogServicesRegistrationQuery processes options in the format of
```
findkim, line 65: Can you update the godoc with the new function name?

templates/tftmpl/condition.go
mkam: Not sure, let me try to force an error and see the behavior!

templates/tftmpl/tmplfunc/service_regex.go
```diff
@@ -0,0 +1,258 @@
...
+			return value.([]*dep.HealthService), nil
+		}
+
+		return result, nil
+	}
+}
+
+// serviceRegexQuery is the representation of the regex service
+// query from inside a template.
+type serviceRegexQuery struct {
```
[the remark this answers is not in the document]
mkam: That makes sense to me! I originally went with the singular service because I was thinking of how hcat has service for health service and services for catalog services. But now that you've pointed it out, I realize that it's returning multiple services (unlike the `service` method) so plural makes more sense.

templates/tftmpl/condition.go
mkam: It does terminate CTS on the error, both here and with catalog services condition. I'll update both places to have `ERR` instead of `WARN`.

PR #354, opened 2021-08-10
Updates to testutils

lornasong:
Various test improvements I came across

Changes:
 - Stop watcher in TestServicesRenderRace
 - Add testutils.Setenv()
 - Tidy up testutils.FreePort()

### DIFF
--- a/templates/hcltmpl/integration_test.go
+++ b/templates/hcltmpl/integration_test.go
@@ -5,7 +5,6 @@ package hcltmpl
 import (
 	"context"
 	"io/ioutil"
-	"os"
 	"testing"
 	"time"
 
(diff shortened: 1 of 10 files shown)


templates/hcltmpl/integration_test.go
```diff
@@ -18,20 +17,8 @@ import (
...
-		defer os.Unsetenv("CTS_TEST_ENV")
-	}
-	os.Setenv("CTS_TEST_ENV", "foobar")
-
-	cachedTestDNEEnv, ok := os.LookupEnv("CTS_TEST_DNE")
-	if ok {
-		defer os.Setenv("CTS_TEST_DNE", cachedTestDNEEnv)
-	} else {
-		defer os.Unsetenv("CTS_TEST_DNE")
-	}
```
lornasong, line 34: I removed this since it wasn't setting anything but may have misunderstood

findkim: That's fair! `CTS_TEST_DNE` is used in a test case, but it's very unlikely that the host has that environment set to set it back. This was to ensure nothing is currently set before the test case is run, or if it did exist it wouldn't just clean someone's setup from this env.

findkim (approved):
I love your cleanup PRs. The env cleanup is lovely 😍

templates/hcltmpl/integration_test.go
lornasong: Gotcha! Thanks for checking and clarifying

PR #355, opened 2021-08-11
Update tests to use new global working_dir option

findkim:
https://github.com/hashicorp/consul-terraform-sync/issues/314


lornasong (approved):
looks good to me!

PR #356, opened 2021-08-16
Refactor services condition to more standard pattern

mkam:
Updated services condition to use SourceIncludesVar to determine whether
to set the template as the services variable, similar to how catalog
condition sets the catalog_services variable. For the Terraform driver,
this is always going to be set to true.

Related to https://github.com/hashicorp/consul-terraform-sync/issues/299


lornasong (approved):
makes sense to me

PR #358, opened 2021-08-17
Add condition "nodes" that triggers a task everytime a Consul node is updated

remilapeyre:


config/condition_nodes.go
```diff
@@ -0,0 +1,88 @@
+package config
+
+import "fmt"
+
+const nodesConditionTypes = "nodes"
+
+type NodesConditionConfig struct {
+	Datacenter *string `mapstructure:"datacenter"`
```
findkim, line 8: I see that the query parameter `Filter` is available and supported in the [node template](https://github.com/hashicorp/consul-terraform-sync/pull/358/files#diff-bedd4bfa666cdb3e010c470094d77cfe00ecf368feeb37a03a2a97042b59bfacR206-R209), it would be a good feature to expose through the configuration to be used.

I like the omission of the `node-meta` parameter since it's marked as [deprecated for the `/v1/catalog/nodes` endpoint](https://www.consul.io/api-docs/catalog#node-meta)

templates/tftmpl/condition.go
```diff
@@ -202,3 +202,83 @@ variable "catalog_services" {
...
+	}
+
+	if len(opts) > 0 {
+		return `"` + strings.Join(opts, `" "`) + `" ` // deliberate space at end
+	}
+	return ""
+}
+
+const nodesRegexTmpl = `
+nodes = [
```
findkim, line 260: The attributes in the variable object look good 👍

We have a larger design choice for CTS to generate input variables as maps instead of lists for the purpose of modules to easily consume it with the [Terraform `for_each` syntax](https://www.terraform.io/docs/language/meta-arguments/for_each.html). Would you mind changing it from a list to map?

As for the key of the map, we could use `id.node.datacenter` for uniqueness. Similar to the ID of the [`var.services` input variable](https://github.com/hashicorp/consul-terraform-sync/blob/afe4d8d1225378fa81fcb5f7dad37ab71fa5295d/templates/tftmpl/tfvars.go#L98)

templates/tftmpl/tmplfunc/nodes.go
```diff
@@ -0,0 +1,140 @@
...
+	"strings"
+
+	consulapi "github.com/hashicorp/consul/api"
+	"github.com/hashicorp/go-bexpr"
+	"github.com/hashicorp/hcat"
+	"github.com/hashicorp/hcat/dep"
+	"github.com/pkg/errors"
+)
+
+func nodesFunc(recall hcat.Recaller) interface{} {
```
findkim, line 15: Although these functions are not public, can you add code comments for them since they contain a lot of broader context that is not immediately noticeable.

You can follow the general format found here https://github.com/hashicorp/consul-terraform-sync/blob/afe4d8d1225378fa81fcb5f7dad37ab71fa5295d/templates/tftmpl/tmplfunc/catalog_services_registration.go#L15-L22

findkim (commented):
Hi @remilapeyre. We are excited about your contribution! @lornasong has caught me up to speed with communication between you two. We rotate community duties and so happy to jump in here and update you on where we are:

After reviewing the changes, the approach and proposed feature look good. It’s great to see how the `nodes` feature you have added extends our new construct of `conditions` quite well.

There are a few items that would get this PR and feature ready.
* Can you write up a PR summary to include: a high level summary of the feature + use case, and a code block with a config example setting a task with the node condition?
* Can you add an integration test and an end-to-end test for the node condition template? Here are some examples that might help with that: [catalog-services integration](https://github.com/hashicorp/consul-terraform-sync/blob/master/templates/tftmpl/integration_test.go#L122-L130), [services-regex e2e](https://github.com/hashicorp/consul-terraform-sync/blob/master/e2e/condition_test.go#L572)

We are planning to release 0.3.0 soon, and so your feature could make it into the next release! It shouldn't be too long after 0.3.0 🙂

findkim:
Adding one more comment that is not an ask.

The content added in [templates/tftmpl/tmplfunc/nodes.go](https://github.com/hashicorp/consul-terraform-sync/blob/afe4d8d1225378fa81fcb5f7dad37ab71fa5295d/templates/tftmpl/tmplfunc/nodes.go) and its test file would be great to move to the [hashicorp/hcat library](https://github.com/hashicorp/hcat/blob/master/consul_v1.go#L13-L28). It's where general Consul API handling is done for templates. The library will be also adopted by Consul-Template, and I can imagine that others would benefit from your addition for `nodes` and filtering support!

@mkam has some hcat work inflight too, so that's why it's not an ask for you to move it there. We can handle that and updating the CTS dependency on the library afterwards.

mkam:
Hi @remilapeyre, just for awareness, we did a fairly major refactoring of our conditions code (see https://github.com/hashicorp/consul-terraform-sync/pull/385), so your PR is out of date with how conditions are structured now. Please let us know how you'd like to proceed with your changes!

PR #359, opened 2021-08-18
Fix issue with resolving global and task level buffer_period

lornasong:
There is a global-level and task-level BufferPeriod configuration where task-
level BufferPeriod config overrides the global-level config for a given task.

This caused an issue when resolving the global and task level BufferPeriod config
where the task-level would not override the global-level when task-level
buffer period was disabled.

As part of the solution, update BufferPeriodConfig so that the global-task-level
override is resolved at the config-level rather than the controller level. Also
included some additional changes on how BufferPeriodConfig is finalized

Changes:
 - Update BufferPeriodConfig Finalize() to resolve with a parent BufferPeriodConfig.
 Also update some Finalize() behavior
 - Thread new Finalize() behavior through config
 - Remove getTemplateBufferPeriod() - the controller-level function to resolve
 BufferPeriodConfig - and replace with just consuming the task-level
 BufferPeriodConfig
 - Remove DefaultTaskBufferPeriodConfig() which would be unused with new 
 changes

Not sure if this PR makes unintentional behavior changes to buffer_period. I believe it
still aligns with how the docs describe the config. Feedback welcome!


config/buffer_period_test.go
```diff
@@ -156,33 +156,115 @@ func TestBufferPeriodConfig_Finalize(t *testing.T) {
...
+			},
+			DefaultBufferPeriodConfig(),
+			&BufferPeriodConfig{
+				Enabled: Bool(true),
+				Min:     TimeDuration(5 * time.Second),
+				Max:     TimeDuration(50 * time.Second),
+			},
+		},
+		{
+			"input and parent configured",
```
findkim, line 250: Thanks for the added edge cases! While we're focused on this, could you add one more case where the input is enabled with values, and the global is disabled with values? It would be useful to test that not only the `Enabled` is overriding correctly, but also the min/max.

```go
"input enabled and parent disabled with values"
i: {
  Enabled: true
  Min: 20s
  Max: 50s
},
p: {
  Enabled: false
  Min: 1m
  Max: 5m
}
r: {
  // same as i
}
```

controller/controller.go
```diff
@@ -314,32 +322,6 @@ func getProvider(providers driver.TerraformProviderBlocks, id string) driver.Ter
 	})
 }
 
-// getTemplateBufferPeriod applies the task buffer period config to its template
-func getTemplateBufferPeriod(conf *config.Config,
```
findkim, line 318: The refactor to resolve the buffer period earlier is a good improvement and helps with clarity and expectation by the time `config.Config` is passed to the controller.

findkim (approved):

config/buffer_period_test.go
lornasong: Yes, added! Thank you

config/buffer_period.go
```diff
@@ -86,14 +77,39 @@ func (c *BufferPeriodConfig) Merge(o *BufferPeriodConfig) *BufferPeriodConfig {
 	return r
 }
 
-// Finalize ensures there no nil pointers.
-func (c *BufferPeriodConfig) Finalize() {
-	if c.Enabled == nil {
-		c.Enabled = Bool(TimeDurationPresent(c.Min))
+// Finalize ensures there no nil pointers. For nil pointers, use 'parent'
+// BufferPeriodConfig and other default values as necessary.
```
wilkermichael, line 81: In regards to this method comment, I wonder if there might be a way to re-phrase this a little so it's a bit clearer as to what the method is actually doing vs what it should be doing? When I read it the first time, I didn't quite understand that 'parent' was the name of the input variable being passed in and it wasn't some special object I had to use. Maybe something like:

```go
// Finalize ensures there are no nil pointers. When there are nil pointers, Finalize sets these values using the 
// passed in parent BufferPeriodConfig along with using other defaults where necessary.
```

config/buffer_period.go
```diff
@@ -86,14 +77,39 @@ func (c *BufferPeriodConfig) Merge(o *BufferPeriodConfig) *BufferPeriodConfig {
...
+//
+// Finalizing global-level buffer period, use DefaultBufferPeriodConfig() as
+// parent. Finalizing task-level buffer period, use global-level buffer period
+// as parent.
+func (c *BufferPeriodConfig) Finalize(parent *BufferPeriodConfig) {
```
wilkermichael, line 86: For the above comment regarding inputs to Finalize, I wonder if this should be included here? I get that we're trying to instruct how this method should be used, but it seems like a comment that might be better suited for where this method is used (like config/config.go)? 

To me it seems like Finalize itself is more agnostic towards global-level vs task-level it just uses the parent as a means for setting default values when nil is present.

config/buffer_period.go
```diff
@@ -86,14 +77,39 @@ func (c *BufferPeriodConfig) Merge(o *BufferPeriodConfig) *BufferPeriodConfig {
...
-// Finalize ensures there no nil pointers.
-func (c *BufferPeriodConfig) Finalize() {
-	if c.Enabled == nil {
-		c.Enabled = Bool(TimeDurationPresent(c.Min))
+// Finalize ensures there no nil pointers. For nil pointers, use 'parent'
+// BufferPeriodConfig and other default values as necessary.
+//
+// Finalizing global-level buffer period, use DefaultBufferPeriodConfig() as
+// parent. Finalizing task-level buffer period, use global-level buffer period
+// as parent.
```
wilkermichael, line 85: For the above comment regarding Finalizing inputs, I'm wondering if we should have these kinds of implementation details here? It seems like a comment like this might be better suited where Finalize is actually used.

To me, Finalize seems to be agnostic towards global-level vs task-level, it just cares about taking in a parent and using it to set defaults when nil values are present.

wilkermichael (approved):

wilkermichael:
Great work Lorna!

config/buffer_period.go
lornasong: Thanks for the suggested wording! It's a lot clearer. I've updated to include your language

config/buffer_period.go
lornasong: So my intention here was to give someone who comes across Finalize() additional clarity on the parent parameter. I find this a little more friendly of an experience than having to look for where its implemented for more information. I updated the language to reframe this section as example usage rather than to be seen as implementation instructions.

I tried adding the comments where the function is consumed. For me, having the comments there didn't add a notable amount of clarity to the code. For example:

```
// finalize the global-level buffer period with the default buffer period
c.BufferPeriod.Finalize(DefaultBufferPeriodConfig())

// finalize task-level buffer period with global-level buffer period
c.BufferPeriod.Finalize(bp)
```

If there's a best practice that I'm not aware of, please feel free to point me to it and I can delete the example!

wilkermichael: I see what you mean, yeah that does get a bit confusing.

I like that you added the term example, that makes it a lot clearer that this is how it **could be used** vs **has to be** used. I'm good with that change. Good idea!

PR #361, opened 2021-08-24
Register notifiers, check for no change result

mkam:
Updates hashicat to the latest, which has breaking changes:

- Notifier is no longer auto registered with the watcher
- Resolver result attribute "Complete" is always true unless
  there are unresolved fields. This makes the NoChange attribute
  more reflective of whether an update happened.


eikenb (approved):
Everything looks good with the changes to match the new semantics.
Hope it was an improvement.

wilkermichael:
I'm getting some errors when I pull it locally, `go mod tidy` resolved it/had some changes. 

Melissa would you mind trying to run `go mod tidy` and commit the changes it makes?

wilkermichael:
Looks good to me as well. It seems through the use of the 'NoChange' flag that the meaning of 'complete' is now a lot clearer in the code, and the logs are more accurately reflected!

mkam:
@wilkermichael Ran the `go mod tidy` and pushed the changes. Will remember to run that next time I make any dependency changes, thanks!

PR #364, opened 2021-08-27
Linux packaging

samsalisbury:
- Migrates to use shared [hashicorp/package](https://github.com/hashicorp/package) action and deletes embedded action.
- Removes unused pre/post install hooks and unused hashicorp/package action arguments.
- Adds _description_ field for packages.


.github/workflows/build.yml
```diff
@@ -131,11 +128,6 @@ jobs:
           binary: "dist/${{ github.event.repository.name }}"
           deb_depends: "openssl"
           rpm_depends: "openssl"
-          config_dir: "packaging/linux"
```
kpenfound, line 134: I think the action will need to be updated to handle this being nil, I can have a look at this

kpenfound (approved):
Looks good. I'm not sure if there's already a PR to rename `./packaging` to `./.release` in this repo, but that's needed too

sarahethompson:
> Looks good. I'm not sure if there's already a PR to rename `./packaging` to `./.release` in this repo, but that's needed too

Once we merge this one, I'll create a PR to move everything to ./.release cc @samsalisbury

sarahethompson (approved):

PR #366, opened 2021-08-31
fixing package names

sarahethompson:
Fixing the package names for linux/386 as we were getting:

Screenshot 2021-08-31 at 10 39 14

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,14 +70,14 @@ jobs:
         - uses: actions/upload-artifact@v2
           if: ${{ matrix.goos == 'linux' }}
           with:
-            name: ${{ github.event.repository.name }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goarch }}.deb
-            path: out/${{ github.event.repository.name }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goarch }}.deb
+            name: ${{ github.event.repository.name }}_${{ needs.get-product-version.outputs.product-version }}_i386.deb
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -70,14 +70,14 @@ jobs:
...
+            name: ${{ github.event.repository.name }}_${{ needs.get-product-version.outputs.product-version }}_i386.deb
+            path: out/${{ github.event.repository.name }}_${{ needs.get-product-version.outputs.product-version }}_i386.deb
 
         - uses: actions/upload-artifact@v2
           if: ${{ matrix.goos == 'linux' }}
           with:
-            name: ${{ github.event.repository.name }}-${{ needs.get-product-version.outputs.product-version }}.x86_64.rpm
-            path: out/${{ github.event.repository.name }}-${{ needs.get-product-version.outputs.product-version }}.x86_64.rpm
+            name: ${{ github.event.repository.name }}-${{ needs.get-product-version.outputs.product-version }}.i386.rpm
+            path: out/${{ github.event.repository.name }}-${{ needs.get-product-version.outputs.product-version }}.i386.rpm
```
samsalisbury, line 80: This whole `uses` section seems to be a duplicate of the one above, maybe we could remove it completely?

samsalisbury (approved):
Approved, but see comment inline, thanks.

.github/workflows/build.yml
sarahethompson: These are 2 separate uploads - one for the .rpm and one for the .deb

samsalisbury: Ah, sorry I missed that thanks!

PR #370, opened 2021-08-31
Add configuration for consul-kv condition

mkam:
Supports consul-kv condition in task block, where path is
a required attribute for the condition.

**Example Configuration:**
```
task {
...
  condition "consul-kv" {
    path = "test"
    source_includes_var = true
    recurse = true
    datacenter = "dc1"
    namespace = "test-namespace"
  }
...
}
```

This PR only has the configuration changes. Follow up PR will have the template and notifier changes that use the configuration.

Part of https://github.com/hashicorp/consul-terraform-sync/issues/150

### DIFF
--- a/config/task.go
+++ b/config/task.go
@@ -281,6 +281,9 @@ func (c *TaskConfig) Validate() error {
 					"task.condition.regexp or at least one service in " +
 					"task.services to be configured")
 			}
+		case *ConsulKVConditionConfig:
+			return fmt.Errorf("consul-kv condition requires at least one service to " +
+				"be configured in task.services")
 		}
 	} else {
 		switch cond := c.Condition.(type) {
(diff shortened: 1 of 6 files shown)


config/task.go
```diff
@@ -281,6 +281,8 @@ func (c *TaskConfig) Validate() error {
 					"task.condition.regexp or at least one service in " +
 					"task.services to be configured")
 			}
+		case *ConsulKVConditionConfig:
+			return fmt.Errorf("at least one service must be configured")
```
findkim, line 285: Could you add more context to this error -- like including that a task with the KV condition requires at least one service to be configured?

findkim (approved):
Config changes look good. Looking forward to the future PRs around the kv feature!

config/task.go
mkam: Added!
```
2021/09/01 15:50:16 [ERR] (cli) error validating configuration: consul-kv condition requires at least one service to be configured in task.services
```

PR #373, opened 2021-09-01
Issue 332 http request logging

### LINKED ISSUE #332: Log HTTP requests
#### Description
Add middleware to log HTTP requests in a standardized format

#### Use Cases
It's useful to have the API server log all HTTP requests for debugging, auditing, and security compliance purposes.

wilkermichael:
This issue resolves #332 

In this issue I've done the following:
1) Refactored logging to be performed by [go-hclog](https://github.com/hashicorp/go-hclog) library
2) Added HTTP middleware which logs information about a request on the way in, as well as on the way out. The middleware also adds a request ID (reqID) so that logs for a particular request can be tracked.

As part of (2) I considered creating a custom wrapper around the http.ResponseWriter so that metadata such as the status of the response could be logged, but ultimately decided that this may be a task for another issue if so desired. If anyone feels strongly that this information should be logged as part of this issue, then I can do this enhancement as well.

**For comparison, New vs Old logs**
new logs

old logs

**Sample Trace Logs with HTTP Logs at the end**
image

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"log"
 	"net/http"
 	"strings"
 	"sync"
@@ -12,6 +11,7 @@ import (
 
 	"github.com/hashicorp/consul-terraform-sync/driver"
 	"github.com/hashicorp/consul-terraform-sync/event"
+	"github.com/hashicorp/consul-terraform-sync/logging"
 )
 
 const (
@@ -50,6 +50,8 @@ const (
 	// recent task updates are stored as an ‘event’ in CTS. A task is
 	// unknown when no event data has been collected yet.
 	StatusUnknown = "unknown"
+
+	logSystemName = "api"
 )
 
 // API supports api requests to the cts binary
@@ -67,17 +69,17 @@ func NewAPI(store *event.Store, drivers *driver.Drivers, port int) *API {
 
 	// retrieve overall status
 	mux.Handle(fmt.Sprintf("/%s/%s", defaultAPIVersion, overallStatusPath),
-		newOverallStatusHandler(store, drivers, defaultAPIVersion))
+		withLogging(newOverallStatusHandler(store, drivers, defaultAPIVersion)))
 	// retrieve task status for a task-name
 	mux.Handle(fmt.Sprintf("/%s/%s/", defaultAPIVersion, taskStatusPath),
-		newTaskStatusHandler(store, drivers, defaultAPIVersion))
+		withLogging(newTaskStatusHandler(store, drivers, defaultAPIVersion)))
 	// retrieve all task statuses
 	mux.Handle(fmt.Sprintf("/%s/%s", defaultAPIVersion, taskStatusPath),
-		newTaskStatusHandler(store, drivers, defaultAPIVersion))
+		withLogging(newTaskStatusHandler(store, drivers, defaultAPIVersion)))
 
 	// crud task
 	mux.Handle(fmt.Sprintf("/%s/%s/", defaultAPIVersion, taskPath),
-		newTaskHandler(store, drivers, defaultAPIVersion))
+		withLogging(newTaskHandler(store, drivers, defaultAPIVersion)))
 
 	srv := &http.Server{
 		Addr:         fmt.Sprintf(":%d", port),
@@ -102,6 +104,7 @@ func (api *API) Serve(ctx context.Context) error {
 	var wg sync.WaitGroup
 	wg.Add(1)
 
+	logger := logging.Global().Named(logSystemName)
 	go func() {
 		defer wg.Done()
 		for {
@@ -111,18 +114,18 @@ func (api *API) Serve(ctx context.Context) error {
 				defer cancel()
 
 				if err := api.srv.Shutdown(ctxShutDown); err != nil {
-					log.Printf("[ERROR] (api) error stopping api server: '%s'", err)
+					logger.Error("error stopping api server", "error", err)
 				} else {
-					log.Printf("[INFO] (api) shutdown api server")
+					logger.Info("shutdown api server")
 				}
 				return
 			}
 		}
 	}()
 
-	log.Printf("[INFO] (api) starting server at '%d'", api.port)
+	logger.Info("starting server", "port", api.port)
 	if err := api.srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
-		log.Printf("[ERROR] (api) serving api at '%d': %s", api.port, err)
+		logger.Error("error serving api", "port", api.port, "error", err)
 		return err
 	}
 
@@ -140,8 +143,12 @@ func jsonResponse(w http.ResponseWriter, code int, response interface{}) error {
 	return json.NewEncoder(w).Encode(response)
 }
 
-func jsonErrorResponse(w http.ResponseWriter, code int, err error) error {
-	return jsonResponse(w, code, NewErrorResponse(err))
+func jsonErrorResponse(ctx context.Context, w http.ResponseWriter, code int, err error) {
+	err = jsonResponse(w, code, NewErrorResponse(err))
+	if err != nil {
+		logging.FromContext(ctx).Named(logSystemName).Error("error, could not generate json error response",
+			"error", err)
+	}
 }
 
 // getTaskName retrieves the taskname from the url. Returns empty string if no
--- a/api/middleware.go
+++ b/api/middleware.go
@@ -1,0 +1,46 @@
+package api
+
+import (
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/hashicorp/consul-terraform-sync/logging"
+	"github.com/hashicorp/go-uuid"
+)
+
+const (
+	timeFormat = "2006-01-02T15:04:05.000Z0700"
+)
+
+func withLogging(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Generate a UUID to be included with all log messages for an incoming request
+		reqID, err := uuid.GenerateUUID()
+
+		logger := logging.Global().Named(logSystemName)
+		if err != nil {
+			logging.Global().Named(logSystemName).Error("error generating uuid", "error", err)
+			return
+		}
+
+		// UUID was successfully created, so add it now
+		logger = logger.With("reqID", reqID)
+		ts := time.Now()
+
+		// Log info before calling the next handler
+		logger.Info("received request",
+			"time", ts.Format(timeFormat),
+			"remote_ip", r.RemoteAddr,
+			"uri", r.RequestURI,
+			"method", r.Method,
+			"host", r.Host)
+
+		r = r.WithContext(logging.WithContext(r.Context(), logger))
+		next.ServeHTTP(w, r)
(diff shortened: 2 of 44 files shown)


client/printer.go
```diff
@@ -29,13 +33,14 @@ type PrinterConfig struct {
 	Writer     io.Writer
 }
 
-// NewPrinter creates a new client
+// NewPrinter creates a new clientF
```
wilkermichael, line 36: I'll fix this

api/middleware.go
```diff
@@ -0,0 +1,46 @@
...
+			"uri", r.RequestURI,
+			"method", r.Method,
+			"host", r.Host)
+
+		r = r.WithContext(logging.WithContext(r.Context(), logger))
+		next.ServeHTTP(w, r)
+
+		// Log info on exit
+		logger.Info("request complete",
+			"duration", fmt.Sprintf("%dus", time.Since(ts).Microseconds()))
```
findkim, line 44: Love that the request duration is added!

api/api.go
```diff
@@ -4,12 +4,13 @@ import (
...
 	"encoding/json"
 	"fmt"
-	"log"
 	"net/http"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/hashicorp/consul-terraform-sync/logging"
+
```
findkim, line 13: Can you remove the spacing between these imports?

api/task.go
```diff
@@ -38,16 +42,19 @@ func newTaskHandler(store *event.Store, drivers *driver.Drivers,
...
 	case http.MethodPatch:
 		h.updateTask(w, r)
 	default:
 		err := fmt.Errorf("'%s' in an unsupported method. The task API "+
 			"currently supports the method(s): '%s'", r.Method, http.MethodPatch)
-		log.Printf("[TRACE] (api.task) unsupported method: %s", err)
-		jsonErrorResponse(w, http.StatusMethodNotAllowed, err)
+		logger.Trace("unsupported method", "error", err)
+		if err = jsonErrorResponse(w, http.StatusMethodNotAllowed, err); err != nil {
+			logger.Error("error, could not generate json error response", "error", err)
```
findkim, line 56: Since this is already logged on the error level + the key value "error", I think it would be better to remove the `error,` at the beginning of the log message

findkim: and similar with the other occurrences of this log

api/task.go
```diff
@@ -64,61 +71,73 @@ type UpdateTaskResponse struct {
...
 
+		if err = jsonErrorResponse(w, http.StatusBadRequest, err); err != nil {
+			logger.Error("error, could not generate json error response", "error", err)
+		}
 		return
 	}
 
 	if taskName == "" {
-		err := fmt.Errorf("No task name was included in the api request. " +
+		err := fmt.Errorf("no task name was included in the api request. " +
```
findkim, line 85: 🙌 for idiomatic go and lower-cased errors

api/task.go
```diff
@@ -133,36 +152,42 @@ func (h *taskHandler) updateTask(w http.ResponseWriter, r *http.Request) {
...
+			logger.Error("error creating new event", "error", err)
+			if err = jsonErrorResponse(w, http.StatusInternalServerError, err); err != nil {
+				logger.Error("error, could not generate json error response", "error", err)
+			}
 			return
 		}
 		defer func() {
 			ev.End(storedErr)
-			log.Printf("[TRACE] (api.task) adding event %s", ev.GoString())
+			logger.Trace("adding event", "event", ev.GoString())
```
findkim, line 163: This is a great pass at updating all of the log lines so my mind is thinking along the next iteration of improving logs. Is there a way to log the event struct as keyed values instead of a string blob? Querying on those event keys could be valuable.

api/task_status.go
```diff
@@ -42,26 +45,33 @@ func newTaskStatusHandler(store *event.Store, drivers *driver.Drivers, version s
...
+		}
 		return
 	}
 
 	include, err := include(r)
 	if err != nil {
-		log.Printf("[TRACE] (api.taskstatus) bad request: %s", err)
-		jsonErrorResponse(w, http.StatusBadRequest, err)
+		logger.Trace("bad request", "error", err)
+		if err = jsonErrorResponse(w, http.StatusBadRequest, err); err != nil {
```
findkim, line 72: There are 3 lines per the # of occurrences of this error handling. What do you think about passing the logger to the `jsonErrorResponse` helper function and do the err checking there?

client/printer.go
```diff
@@ -50,42 +55,36 @@ func NewPrinter(config *PrinterConfig) (*Printer, error) {
...
-		"'%s', workingdir: '%s'", p.workspace, p.workingDir)
+	p.logger.Info("setting environment for workspace")
 	return nil
 }
 
 // SetStdout logs out 'set standard out'
-func (p *Printer) SetStdout(w io.Writer) {
-	p.logger.Printf("[INFO] (client.printer) setting standard out for workspace: "+
-		"'%s', workingdir: '%s'", p.workspace, p.workingDir)
+func (p *Printer) SetStdout(_ io.Writer) {
```
findkim, line 63: I think omitting the `_` variable would work too, and just leave the type `SetStdout(io.Writer) {}`. Same with the other methods for `Printer`.

findkim (commented):
First pass at reviewing -- there's a lot!

config/config.go
```diff
@@ -395,7 +398,8 @@ func decodeConfig(content []byte, file string) (*Config, error) {
 		return nil, fmt.Errorf("invalid format: %s", format)
 	}
 	if err != nil {
-		log.Printf("[ERR] (config) failed to decode %s", format)
+		logger.Error("failed to decode config",
+			"file_format", format, "error", err)
```
findkim, line 402: Just wondering if you have a line length limit set in your IDE and if so how many characters do you have it at?

config/task.go
```diff
@@ -286,11 +290,13 @@ func (c *TaskConfig) Validate() error {
...
-				log.Printf("[ERR] (config.task) list of services and service condition regex " +
-					"both provided. If both are needed, consider including the list in the regex " +
-					"or creating separate tasks.")
-				return fmt.Errorf("task.services is not allowed if task.condition.regexp " +
+				err := fmt.Errorf("task.services is not allowed if task.condition.regexp " +
 					"is configured for a services condition")
+				logging.Global().Named(logSystemName+taskSubsystemName).Error("[ERR] (config.task) list of "+
+					"services and service condition regex "+
+					"both provided. If both are needed, consider including the list in the regex "+
+					"or creating separate tasks.", "error", err)
```
findkim, line 298:
```suggestion
				logging.Global().Named(logSystemName+taskSubsystemName).Error("[ERR] (config.task) list of "+
					"services and service condition regex both provided. If both are needed, consider "+
					"including the list in the regex or creating separate tasks.", "error", err)
```

a nice-to-have to reduce lines where we can but still keep it readable.

controller/readonly.go
```diff
@@ -96,14 +95,14 @@ func (ctrl *ReadOnly) checkInspect(ctx context.Context, d driver.Driver) (bool,
 	// rendering a template may take several cycles in order to completely fetch
 	// new data
 	if rendered {
-		log.Printf("[TRACE] (ctrl) template for task %q rendered: %+v", taskName, rendered)
+		ctrl.logger.Trace("template for task rendered", taskNameLogKey, taskName)
 
-		log.Printf("[INFO] (ctrl) inspecting task %s", taskName)
+		ctrl.logger.Info("inspecting task", taskName)
```
findkim, line 100: I think it's missing the key?
```suggestion
		ctrl.logger.Info("inspecting task", taskNameLogKey, taskName)
```

handler/panos.go
```diff
@@ -173,7 +177,7 @@ func (h *Panos) commit(ctx context.Context) error {
 		return err
 	}
 
-	log.Printf("[DEBUG] (handler.panos) commit successful")
+	h.logger.Debug("commit successful")
```
findkim, line 180: I thinking this would be useful to elevate from debug to info level.

logging/logging.go
```diff
@@ -40,89 +38,100 @@ type Config struct {
...
+		Output:     logOutput,
+		TimeFormat: hclog.TimeFormat,
+	})
 
+	hclog.SetDefault(logger)
 	return nil
 }
 
+// Logger is a type alias for hclog.Logger
+type Logger hclog.Logger
```
findkim, line 78: Although the Logger type is used below, it's a quicker to reference defined types at the top of the file. Could you move this up before functions?

findkim (commented):
Found some small things to update. Really appreciate this a large effort to update all of the logs! Thanks @wilkermichael I'm excited for this improvement.

command/cli.go
```diff
@@ -185,40 +187,40 @@ func (cli *CLI) runBinary(configFiles, inspectTasks config.FlagAppendSliceValue,
...
 	switch {
 	case isInspect:
-		log.Printf("[INFO] (cli) running controller in inspect mode")
+		logger.Info("running controller in inspect mode")
 	case isOnce:
-		log.Printf("[INFO] (cli) running controller in once mode")
+		logger.Info("running controller in once mode")
 	default:
-		log.Printf("[INFO] (cli) running controller in daemon mode")
+		logger.Info("running controller in deamon mode")
```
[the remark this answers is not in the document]
wilkermichael: Demon mode is way more hardcore though! Good catch

api/task.go
wilkermichael: I can definitely look into this!

api/task_status.go
wilkermichael: I agree, will do this

api/task.go
wilkermichael: I think the easiest thing to do would be a create a function like GoString(), but for KeyValues

```
func (e *Event) KeyValues() []interface{} {
	kv := []interface{"val_1", e.Val1, "val_2", e.Val2}
	return kv
}
```

The only downside is when you pass this to a function you have to be sure to pass it as a variadic number of interfaces rather than a slice:

`logger.Trace("adding event", ev.KeyValues()...)`

wilkermichael: We can definitely consider doing this in the future, I'll create an issue to track it

wilkermichael: https://github.com/hashicorp/consul-terraform-sync/issues/377

driver/terraform_download.go
```diff
@@ -173,23 +177,23 @@ func installTerraform(ctx context.Context, conf *config.TerraformConfig) (*goVer
...
+	// Create path if one doesn't already exist
 	os.MkdirAll(*conf.Path, os.ModePerm)
 
 	installedPath, err := tfinstall.ExactVersion(v.String(), *conf.Path).ExecPath(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	log.Printf("[DEBUG] (driver.terraform) successfully installed terraform %s: %s", v.String(), installedPath)
+	logger.Debug("successfully installed terraform", "version", v.String, "install_path", installedPath)
```
mkam, line 197: This still has the `v.String` instead of `v.String()`, so it's not printing out the version:

```
[DEBUG] driver.terraform: successfully installed terraform: version=0x20e0a60 install_path=/Users/...
```

wilkermichael: ah good catch, thanks!

mkam (approved):

PR #374, opened 2021-09-02
Add unit test running Once and Run to test Watcher semantics

findkim:
Once mode should not call an extra Watcher.WaitCh() which could
inadvertantly mute the first trigger after Once mode has completed
from when Run expects it.

Also improve taskNotify by simplifying how to enable for tests


wilkermichael:
Looks good to me, I think that just calling EnableTestMode() rather than having to also set a flag is a lot simpler!

Review complete

wilkermichael (approved):

PR #376, opened 2021-09-02
Consul KV condition template

mkam:
Adds the templating work for Consul KV triggers, using the latest hcat changes for `key`, `keyExists`, and `keys`. The recurse option determines whether to use `keys` or `keyExists` + `key`. This is because of the Consul client used by hcat has a get method for `recurse=false` and a separate list method for `recurse=true`.

Follow up PR will have the notifier changes so that only KV dependencies trigger tasks and the end-to-end tests.

Part of https://github.com/hashicorp/consul-terraform-sync/issues/150

### DIFF
--- a/templates/tftmpl/condition_consul_kv_test.go
+++ b/templates/tftmpl/condition_consul_kv_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestConsulKVCondition_hcatQuery(t *testing.T) {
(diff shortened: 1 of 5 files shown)


eikenb (approved):
I got curious and looked this over. Looks good.  :+1: 

The only improvement I could think of was that it did seem like it could have used a bit more in the way of comment docs. Maybe a little more to the feature summary at the top or on some of the longer methods. Not that big a deal, just a passing thought.

templates/tftmpl/condition_consul_kv_test.go
```diff
@@ -0,0 +1,125 @@
...
+{{- end}}
+}
+`,
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			w := new(strings.Builder)
+			tc.c.appendTemplate(w)
```
wilkermichael, line 125: should we require that appendTemplate returns a nil error here?

mkam: I can add that check!

Issue #308, opened 2021-06-10
Run tasks on a schedule


findkim:
#### Description
Feature to configure tasks to execute Terraform automation on a schedule instead of immediately after each relevant Consul catalog change. When the task is scheduled to run and there were no changes since the last execution, Terraform would result in no diffs, and make no changes.

```hcl
task {
  name = “example-task”
  condition "schedule" {
    cron = "*/15 * * * *" // every 15 mins
  }
}
```

#### Additional context
[Cron syntax supported by HashiCorp](https://github.com/hashicorp/cronexpr#implementation)

lornasong:
Support is added to configure a task with a schedule using a `condition "schedule" {}` configuration block. Support exists for a task to be configured with `task.services` such that when the task triggers on schedule, it will use the latest Consul service information from `task.services` for the module's source input variable.

Separate issue is open to support configuring service information by regexp: https://github.com/hashicorp/consul-terraform-sync/issues/382
Another issue will be open to support providing Consul KV information to the module

Issue #377, opened 2021-09-02
Log struct key-values instead of struct string blobs


wilkermichael:
#### Description
Consider improving logs by allowing for structs to be logged as a list of key-values rather than as a string blob. This would be beneficial in allowing these fields to be queryable.

#### Additional context
https://github.com/hashicorp/consul-terraform-sync/pull/373/commits/bb96e8ff63348c19d06e02e10f63aaa16de36856#r701211522

fransafu:
Can I try to fix this problem?

thisisommore:
Hey, can I give it a try?

wilkermichael:
Hello @thisisommore Hacktoberfest is now over, but we still value community contributions to the project! If you're still interested in contributing, feel free to submit a PR.

thisisommore:
@wilkermichael cool, I am still interested in this, will get my hands dirty and will try to create PR :)

PR #378, opened 2021-09-07
Add schedule task condition configuration

lornasong:
Supports configuring a task with a schedule condition allowing users to
configure a task that monitors services in task.services but only execute
on a configured schedule rather than be dynamically triggered on changes.

Details:
 - schedule condition's cron field is required
 - uses the hashicorp fork of `cronexpr` to validate cron expression
 - disables buffer_period and logs a warning if task's buffer_period is
 configured. buffer_period does not make sense to be enabled for a scheduled
 task

Example config:
```
task {
  name = "scheduled_task"
  description = "run task every 10 seconds"
  services = ["web", "db"]
  source = "path/to/module"
  condition "schedule" {
    cron = "*/10 * * * * * *"
  }
}
```
Part of: https://github.com/hashicorp/consul-terraform-sync/issues/308

Subsequent prs
 - Consume schedule condition configuration to trigger task on schedule
 - Support configuring monitoring services by regexp
 - Support configuring monitoring kv

### DIFF
--- a/config/condition_schedule.go
+++ b/config/condition_schedule.go
@@ -1,0 +1,102 @@
+package config
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/cronexpr"
+)
+
+const scheduleConditionType = "schedule"
+
+var _ ConditionConfig = (*ScheduleConditionConfig)(nil)
+
+// ScheduleConditionConfig configures a condition configuration block of type
+// 'schedule'. A schedule condition is triggered by a configured cron schedule
+type ScheduleConditionConfig struct {
+	Cron *string `mapstructure:"cron"`
+}
+
+// Copy returns a deep copy of this configuration.
+func (c *ScheduleConditionConfig) Copy() ConditionConfig {
+	if c == nil {
+		return nil
+	}
+
+	var o ScheduleConditionConfig
+	o.Cron = StringCopy(c.Cron)
+
+	return &o
+}
+
+// Merge combines all values in this configuration with the values in the other
+// configuration, with values in the other configuration taking precedence.
+func (c *ScheduleConditionConfig) Merge(o ConditionConfig) ConditionConfig {
+	if c == nil {
+		if isConditionNil(o) { // o is interface, use isConditionNil()
+			return nil
+		}
+		return o.Copy()
+	}
+
+	if isConditionNil(o) {
+		return c.Copy()
+	}
+
+	r := c.Copy()
+	o2, ok := o.(*ScheduleConditionConfig)
+	if !ok {
+		return r
+	}
+
+	r2 := r.(*ScheduleConditionConfig)
+
+	if o2.Cron != nil {
+		r2.Cron = StringCopy(o2.Cron)
+	}
+
+	return r2
+}
+
+// Finalize ensures there no nil pointers.
+func (c *ScheduleConditionConfig) Finalize([]string) {
+	if c == nil { // config not required, return early
+		return
+	}
+
+	if c.Cron == nil {
+		c.Cron = String("")
+	}
+}
+
+// Validate validates the values and required options. This method is recommended
+// to run after Finalize() to ensure the configuration is safe to proceed.
+func (c *ScheduleConditionConfig) Validate() error {
+	if c == nil { // config not required, return early
+		return nil
+	}
+
+	if c.Cron == nil || len(*c.Cron) == 0 {
+		return fmt.Errorf("cron config is required for schedule condition")
+	}
+
+	if _, err := cronexpr.Parse(*c.Cron); err != nil {
+		return fmt.Errorf("unable to parse schedule condition's cron config "+
+			"%q: %s. for more information on writing cron expressions, see %s",
+			StringVal(c.Cron), err, "https://github.com/hashicorp/cronexpr")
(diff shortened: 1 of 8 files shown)


config/task.go
```diff
@@ -226,6 +227,20 @@ func (c *TaskConfig) Finalize(bp *BufferPeriodConfig, wd string) {
 		c.TFVersion = String("")
 	}
 
+	bp := globalBp
+	if _, ok := c.Condition.(*ScheduleConditionConfig); ok {
+		// disable buffer_period for schedule condition
+		if c.BufferPeriod != nil {
+			log.Printf("[WARN] (config.task) disabling buffer_period for "+
```
lornasong, line 234: will update this log with new logging format after https://github.com/hashicorp/consul-terraform-sync/issues/377 is merged

wilkermichael:
General question, I'm not quite following some of the short variable names, I was wondering if you could clarify the usage of a, r, and o2? Thanks!

config/condition_schedule_test.go
```diff
@@ -0,0 +1,183 @@
...
+	}{
+		{
+			"empty",
+			[]string{},
+			&ScheduleConditionConfig{},
+			&ScheduleConditionConfig{
+				Cron: String(""),
+			},
+		},
+	}
```
wilkermichael, line 134: For completeness, I think we should also test the other conditions, namely:
```
		{
			"empty",
			[]string{},
			&ScheduleConditionConfig{},
			&ScheduleConditionConfig{
				Cron: String(""),
			},
		},		
                 {
			"nil config",
			[]string{},
			nil,
			nil,
		},		
                {
			"chron povided",
			[]string{},
			&ScheduleConditionConfig{
				Cron: String("* * * * *"),
			},
			&ScheduleConditionConfig{
				Cron: String("* * * * *"),
			},
		},
```

lornasong:
Hi @wilkermichael a, r, o are used in a number of places so feel free to clarify if there's a specific usage that you want to know more about. Below is how I interpret them throughout our config code and test code.

Regarding 'a' - I think this is mostly seen in tests as one "input" for whatever is being tested tested e.g. a.testFunc(). It may have an accompanying 'b' which is another "input" a.testFunc(b)

Regarding 'r' - In tests it is often used to indicate expected returned object. This is used in our Merge() code like `r := c.Copy()` in which case I interpret 'r' to mean the replica

Regarding 'o' - I typically think of this is as meaning 'other'. For example, in Merge() "o" is the other config to merge in.

config/condition_schedule.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if o2.Cron != nil {
+		r2.Cron = StringCopy(o2.Cron)
+	}
+
+	return r2
+}
+
+// Finalize ensures there no nil pointers.
+func (c *ScheduleConditionConfig) Finalize(services []string) {
```
wilkermichael, line 61: maybe since services isn't actually used, and since we're just adding it to satisfy the interface, we can make this anonymous?

`func (c *ScheduleConditionConfig) Finalize([]string)`

wilkermichael:
I wonder if it's worth adding a test that tests all the methods with a nil config? That way we have peace of mind that all the nil guards are in place?

wilkermichael:
@lornasong review complete! Great work, I think being able to schedule when services trigger automation will add a lot of value

wilkermichael (approved):

wilkermichael:
> Hi @wilkermichael a, r, o are used in a number of places so feel free to clarify if there's a specific usage that you want to know more about. Below is how I interpret them throughout our config code and test code.
> 
> Regarding 'a' - I think this is mostly seen in tests as one "input" for whatever is being tested tested e.g. a.testFunc(). It may have an accompanying 'b' which is another "input" a.testFunc(b)
> 
> Regarding 'r' - In tests it is often used to indicate expected returned object. This is used in our Merge() code like `r := c.Copy()` in which case I interpret 'r' to mean the replica
> 
> Regarding 'o' - I typically think of this is as meaning 'other'. For example, in Merge() "o" is the other config to merge in.

Thanks Lorna! That clarifies it

config/condition_schedule_test.go
lornasong: Good idea, added!

lornasong:
Thanks for the review @wilkermichael

> I wonder if it's worth adding a test that tests all the methods with a nil config? That way we have peace of mind that all the nil guards are in place?

I just added a nil test case to Validate() in addition to your suggestion to add a nil test case to Finalize(). This will cover the nil config test case within each individual test functions.

config/task.go
lornasong: Updated log line to:

```
logging.Global().Named(logSystemName).Named(taskSubsystemName).Warn(
	"disabling buffer_period for schedule condition. overriding "+
	"buffer_period configured for this task",
	"task_name", StringVal(c.Name), "buffer_period", c.BufferPeriod.GoString())
```
Output:
2021-09-08T12:50:12.631-0400 [WARN]  config.task: disabling buffer_period for schedule condition. overriding buffer_period configured for this task: task_name=periodic10_task buffer_period="&BufferPeriodConfig{Enabled:true, Min:8s, Max:0s}"

PR #383, opened 2021-09-09
updates to make this file as similar to ENT as possible

sarahethompson:


wilkermichael:
I don't have a ton of context on this, so I just gave it a quick look through to make sure env.PKG_NAME matched what is in ENT. From that perspective, looks good to me.

wilkermichael (approved):

99 (approved):

PR #388, opened 2021-09-22
Crt cts cleanup

sarahethompson:
- crt cleanup includes fixing push event to only trigger on master and release branches
- Add new repo value to metadata - use PKG_NAME for artifact names (not repository name)


claire-labry (approved):
again good eye on re: `PKG_NAME`!

mkam (approved):
LGTM

PR #390, opened 2021-09-22
removed deprecated "tag" field leftover in integration test

wilkermichael:


mkam (approved):
Thanks for catching that!

PR #393, opened 2021-09-28
Reorganize TLS finalize config

findkim:
TLS config loaded from env vars are different for Consul and Vault.
This reorganizes it to more easily be found in one place

### DIFF
--- a/config/tls.go
+++ b/config/tls.go
@@ -1,6 +1,10 @@
 package config
 
-import "fmt"
+import (
+	"fmt"
+
+	vaultapi "github.com/hashicorp/vault/api"
+)
 
 const (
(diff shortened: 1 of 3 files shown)


config/tls.go
```diff
@@ -101,6 +105,37 @@ func (c *TLSConfig) Finalize() {
 			StringPresent(c.ServerName) ||
 			BoolPresent(c.Verify))
 	}
+	if c.CACert == nil {
```
mkam, line 108: Just wondering, are we using the non-Consul/non-Vault specific `Finalize` anywhere else?

mkam (approved):
Looks good to me!

PR #400, opened 2021-09-30
update alpine version used in docker image

eikenb:
3.13 had open security issues.
Changing to use the latest alpine:3 to better stay up to date.
Note alpine:3 is currently an alias for 3.14, which fixes the security issues.


wilkermichael (approved):
@eikenb this looks good, thanks!

mkam (approved):

PR #401, opened 2021-09-30
Replace keyExists/key combination with keyExistsGet

mkam:
Having the two dependencies where the second dependency (the get) depends on the
value of the first dependency (if it exists) has made this too much of a special
case and adds too much complexity. Replacing with a single dependency keeps it
more in line with all our other templates.

### DIFF
--- a/e2e/condition_consul_kv_test.go
+++ b/e2e/condition_consul_kv_test.go
@@ -170,7 +170,7 @@ func TestConditionConsulKV_NewKey(t *testing.T) {
 
 // TestConditionConsulKV_ExistingKey runs the CTS binary using a task with a consul-kv
 // condition block, where the monitored KV pair will exist initially in Consul. The
-// test will update the value and delete the key.
+// test will update the value, delete the key, and then add the same key back.
 func TestConditionConsulKV_ExistingKey(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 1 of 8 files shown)


e2e/condition_consul_kv_test.go
```diff
@@ -270,6 +270,18 @@ func TestConditionConsulKV_ExistingKey(t *testing.T) {
...
+			now = time.Now()
+			value = "new-test-value-2"
+			srv.SetKVString(t, path, value)
+			api.WaitForEvent(t, cts, taskName, now, defaultWaitForEvent)
+			eventCountNow = eventCount(t, taskName, cts.Port())
+			eventCountBase++
+			require.Equal(t, eventCountBase, eventCountNow,
+				"event count did not increment once. task was not triggered as expected")
+			content = testutils.CheckFile(t, true, resourcesPath, fmt.Sprintf("%s.txt", path))
+			assert.Equal(t, value, content)
```
wilkermichael, line 284: perfect :D

templates/tftmpl/notifier/consul_kv_test.go
```diff
@@ -195,51 +147,17 @@ func Test_ConsulKV_Notify_Once_Mode_Single_Key(t *testing.T) {
...
 		tmpl.AssertExpectations(t)
 	})
 
-	t.Run("consul-kv-last-with-existing-key", func(t *testing.T) {
+	t.Run("consul-kv-last", func(t *testing.T) {
 		// Test that notifier notifies at the end of once-mode, particularly for
 		// the case when the consul-kv dependency is received last.
 
-		// Notifier in test has 3 dependencies: 1 services, 1 Consul KV exists, 1 Consul KV value
+		// Notifier in test has 2 dependencies: 1 services, 1 Consul KV exists
```
wilkermichael, line 158: I guess the dependencies are actually 1 services, 1 Consul KV? probably don't need to specify exists.

wilkermichael (approved):
@mkam looks great! Thanks so much for doing this, you're a superstar!

PR #403, opened 2021-09-30
Issue 389 scheduled tasks support to monitor consul kv

### LINKED ISSUE #389: Scheduled tasks support to monitor Consul KV
#### Description
Add support for a new monitor block for the type "consul-kv" that can be configured alongside a schedule condition block. This monitor-consul-kv block will be similar to the existing "consul-kv" condition block in configuration (with the exception that condition-consul-kv supports source_includes_var that monitor-consul-kv should not support).

```hcl
task {
  name = "scheduled_task"
  source = "path/to/module"
  condition "schedule" {
    cron = "*/10 * * * * * *"
  }
  monitor "consul-kv" {
    path = "test"
    recurse = true
    datacenter = "dc1"
    namespace = "test-namespace"
  }
}
```

Things to Consider:
We will support other types of monitor blocks e.g. "services" monitor block
Only allow a monitor block to be configured with a "schedule" condition
Only allow at most one monitor block to be configured per task block (this may change in the future)

#### Use Cases
Allow users to have scheduled tasks monitor Consul KV information.

#### Additional context

Code already exists to run a scheduled tasks with support for configuring monitoring services with  task.services. For example in the configuration below, the user can configure task.services to monitor web and db so that when the task triggers every 10 seconds, the tasks will use web and db services information

```hcl
task {
  name = "scheduled_task"
  description = "run task every 10 seconds"
  services = ["web", "db"]
  source = "path/to/module"
  condition "schedule" {
    cron = "*/10 * * * * * *"
  }
}
```

wilkermichael:
This PR resolves #389

The purpose of this PR is to add support for the source_input block type "consul-kv." This block can be configured alongside a schedule condition block. The use case for this block is to enable a means of defining Consul objects (in this case key-values in consul-kv) whose values are provided as the task source’s input variables.

For this iteration, a source_input block is only allowed to be configured with a "schedule" condition and at most one source_input block can be configured per task block.

For simplicity and to aid in a possible refactor of condition "consul-kv" in the future, this first iteration will have the `source_input "consul-kv"` code block closely match its condition counterpart. This means that like condition "consul-kv" if `source_input "consul-kv"` is configured, task.services will be required. Likewise the fields available to source_input "consul-kv" will mirror those found in condition "consul-kv" with one caveat. The `source_includes_var` field will not be available in `source_input "consul-kv"` since it is implicitly true by the nature of a source_input.


config/task.go
```diff
@@ -555,13 +555,16 @@ func (c *TaskConfig) validateSourceInput() error {
 					return fmt.Errorf("at least one service is required in task.services " +
 						"or task.source_input.regexp must be configured")
 				}
+			case *ConsulKVSourceInputConfig:
+				return fmt.Errorf("consul-kv condition requires at least one service to " +
```
wilkermichael, line 559: I'll fix this, should read `consul-kv source_input`

controller/readwrite.go
```diff
@@ -228,6 +228,7 @@ func (rw *ReadWrite) Once(ctx context.Context) error {
 
 		select {
 		case err := <-rw.watcher.WaitCh(ctx):
+			rw.logger.Debug("waiting watcher on channel")
```
wilkermichael, line 231: I'll remove this

wilkermichael:
The SourceIncludesVariable() Methods are now part of the monitor base struct, and always just return true unless overridden in a derived struct.

There were cases where I needed to be able to pass a monitor interface to a function where it could be either a source_input or condition, and the interface required that SourceIncludesVariable() be present.

config/monitor_consul_kv.go
```diff
@@ -4,16 +4,18 @@ import (
 	"fmt"
 )
 
-const consulKVMonitorType = "consul-kv"
+const consulKVType = "consul-kv"
 
 var _ MonitorConfig = (*ConsulKVMonitorConfig)(nil)
 
+// ConsulKVMonitorConfig configures a configuration block adhering to the monitor interface
+// of type 'consul-kv'. A catalog-services monitor watches for changes
```
wilkermichael, line 12: Will fix this comment to be `A consul-kv monitor`

driver/terraform.go
```diff
@@ -119,7 +119,12 @@ func NewTerraform(config *TerraformConfig) (*Terraform, error) {
 		for k, v := range taskEnv {
 			env[k] = v
 		}
-		tfClient.SetEnv(env)
+		err = tfClient.SetEnv(env)
+		if err != nil {
+			logger.Error("error setting the environment for the client",
+				"client_type", config.ClientType, "error", err)
+			return nil, err
```
wilkermichael, line 126: Since we weren't checking this error before... maybe I should just log a warning and then do nothing? (not return nil,err). I think it would still be good to check for the error returned here

templates/tftmpl/integration_test.go
```diff
@@ -109,6 +109,8 @@ func TestRenderTFVarsTmpl(t *testing.T) {
 		registerAPI2    bool
 		registerAPISrv2 bool
 		registerWeb     bool
+		registerKV      bool
+		registerKVMulti bool
```
wilkermichael, line 113: Actually I can just get rid of these two conditions and just always add the KV, then it wouldn't create so many extra paths in the test code, and the checking should still work... it would even make sure that `recurse true` is the only one that picks up the services in the `registerKVMulti `block

templates/tftmpl/monitor_services.go
```diff
@@ -71,6 +75,10 @@ func (m ServicesMonitor) appendVariable(io.Writer) error {
 	return nil
 }
 
+func (m ServicesMonitor) SourceIncludesVariable() bool {
```
wilkermichael, line 78: I'll add a comment here:

`SourceIncludesVariable returns whether or not to include the source variables in the template. For the case of a monitor, this always returns true and must be overridden to return based on other conditions.`

templates/tftmpl/monitor_catalog_services.go
```diff
@@ -27,6 +27,10 @@ func (m CatalogServicesMonitor) ServicesAppended() bool {
 	return false
 }
 
+func (m CatalogServicesMonitor) SourceIncludesVariable() bool {
```
wilkermichael, line 30: Will add comment:

`SourceIncludesVariable returns whether or not to include the source variables in the template. For the case of a monitor, this always returns true and must be overridden to return based on other conditions.`

templates/tftmpl/monitor_consul_kv.go
```diff
@@ -28,6 +28,10 @@ func (m ConsulKVMonitor) ServicesAppended() bool {
 	return false
 }
 
+func (m ConsulKVMonitor) SourceIncludesVariable() bool {
```
wilkermichael, line 31: Will add comment:

`SourceIncludesVariable returns whether or not to include the source variables in the template. For the case of a monitor, this always returns true and must be overridden to return based on other conditions.`

templates/tftmpl/variables.go
```diff
@@ -67,6 +67,12 @@ func newVariablesTF(w io.Writer, filename string, input *RootModuleInputData) er
 		}
 	}
 
+	if input.SourceInput != nil && input.SourceInput.SourceIncludesVariable() {
```
wilkermichael, line 70: SourceIncludesVariable() should always be true here so maybe I should just remove it from the check? Could also do a panic here if SourceIncludes() variable isn't true... this way we could catch programmer errors.

e2e/condition_schedule_test.go
```diff
@@ -128,11 +176,42 @@ func TestCondition_Schedule_Basic(t *testing.T) {
...
 			checkScheduledRun(t, taskName, registerTime, taskSchedule, port)
 
-			// confirm resources created
+			// confirm service resources created
 			resourcesPath := filepath.Join(tempDir, taskName, resourcesDir)
 			testutils.CheckFile(t, true, resourcesPath, "api-1.txt")
 			testutils.CheckFile(t, true, resourcesPath, "web-1.txt")
 
+			// Check KV events
+			if tc.isConsulKV {
```
wilkermichael, line 185: I don't "love" having this branching path here, but I figured that this can be resolved when someone takes on the e2e refactoring we've talked about.

wilkermichael:
@mkam I just left some comments for myself to fix, and added some context around some decisions I made, all done now

driver/terraform.go
mkam: I'm thinking it's fine to error out now because if setting the environment variables fails, then the terraform-exec won't behave as expected.

driver/terraform.go
```diff
@@ -547,6 +541,26 @@ func (tf *Terraform) initTaskTemplate() error {
 	return nil
 }
 
+func (tf *Terraform) setNotifier(tmpl templates.Template, serviceCount int) {
```
mkam, line 544: I like this new separate method for setting the notifier since we have so many cases now!

driver/terraform.go
```diff
@@ -547,6 +541,26 @@ func (tf *Terraform) initTaskTemplate() error {
...
+	switch tf.task.Condition().(type) {
+	case *config.CatalogServicesConditionConfig:
+		tf.template = notifier.NewCatalogServicesRegistration(tmpl, serviceCount)
+	case *config.ConsulKVConditionConfig:
+		tf.template = notifier.NewConsulKV(tmpl, serviceCount)
+	case *config.ScheduleConditionConfig:
+		// If a ConsulKVSourceInputConfig is specified, then we need to add
+		// to the number of services passed to the notifier, since consul-kv adds a dependency
+		serviceCountAddition := 0
```
mkam, line 553: This isn't really a services count at this point, it's more a dependencies count. I suggest changing the name of this variable here, and also in the `NewSuppressNotification` method so it's more accurate.

templates/tftmpl/variables.go
mkam: Hm, I don't have a lot of experience in the conventions around panics in go, but I think removing the check would be good if it is always going to be true!

mkam (approved):
Looks good to me! Just left a few minor comments!

templates/tftmpl/integration_test.go
wilkermichael: never mind, can't do that right now because of the spacing issue. Leaving this one as is

templates/tftmpl/variables.go
wilkermichael: I was thinking something like this:

```
	if input.SourceInput != nil {
		if input.SourceInput.SourceIncludesVariable() {
			if err = input.SourceInput.appendVariable(w); err != nil {
				return err
			}
		} else {
			// SourceIncludesVariable should always be true for a non-nil source input
			// if we got here, it means that a source_input's SourceIncludesVariable() function
			// was not setup correctly
			panic("SourceIncludesVariable is false for a non-nil source input")
		}
	}
```

I see panics as basically the equivalent of an assert, so for catching programmer errors.

e2e/condition_schedule_test.go
```diff
@@ -45,27 +46,76 @@ func TestCondition_Schedule_Basic(t *testing.T) {
...
+			tempDir:       "schedule_consulKV_recurse",
+			isConsulKV:    true,
+			isRecurse:     true,
 		},
 	}
 
 	for _, tc := range testcases {
+		tc := tc // rebind tc into this lexical scope for parallel use
 		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel() // Run table tests in parallel as they can take a lot of time
```
mkam, line 118: Great to see that we can run these cases in parallel!

PR #405, opened 2021-10-01
Weekly Consul compat tests

findkim:
Runs compatibility tests with Consul in parallel and adds it to weekly CI jobs.
* Fix tests looking for artifacts in old paths (missing task subdir)
* Refactor tests to run in parallel by version
  * Reworked temp directories to not overlap across versions

Tests passing https://app.circleci.com/pipelines/github/hashicorp/consul-terraform-sync/1858/workflows/7185eedf-177e-4c37-9b61-0bb8a4690876/jobs/3655

TODO follow up and add Consul 1.11 when released.

I tried using hc-install library to fetch the latest version within minor releases, but it requires adding an internal dependency which would break builds for OSS users locally.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,6 +79,28 @@ jobs:
           paths:
             - /go/pkg/mod
 
+  consul_compatibility_tests:
+    environment:
+      GO111MODULE: "on"
+      GO_VERSION: "1.16"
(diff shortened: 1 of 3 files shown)


mkam (approved):
Excited to have these running more regularly! 🎊

PR #433, opened 2021-10-11
Update CRT configuration to build on release branches

mkam:
Regex support is not supported yet, so listing n-2 release branches.

### DIFF
--- a/.release/ci.hcl
+++ b/.release/ci.hcl
@@ -9,7 +9,12 @@ project "consul-terraform-sync" {
   github {
     organization = "hashicorp"
     repository = "consul-terraform-sync"
-    release_branches = ["main"]
+    release_branches = [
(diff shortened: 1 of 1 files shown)


sarahethompson:
> In hindsight, I included `release/0.2.x` and `release/0.3.x`, but if the CRT code isn't on those branches, I'm not sure that we'd be able to use CRT to release these older versions.

Yeah, we would need to backport the CRT related things to those branches? Will releases be going out soon from those branches? We can open some PRs tomorrow to get things enabled!

mkam:
@sarahethompson We don't have any plans right now to release to those branches, I had mostly included them in case we require a patch version in the future. I'd say it would be nice to have it in those branches so that we're ready if a patch is needed, but it's not urgent!

PR #446, opened 2021-10-13
Fix test issues when using datacenter filtering

mkam:
The consul service was not immediately registered in dc2, so the tests
were occasionally failing.


mkam:
Ran this multiple times and a different test failed. 🤦‍♀️  But neither of these tests failed, and I also tested locally by running the cases in for loops!

PR #472, opened 2021-10-29
add ACL configuration

wilkermichael:
To configure CTS to support ACL tokens I've added a new ACL block. This first version of this block contains two fields:

```hcl
acl {
  enabled         = true
  bootstrap_token = "+aHuWFH0bNzERaJpwdAPteD5EYzEQSWWNUxFsiVWt4ADIbHDU95ytJoYfHd/M22Q"
}
```

```Go
// ACLConfig is the configuration for an Access Control List (ACL).
type ACLConfig struct {
	// Boolean if ACLs are enabled or not. True if enabled, false otherwise.
	Enabled *bool `mapstructure:"enabled"`

	// The encrypted bootstrap token to be used by CTS. If this is configured then
	// CTS will not need to be bootstrapped on restart
	BootstrapToken *string `mapstructure:"bootstrap_token"`
}
```

The intent for the BootstrapToken is that if this field is present and non-empty, this token will be used on restart, requiring no further bootstrapping. Enabled by default is currently set to `false` this will change in the future to `true` once this is officially supported by CTS.

EDIT***

Remodelled the block to be closer to consul, bootstrap_token is now called `master` and the block looks like this instead:

```hcl
acl {
  enabled = true

  tokens = {
    master = "da666809-98ca-0e94-a99c-893c4bf5f9eb"
  }
}
```

```Go
// ACLConfig is the configuration for an Access Control List (ACL).
type ACLConfig struct {
	// Boolean if ACLs are enabled or not. True if enabled, false otherwise.
	Enabled *bool `mapstructure:"enabled"`

	// The list of tokens supported for this CTS instance
	Tokens *TokensConfig `mapstructure:"tokens"`
}
```

### DIFF
--- a/config/tokens.go
+++ b/config/tokens.go
@@ -1,0 +1,88 @@
+package config
+
+import (
+	"fmt"
+)
+
+// TokensConfig is the configuration for the CTS tokens.
+type TokensConfig struct {
+
+	// Master allows operators to bootstrap the ACL system with a token Secret ID that is well-known.
+	Root *string `mapstructure:"root"`
+}
+
+// Copy returns a deep copy of this configuration.
+func (c *TokensConfig) Copy() *TokensConfig {
+	if c == nil {
+		return nil
+	}
+
+	var o TokensConfig
+	o.Root = StringCopy(c.Root)
+
+	return &o
+}
+
+// Merge combines all values in this configuration with the values in the other
+// configuration, with values in the other configuration taking precedence.
+// Maps and slices are merged, most other values are overwritten. Complex
+// structs define their own merge functionality.
+func (c *TokensConfig) Merge(o *TokensConfig) *TokensConfig {
+	if c == nil {
+		if o == nil {
+			return nil
+		}
+		return o.Copy()
+	}
+
+	if o == nil {
+		return c.Copy()
+	}
+
+	r := c.Copy()
+
+	if o.Root != nil {
+		r.Root = StringCopy(o.Root)
+	}
+
+	return r
+}
+
+// Finalize ensures there no nil pointers.
+func (c *TokensConfig) Finalize() {
+	if c == nil {
+		return
+	}
+
+	if c.Root == nil {
+		c.Root = String("")
+	}
+}
+
+// Validate validates the values and required options. This method is recommended
+// to run after Finalize() to ensure the configuration is safe to proceed.
+func (c *TokensConfig) Validate() error {
+	if c == nil {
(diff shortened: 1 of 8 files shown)


lornasong (approved):
looks good to me!

config/tokens.go
```diff
@@ -0,0 +1,88 @@
+package config
+
+import (
+	"fmt"
+)
+
+// TokensConfig is the configuration for the CTS tokens.
+type TokensConfig struct {
+
+	// Master allows operators to bootstrap the ACL system with a token Secret ID that is well-known.
```
lornasong, line 10: "Master" => "Root"

config/tokens.go
```diff
@@ -0,0 +1,88 @@
...
+// Sensitive information is redacted.
+func (c *TokensConfig) GoString() string {
+	if c == nil {
+		return "(*TokensConfig)(nil)"
+	}
+
+	return fmt.Sprintf("&TokensConfig{"+
+		"Root:%s, "+
+		"}",
+		StringVal(c.Root),
```
lornasong, line 86: Wanted to double-check, in a comment earlier about `Root` it sounds like the root token value is "well-known" and therefore not sensitive? Similar to an earlier comment about token GoString() - do we think that it'll have sensitive info down the line where it may be better to leave this GoString redacted?

config/tokens_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+func TestTokensConfig_Finalize(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name string
+		i    *TokensConfig
+		r    *TokensConfig
+	}{
+		{
```
lornasong, line 116: Also suggestion here to add nil test case

config/tokens_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+func TestTokensConfig_Validate(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name      string
+		expectErr bool
+		c         *TokensConfig
+	}{
+		{
```
lornasong, line 148: Also suggestion here to add nil test case

lornasong (approved):
makes sense to me! re-reviewed and thought of a couple new things that are left in comments - non-blocking, primarily want to get your insight on outputting token information.

for posterity - we also chatted off-line that this PR would be merged depending on how we prioritize authorization

wilkermichael:
Closing for the time being

PR #479, opened 2021-11-02
add dumb-init package to dockerfile

alvin-huang:
This package is needed in the docker entrypoint.


findkim (approved):
Thanks for figuring out my cryptic error!
> standard_init_linux.go:228: exec user process caused: no such file or directory

eikenb:
Just an informative aside to this... dumb-init is no longer needed for the zombie process reaping as docker now has it built in. You can just use the standard `#!/bin/sh` to run the entrypoint script and not worry about installing/using dumb-init.

PR #484, opened 2021-11-02
Adding sign-linux-rpms event to ci.hcl

sarahethompson:


kpenfound (approved):

findkim (approved):
Yay. Thanks for adding to linux rpms to CTS. This is more of async ask, but if you could also share here/slack an example installation with this addition?

We're planning to release tomorrow and so we will include this in the following release.

sarahethompson:
> Yay. Thanks for adding to linux rpms to CTS. This is more of async ask, but if you could also share here/slack an example installation with this addition?
> 
> We're planning to release tomorrow and so we will include this in the following release.

Hi @findkim! This is to fix this issue properly - https://hashicorp.slack.com/archives/C026W707YHJ/p1634748596008800 so that we are signing the rpms during the CRT pipeline and so that we don't have to do any manual fixes after the fact! Let us know if you need any more info cc @kpenfound

findkim:
Oh! it's been a long two weeks and had forgotten about this. Thanks for the reminder Sarah :)

findkim:
In that case, bring it on over to 0.4.1 for tomorrow!

Issue #466, opened 2021-10-27
Support TLS and Mutual TLS for CTS API


mkam:
#### Description
Add support for configuring TLS and mTLS to allow HTTPS connections to the CTS API. CTS currently only supports TLS for communicating with external services like Consul and Terraform Cloud. Also add support for the CTS CLI to connect to a TLS or mTLS enabled CTS API.

#### Use Cases
As the CTS API expands, users need a way to securely connect to it. Some planned features that will depend on this are:
- Authentication on CTS API
- Task Lifecycle API
- High Availability

PR #488, opened 2021-11-03
Support TLS on CTS API

mkam:
Adds support for enabling TLS on the API on the CTS API using a new root-level TLS configuration block.
```
tls {
  enabled = true
  cert = "/path/to/cert.pem"
  key = "/path/to/key.pem" 
}
```
Does not implement mutual TLS or the client/CLI related changes.

For reference, I generated the new test certificate and key with the following command:
```
openssl req -nodes -x509 -addext "subjectAltName = DNS:localhost" -subj '/CN=localhost' -keyout localhost_key.pem -out localhost_cert.pem -sha256 -days 3650 -newkey rsa:4096
```

Part of https://github.com/hashicorp/consul-terraform-sync/issues/466


api/api.go
```diff
@@ -124,9 +127,16 @@ func (api *API) Serve(ctx context.Context) error {
...
+	if config.BoolVal(api.tls.Enabled) {
+		if err := api.srv.ListenAndServeTLS(*api.tls.Cert, *api.tls.Key); err != nil && err != http.ErrServerClosed {
+			logger.Error("error serving api", "port", api.port, "error", err)
+			return err
+		}
+	} else {
+		if err := api.srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			logger.Error("error serving api", "port", api.port, "error", err)
+			return err
+		}
```
findkim, line 139:
```suggestion
	var err error
	if config.BoolVal(api.tls.Enabled) {
		err = api.srv.ListenAndServeTLS(*api.tls.Cert, *api.tls.Key)
	} else {
		err = api.srv.ListenAndServe()
	}
	if err != nil && err != http.ErrServerClosed {
		logger.Error("error serving api", "port", api.port, "error", err)
		return err
```

minor suggestion to reduce the nested ifs and identical error handling

api/api.go
```diff
@@ -61,10 +62,11 @@ type API struct {
 	port    int
 	version string
 	srv     *http.Server
+	tls     *config.TLSConfig
 }
 
 // NewAPI create a new API object
-func NewAPI(store *event.Store, drivers *driver.Drivers, port int) *API {
+func NewAPI(store *event.Store, drivers *driver.Drivers, port int, tls *config.TLSConfig) *API {
```
findkim, line 69: Nit: 3 parameters is a load, and now with 4 it might be a good time to make a dedicated config struct for the API to pass in. Where can configure the options optionally and the function can handle default values.

Like

```
func NewAPI(cfg myStruct) *API {
  var api &API{}
  if cfg.Store == nil {
    api.Store = event.NewStore()
  }
}

File: api/api_test.go [hunk 3]
```diff
@@ -115,6 +119,53 @@ func TestServe_context_cancel(t *testing.T) {
...
+	client := &http.Client{Transport: &http.Transport{
+		TLSClientConfig: &tls.Config{
+			RootCAs: caCertPool,
+		},
+	}}
+
+	// Make request to HTTPS endpoint
+	u := fmt.Sprintf("https://localhost:%d/%s/status",
+		port, defaultAPIVersion)
+	resp, err := client.Get(u)
```
findkim, line 164: Can you add coverage to test a request from a client w/o certs to a sever with it enabled?

findkim (approved):

api/api.go
mkam: Makes sense, will put these options in a new struct!

findkim (approved):

PR #490, opened 2021-11-03
Add tls cli client

wilkermichael:
Adds TLS flags to support TLS when using the CTS CLI

- TLS flags are passed via command line flags, or by environment variable
- command line flag values take precedence over corresponding environment variable values
- meta.go is responsible for handling flags set
- client.go handles environment variables and defaults required for creating the client

Example using flags:
```
consul-terraform-sync task enable \
    -http-addr="https://localhost:8501" \
    -ca-file="/Users/mwilkerson/dev/michael-cts-example/tls_consul.d/consul-agent-ca.pem" \
    -client-cert="/Users/mwilkerson/dev/michael-cts-example/tls_consul.d/dc1-server-consul-0.pem" \
    -client-key="/Users/mwilkerson/dev/michael-cts-example/tls_consul.d/dc1-server-consul-0-key.pem" \ 
    example-task
```

Example using some environment variables (same effect)
```
export CTS_CACERT="/Users/mwilkerson/dev/michael-cts-example/tls_consul.d/consul-agent-ca.pem" 
export CTS_CLIENT_CERT="/Users/mwilkerson/dev/michael-cts-example/tls_consul.d/dc1-server-consul-0.pem"
export CTS_CLIENT_KEY="/Users/mwilkerson/dev/michael-cts-example/tls_consul.d/dc1-server-consul-0-key.pem"

consul-terraform-sync task enable -http-addr="https://localhost:8501" example-task
```

*Note* the client won't work until the CTS server handles TLS, as of right now this hasn't been completed

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -1,15 +1,39 @@
 package api
 
 import (
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"net/http"
 	"net/url"
+	"os"
+	"strconv"
 	"strings"
 	"time"
+
+	"github.com/hashicorp/consul-terraform-sync/config"
+	"github.com/hashicorp/go-rootcerts"
 )
 
 //go:generate mockery --name=httpClient  --structname=HttpClient --output=../mocks/api
+
+const (
+	httpScheme  = "http"
+	httpsScheme = "https"
+
+	DefaultAddress   = "http://localhost:8558"
+	DefaultSSLVerify = true
+
+	// Environment Variables
+	EnvAddress = "CTS_ADDRESS" // The address of the CTS daemon, supports http or https by specifying as part of the address (e.g. https://localhost:8558)
+
+	// TLS Environment Variables
+	EnvTLSCACert     = "CTS_CACERT"      // Path to a directory of CA certificates to use for TLS when communicating with Consul-Terraform-Sync
+	EnvTLSCAPath     = "CTS_CAPATH"      // Path to a CA file to use for TLS when communicating with Consul-Terraform-Sync
+	EnvTLSClientCert = "CTS_CLIENT_CERT" // Path to a client cert file to use for TLS when verify_incoming is enabled
+	EnvTLSClientKey  = "CTS_CLIENT_KEY"  // Path to a client key file to use for TLS when verify_incoming is enabled
+	EnvTLSSSLVerify  = "CTS_SSL_VERIFY"  // Boolean to verify SSL or not. Set to true to verify SSL. Default is true
+)
 
 // httpClient describes the interface for the client to make http calls
 type httpClient interface {
@@ -18,7 +42,8 @@ type httpClient interface {
 
 // Client to make api requests
 type Client struct {
-	port    int
+	port    int // remain for backwards compatibility but prefer addr
+	addr    string
 	version string
 	scheme  string
 	http    httpClient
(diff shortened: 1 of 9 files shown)


command/cli.go
```diff
@@ -98,7 +98,7 @@ func (cli *CLI) Run(args []string) int {
 
 	// Development only flags. Not printed with -h, -help
 	f.StringVar(&clientType, "client-type", "", "Use only when developing"+
-		"consul-terraform-sync binary. Defaults to Terraform client if empty or"+
+		" consul-terraform-sync binary. Defaults to Terraform client if empty or"+
 		"unknown value. Values can also be 'development' or 'test'.")
```
findkim, line 102:
```suggestion
	f.StringVar(&clientType, "client-type", "", "Use only when developing "+
		"consul-terraform-sync binary. Defaults to Terraform client if empty or "+
		"unknown value. Values can also be 'development' or 'test'.")
```

Believe there is also a missing space between `orunknown`

command/meta.go
```diff
@@ -6,8 +6,9 @@ import (
 	"io/ioutil"
 	"strings"
 
-	"github.com/hashicorp/consul-terraform-sync/api"
 	"github.com/hashicorp/consul-terraform-sync/config"
+
```
findkim, line 10: 👀

command/meta.go
```diff
@@ -17,26 +18,61 @@ const width = uint(78)
...
+	flagSSLVerify  = "ssl-verify"
+)
+
+func (m *meta) defaultFlagSet(name string) *flag.FlagSet {
+	m.flags = flag.NewFlagSet(name, flag.ContinueOnError)
+
+	// Values provide both default values, and documentation for the default value when -help is used
+	m.port = m.flags.Int(flagPort, config.DefaultPort,
+		fmt.Sprintf("The port to use for the Consul Terraform Sync API server, it is preferred to use the %s field instead", flagHTTPAddr))
+	m.addr = m.flags.String(flagHTTPAddr, api.DefaultAddress, "The address to use for the Consul Terraform Sync API server")
```
findkim, line 57: Which flag would someone use with an address with the https scheme?

wilkermichael: They just need to include "https" in the address to enable the https scheme, this is what Consul does I believe, I pulled the scheme parsing code from there. So for example: https://localhost:8558 would set the https scheme, we probably would want to recommend a port for https as well, but that's more a server config thing.

wilkermichael: https://github.com/hashicorp/consul-terraform-sync/blob/4f234c37222bb86da9b165a9db19dff141500bee/api/client.go#L394

api/client.go
```diff
@@ -1,45 +1,176 @@
...
 	"fmt"
 	"net/http"
 	"net/url"
+	"os"
+	"strconv"
 	"strings"
 	"time"
+
+	"github.com/hashicorp/consul-terraform-sync/config"
+
```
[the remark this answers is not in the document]
wilkermichael: I'll watch for this again, I can't reproduce the problem to test if I fixed it, I tried

1) deleting both hashicorp packages, and it adds them back without the whitespace
2) deleting the config package, using the automation. deleting the go-rootcerts package and running the automation. In both cases it groups without whitespace :(

Hopefully I can catch the scenario next time.

command/meta.go
findkim: Yeah my concern is the clarity in the flag name and description with its actual capabilities. Can you update the description to include HTTPS?

This is how Consul CLI has it documented
> The address and port of the Consul HTTP agent. The value can be
     an IP address or DNS address, but it must also include the port.
     This can also be specified via the CONSUL_HTTP_ADDR environment
     variable. The default value is http://127.0.0.1:8500. The scheme
     can also be set to HTTPS by setting the environment variable
     CONSUL_HTTP_SSL=true.

wilkermichael: ah I see, sure I'll add this

wilkermichael: @findkim added!

findkim (approved):
🎉

command/meta.go
findkim: Thank you!

PR #491, opened 2021-11-08
Use base version for GitDescribe for app to build version string

findkim:
`{{ needs.get-product-version.outputs.product-version }}` uses [build-scripts/version.sh](https://github.com/hashicorp/consul-terraform-sync/blob/main/build-scripts/version.sh) to generate the version that is then set to `GitDescribe` as a linker flag. This results in redundant prerelease and metadata added to the version during runtime.

```
% ./consul-terraform-sync -version
consul-terraform-sync v0.5.0+dev+dev (xxxx)
Compatible with Terraform >= 0.13.0, < 1.1.0
```


.release/docker/Makefile
```diff
@@ -20,4 +20,4 @@ docker:
 	echo "Building image for $(PROJECT_NAME):$(GITHUB_SHA)-$(VERSION)-$(ARCH)"
 	docker buildx create --use
 	docker buildx build --load --build-arg VERSION=$(VERSION) --build-arg NAME=$(PROJECT_NAME) --platform=linux/$(ARCH)$(GOARM_DOCKER) -t $(PROJECT_NAME):$(GITHUB_SHA)-$(VERSION)-$(ARCH) .
-	docker save --output $(PROJECT_NAME)_$(VERSION)_docker_linux_$(ARCH).tar $(PROJECT_NAME):$(GITHUB_SHA)-$(VERSION)-$(ARCH)
\ No newline at end of file
+	docker save --output $(PROJECT_NAME)_$(VERSION)_docker_linux_$(ARCH).tar $(PROJECT_NAME):$(GITHUB_SHA)-$(VERSION)-$(ARCH)
```
wilkermichael, line 23: Since it's just the addition of a newline, should we just revert this change?

wilkermichael (approved):
This looks good to me, just to check my understanding:

You set base=1, then in version.sh it will check if this field is null. Since it isn't it will set git describe to be the base version, so when we check the version we should get:

```
% ./consul-terraform-sync -version
consul-terraform-sync v0.5.0 
Compatible with Terraform >= 0.13.0, < 1.1.0
```

?

.release/docker/Makefile
findkim: To be honest, I don't like the red circle that appears on GitHub for files w/o new lines 😅

But more seriously, your question got me curious about the context of the new line at the EOF, and why GitHub displays it. It is a POSIX standard for a text file which contain lines, and lines include the terminating newline character. So this one is considered an invalid line to POSIX purists.

[Text file](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_403):
> A file that contains characters organized into zero or more lines. The lines do not contain NUL characters and none can exceed {LINE_MAX} bytes in length, including the <newline> character. Although POSIX.1-2008 does not distinguish between text files and binary files (see the ISO C standard), many utilities only produce predictable or meaningful output when operating on text files. The standard utilities that have such restrictions always specify "text files" in their STDIN or INPUT FILES sections.

[Line](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206)
> A sequence of zero or more non- <newline> characters plus a terminating <newline> character.

findkim: If you don't mind, I'd like to keep the newline to keep consistency with the rest of the files in the repo.

findkim:
@wilkermichael yep! The `version.sh` command is used for everything else in the release process so I tried to make a minimal change for the GitDescribe use-case.

PR #492, opened 2021-11-08
Add tls tests

wilkermichael:


e2e/command_test.go
```diff
@@ -147,7 +252,144 @@ func TestE2E_EnableTaskCommand(t *testing.T) {
 	}
 }
 
-// TestE2E_DisableTaskCommand tests the CLI to disable an enabled task. This
+// TestE2E_EnableTaskCommand tests the task Enable CLI and confirms the expected
+// output and state given different paths. This test starts up a local Consul server
+// and runs CTS with TLS enabled, with a disabled task.
+func TestE2E_EnableTaskCommand_TLS(t *testing.T) {
```
mkam, line 258: Totally down for testing TLS on our existing enable and disable commands, but wanted to start thinking about what our long term strategy is for testing TLS with our CLI. We'll be eventually adding more CLI commands, do we want to always be having test versions for both TLS enabled and disabled?

Doesn't have to be in this PR, but we could have a separate test or test suite that checks all these combinations on all our commands just to see that the CLI is connecting and not necessarily checking the behavior of the command. Then we wouldn't have to always include the TLS test versions for the functionality tests.

e2e/command_test.go
```diff
@@ -82,7 +92,99 @@ func TestE2E_MetaCommandErrors(t *testing.T) {
...
+// starts up a local Consul server and runs CTS with TLS in dev mode.
+func TestE2E_TLSErrors(t *testing.T) {
+	t.Parallel()
+
+	srv := newTestConsulServer(t)
+	defer srv.Stop()
+
+	tempDir := fmt.Sprintf("%s%s", tempDirPrefix, "command_tls_errors")
+
+	tlsc := tlsConfig{
```
mkam, line 108: You could make a defaultTLSConfig method that returns this value. I'd also suggest having no caCert set for the default CTS TLS config, and when we add mutualTLS, we could add defaultMTLSConfig that has caCert.

mkam (approved):

wilkermichael:
Adding check for https scheme used to connect with CTS using http

e2e/command_test.go
wilkermichael: Talked to Melissa about this, I agree that TLS should just be checking the connection was successful. I will refactor this at a later date

e2e/config.go
```diff
@@ -48,6 +48,17 @@ consul {
 `, consul.HTTPSAddr, consul.Config.CertFile))
 }
 
+func (c hclConfig) appendTLSBlock(clientCert, clientKey, caCert string) hclConfig {
```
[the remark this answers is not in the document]
wilkermichael: I changed this to build the block based on the TLSConfig provided

PR #495, opened 2021-11-09
refactor command tls e2e tests to just test for connection success

wilkermichael:
I added a new test, **TestE2E_CommandTLS**, for commands using TLS, this test **does**:

- verify that there were no errors
- verify some contained output to assure that a connection has been made

This test **does not** test any functionality outside of the initial connection.

I also removed two existing tests that were testing TLS connections alongside command functionality. Command functionality is already tested in other tests, so this was deemed to be redundant.

I had to split this into two commits, if you look at the unified commits it thinks I changed one of the tests rather than adding a new test and removing two tests.


mkam (approved):
Looks good to me, the commit split was super helpful, thanks for breaking it up like that!

PR #499, opened 2021-11-10
Log warning if both ca_cert and ca_path are configured

mkam:
The go-rootcerts library has ca_cert take precedence over ca_path, so
calling out this behavior in a warning. Also fixes the mTLS check to look
for either ca_path or ca_cert, which was the original intent.

The diff isn't great to look at, but I moved one TLS check up further, modified the mTLS check to care about the CA options and not cert/key, and added the check that warns if both CA options are configured.

Warning will look like the following:
```
$ consul-terraform-sync --config-file=config.hcl
2021-11-10T11:57:44.155-0600 [WARN]  config: ca_cert and ca_path are both configured, but only ca_cert will be used
2021-11-10T11:57:44.156-0600 [INFO]  cli: v0.4.1 (5ff7739)
2021-11-10T11:57:44.156-0600 [INFO]  cli: running controller in daemon mode
```

### DIFF
--- a/config/tls_cts.go
+++ b/config/tls_cts.go
@@ -3,6 +3,8 @@ package config
 import (
 	"crypto/tls"
 	"fmt"
+
+	"github.com/hashicorp/consul-terraform-sync/logging"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


config/tls_cts.go
```diff
@@ -144,6 +146,10 @@ func (c *CTSTLSConfig) Validate() error {
 	}
 
 	// TLS validation
+	if BoolVal(c.Enabled) && ((StringVal(c.Key) == "") || (StringVal(c.Cert) == "")) {
+		return fmt.Errorf("key and cert are required if TLS is enabled")
```
lornasong, line 150: nit suggestion - just was realizing that other configs also have tls related fields - it might be helpful in this error message (also below and the warning log) to add some sort of context that its specifically for the cts-tls config.

lornasong (commented):
looks good to me!

lornasong (approved):

config/tls_cts.go
mkam: That's a good point, I can update these errors to include that context!

PR #500, opened 2021-11-10
Use standard logger for any server errors

mkam:
Added to address TLS handshake errors when the client is misconfigured,
but will handle any errors the server code logs.

Example of previous error format:
```
2021/11/09 12:30:44 http: TLS handshake error from [::1]:54789: tls: failed to verify client certificate: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "localhost")
```

Updated version:

```
2021-11-12T13:39:06.898-0600 [WARN]  api: http: TLS handshake error from [::1]:62111: tls: failed to verify client certificate: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "localhost")
```

~~I have it set to infer the log level, but the log level selection looks for a prefix (like [ERROR] or [DEBUG]) to determine what level. The server code doesn't have that prefix, so it defaults to INFO.~~

~~INFO seems reasonable to me because even though it's an error, it's the fault of the client and not the server. We're also logging successful API requests at the INFO level. The alternative is that we force a log level for any log that is printed by the server code.~~

After discussing the appropriate level, went with WARN instead of the default INFO as there are some scenarios in which action needs to be taken on the server side.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -14,6 +14,7 @@ import (
 	"github.com/hashicorp/consul-terraform-sync/driver"
 	"github.com/hashicorp/consul-terraform-sync/event"
 	"github.com/hashicorp/consul-terraform-sync/logging"
+	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-rootcerts"
 )
 
@@ -77,6 +78,7 @@ type APIConfig struct {
 // NewAPI create a new API object
 func NewAPI(conf *APIConfig) (*API, error) {
 	mux := http.NewServeMux()
+	logger := logging.Global().Named(logSystemName)
 
 	api := &API{
 		port:    conf.Port,
@@ -119,7 +121,6 @@ func NewAPI(conf *APIConfig) (*API, error) {
 			CAPath: config.StringVal(api.tls.CAPath),
 		})
 		if err != nil {
-			logger := logging.Global().Named(logSystemName)
 			logger.Error("error loading TLS configs for api server", "error", err)
 			return nil, err
 		}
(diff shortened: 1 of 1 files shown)


eikenb:
INFO level for errors, in my mind, has basically 1 requirement... That is that it doesn't impact the functionality of the process, that it can be safely ignored. If it can be ignored then INFO or lower is appropriate. If it can't be ignored then it needs to be WARN or higher.

wilkermichael:
> INFO level for errors, in my mind, has basically 1 requirement... That is that it doesn't impact the functionality of the process, that it can be safely ignored. If it can be ignored then INFO or lower is appropriate. If it can't be ignored then it needs to be WARN or higher.

I agree with John in this case, since this a configuration for logs from the http.ErrorLog I think it would be safe to assume that the level is an error or warning. The ErrorLog comes with the following comment:

```
	// ErrorLog specifies an optional logger for errors accepting
	// connections, unexpected behavior from handlers, and
	// underlying FileSystem errors.
	// If nil, logging is done via the log package's standard logger.
```

It says errors or unexpected behaviour... so maybe "WARN" would be a better log level?

api/api.go
```diff
@@ -134,6 +135,9 @@ func NewAPI(conf *APIConfig) (*API, error) {
 		IdleTimeout:  time.Second * 60,
 		Handler:      mux,
 		TLSConfig:    t,
+		ErrorLog: logger.StandardLogger(&hclog.StandardLoggerOptions{
```
wilkermichael, line 138: See my comment replying to John's comment about the log level

wilkermichael (approved):
Thanks for doing this Melissa! This is a great fix, thanks for digging into this.

api/api.go
mkam: Yeah, based on that description, it sounds like there can be errors that are logged but do need intervention on the server side. I'll update it to always log with WARN!

mkam:
```
2021-11-12T13:36:59.644-0600 [WARN]  api: http: TLS handshake error from [::1]:62024: EOF
```

PR #504, opened 2021-11-12
Added wrapper message around EOF error returned by CTS client when co…

wilkermichael:
#### Background
The purpose of this PR is to have more meaningful TLS errors returned from our CTS CLI when connecting using an http scheme with a CTS instance using https.

From research:

- **GET/PUT** receives a nice response from the CTS server: `Client sent an HTTP request to an HTTPS server.`
- **PATCH/DELETE/POST** however get a nil response and an EOF error.
- These errors do not happen when connecting using an https client to an http server. In that case we get the nice `Client sent an HTTPS request to an HTTP server.`

Investigating how Consul handles things, it looks like they have similar behaviour. It also looks like they just return the EOF response as is. For example:
```
consul kv put -http-addr="http://localhost:8501" key-value "blue"
Error! Failed writing data: Unexpected response code: 400 (Client sent an HTTP request to an HTTPS server.)

consul kv delete -http-addr="http://localhost:8501" "blue"
Error deleting key blue: Delete "http://localhost:8501/v1/kv/blue": EOF
```

#### Solution
EOF seems to be a common error returned in this scenario, and there isn't much we can do about what is returned from the server. What we can improve however is the information the CLI provides. Therefore, I've added a check that `if strings.Contains(err.Error(), "EOF") && scheme == api.HttpScheme` then a more detailed message will be returned from the CLI:

```
==> Error: unable to disable 'example-task'
    Patch "http://localhost:8558/v1/tasks/example-task": EOF. Scheme http was
used, this error can be caused by using http to connect to an https CTS
instance, consider using https scheme instead
```

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -18,8 +18,8 @@ import (
 //go:generate mockery --name=httpClient  --structname=HttpClient --output=../mocks/api
 
 const (
-	httpScheme  = "http"
-	httpsScheme = "https"
+	HTTPScheme  = "http"
+	HTTPSScheme = "https"
 
 	DefaultAddress   = "http://localhost:8558"
 	DefaultSSLVerify = true
@@ -172,12 +172,19 @@ func setupTLSConfig(c *ClientConfig) (*tls.Config, error) {
 	return tlsClientConfig, nil
 }
 
+// Port returns the port being used by the client
 func (c *Client) Port() int {
 	return c.port
 }
 
+// FullAddress returns the client address including the scheme. eg. http://localhost:8558
 func (c *Client) FullAddress() string {
 	return fmt.Sprintf("%s://%s", c.scheme, c.addr)
+}
+
+// Scheme returns the scheme being used by the client
+func (c *Client) Scheme() string {
+	return fmt.Sprintf(c.scheme)
 }
 
 // WaitForAPI polls the /v1/status endpoint to check when the CTS API is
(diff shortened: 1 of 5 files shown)


api/client.go
```diff
@@ -18,8 +18,8 @@ import (
 //go:generate mockery --name=httpClient  --structname=HttpClient --output=../mocks/api
 
 const (
-	httpScheme  = "http"
-	httpsScheme = "https"
+	HttpScheme  = "http"
+	HttpsScheme = "https"
```
findkim, line 22:
```suggestion
	HTTPScheme  = "http"
	HTTPSScheme = "https"
```

Can you update the constants to follow golang idiomatic casing for initialisms?

command/task_enable.go
```diff
@@ -105,6 +105,8 @@ func (c *taskEnableCommand) Run(args []string) int {
 	}, &api.QueryParam{Run: driver.RunOptionInspect})
 	if err != nil {
 		c.UI.Error(fmt.Sprintf("Error: unable to generate plan for '%s'", taskName))
+
+		err = processEOFError(client.Scheme(), err)
```
findkim, line 109: can you make the spacing consistent between here and `command/task_disable.go`?

command/task_disable.go
```diff
@@ -101,3 +102,13 @@ func (c *taskDisableCommand) Run(args []string) int {
 
 	return ExitCodeOK
 }
+
+func processEOFError(scheme string, err error) error {
+	if strings.Contains(err.Error(), "EOF") && scheme == api.HttpScheme {
+		err = fmt.Errorf("%s. Scheme %s was used, "+
+			"this error can be caused by using http to connect to an https CTS instance, "+
+			"consider using %s scheme instead", err, scheme, api.HttpsScheme)
```
findkim, line 110: > this error can be caused by a client using http to connect to a CTS server with TLS enabled, consider using https scheme instead

took a jab at rephrasing "an https CTS instance" since it reads colloquial. Feel free to revise as you see fit

findkim (approved):
Appreciate the friendlier error message that is actionable 👍

PR #507, opened 2021-11-15
Added new tests for CA path

wilkermichael:
- Added test cases for both TLS and MTLS using CAPath instead of CACert. CAPath allows for a directory of certs to be passed to server and client instead of just a single file. Then the server and/or client is able to authenticate using the correct cert.
- Modified test cases so that they don't use self-signed certificates (CA and Cert being the same file). Instead using a CA, `localhost_cert.pem` and a cert signed by the CA, `localhost_leaf_cert.pem` which would be a more realistic scenario.
- Added testutil functions for copying test files to test directories. This was needed since CAPath requires that only certs (no keys, readmes, etc.) be in the provided directory.

Example CTS Configuration:
```
tls {
  enabled = true
  cert = "/path/to/certs/localhost_cert.pem"
  key = "/path/to/certs/localhost_key.pem" 
  verify_incoming = true
  ca_path = "/path/to/certs"
}
```

Example CTS mTLS client using flags:
```
consul-terraform-sync task enable \
    -http-addr="https://localhost:8501" \
    -ca-path="/path/to/certs" \
    -client-cert="/path/to/certs/localhost_cert.pem" \
    -client-key="/path/to/certs/localhost_key.pem"  \ 
    example-task
```

Same example, this time using some environment variables
```
export CTS_CAPATH="/path/to/certs"
export CTS_CLIENT_CERT="/path/to/certs/localhost_cert.pem"
export CTS_CLIENT_KEY="/path/to/certs/localhost_key.pem"
export CTS_ADDRESS="https://localhost:8501"

consul-terraform-sync task enable example-task
```

Part of #466

### DIFF
--- a/e2e/command_test.go
+++ b/e2e/command_test.go
@@ -22,7 +22,8 @@ import (
 )
 
 const (
-	invalidCert = "../testutils/certs/consul_cert.pem"
+	invalidCACert = "../testutils/certs/consul_cert.pem"
+	missingCACert = "../testutils/certs/localhost_cert2.pem"
 )
 
 // TestE2E_MetaCommandErrors tests cases that cross subcommands coded in
@@ -115,6 +116,13 @@ func TestE2E_CommandTLSErrors(t *testing.T) {
 	cts := ctsSetupTLS(t, srv, tempDir, dbTask(), tlsc)
 	address := cts.FullAddress()
 
+	// Get all test certs and move them to the CA path directory
+	certs := []string{
+		invalidCACert,
+		missingCACert,
+	}
+	clientCAPath := copyClientCerts(t, certs, tempDir)
+
 	commands := map[string][]string{
 		"task disable": {"task", "disable"},
 		"task enable":  {"task", "enable"},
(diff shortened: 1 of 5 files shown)


e2e/utils.go
```diff
@@ -62,6 +66,14 @@ func defaultTLSConfig() tlsConfig {
 	}
 }
 
+func newTLSConfig(tempDir string) tlsConfig {
```
mkam, line 69: These method names (newTLSConfig and newMTLSConfig) feel a little off to me because they're specific to CAPath, but the name doesn't imply that.

e2e/command_test.go
```diff
@@ -467,6 +562,100 @@ func TestE2E_CommandMTLS(t *testing.T) {
...
+		{
+			name: "ca path environment variables",
+			args: []string{
+				dbTaskName,
+			},
+			envVariables: []string{
+				fmt.Sprintf("%s=%s", api.EnvAddress, cts.FullAddress()),
+				fmt.Sprintf("%s=%s", api.EnvTLSCAPath, tlsc.caPath),
+				fmt.Sprintf("%s=%s", api.EnvTLSSSLVerify, "true"),
+				fmt.Sprintf("%s=%s", api.EnvTLSClientCert, defaultCTSClientCert),
```
mkam, line 620: Can we test different client certs on different test cases? It would be good to check that multiple certificates in the CAPath are being used.

e2e/utils.go
```diff
@@ -134,6 +156,22 @@ func ctsSetup(t *testing.T, srv *testutil.TestServer, tempDir string, taskConfig
...
 	cleanup := testutils.MakeTempDir(t, tempDir)
+
+	// Setup CA path directory if provided
+	var cleanupCAPath func() error
+	if len(tlsConfig.caPath) != 0 {
+		// Create CA Path directory
+		cleanupCAPath = testutils.MakeTempDir(t, tlsConfig.caPath)
+
+		// Get all test certs and move them to the CA path directory
+		certs := testutils.FindFileMatches(t, testCertPath, "*cert*.pem")
```
mkam, line 167: Should we be moving all test certificates? If we're more selective, we can then test that cert A and B in the CA path are valid and cert C that is not in the directory does not work.

mkam (approved):
Looks good to me, thanks for making those changes!

PR #509, opened 2021-11-17
Update backport assistant for 0.4.2 custom release

lornasong:
The current backporting assistant supports a label like `backport/0.4` to backport to a branch like `release/0.4.x`.

For 0.4.2 release we deviated from our typical release process and are planning to release from a branch like `release/0.4.2`. If want an associated label like `backport/0.4.2`, this requires different backporting logic.

Solution in PR is the most readable (and easy to delete) solution I could think of that didn't change our existing release process. Welcome other solutions. Opening a PR was the easiest way to share context on this problem.


lornasong:
Closing: will instead merge `main` into `release/0.4.x` so we can continue to use `release/0.4.x` instead of `release/0.4.2` https://github.com/hashicorp/consul-terraform-sync/pull/512

PR #510, opened 2021-11-17
Add openapi and codegen

wilkermichael:
@mkam I've created this PR Draft so we can discuss the API. This is a WIP so where these files live is still kinda up in the air, this is more about just reviewing the api itself. I generated the Go code as well, so you can take a look at the structs generated, you can also copy the openapi.yaml contents into the [swagger editor](https://editor.swagger.io/)

Part of #522


openapi.yaml
```diff
@@ -0,0 +1,511 @@
...
+          type: object
+          properties:
+            services:
+              $ref: '#/components/schemas/ServicesSourceInput'
+            consul-kv:
+              $ref: '#/components/schemas/ConsulKVSourceInput'
+      required:
+        - name
+        - source
+        - version
```
mkam, line 358: Version isn't required, it defaults to the latest if not provided.

openapi.yaml
```diff
@@ -0,0 +1,511 @@
...
+          type: string
+          example: "* * * * Mon"
+
+    ServicesSourceInput:
+      type: object
+      properties:
+        regexp:
+          type: string
+          example: "web.*"
+    CatalogSourceInput:
```
mkam, line 447: I don't believe we have catalog as an option for source input.

openapi.yaml
```diff
@@ -0,0 +1,511 @@
...
+    TerraformProvider:
+      type: object
+
+    CatalogServicesCondition:
+      type: object
+      properties:
+        regexp:
+          type: string
+          example: "web.*"
+        namespace:
```
mkam, line 404: Datacenter is missing from here.

openapi.yaml
```diff
@@ -0,0 +1,511 @@
...
+            $ref: '#/components/schemas/Provider'
+        ruquest_id:
+          $ref: '#/components/schemas/RequestID'
+
+    ProviderResponse:
+      type: object
+      properties:
+        provider:
+          $ref: '#/components/schemas/Provider'
+        ruquest_id:
```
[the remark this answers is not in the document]
wilkermichael: nice catch, thanks!

openapi.yaml
```diff
@@ -0,0 +1,511 @@
...
+    Error:
+      required:
+        - message
+      properties:
+        error:
+          type: object
+          properties:
+            message:
+              type: string
+              example: "task is currently running and cannot be deleted at this time"
```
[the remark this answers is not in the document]
wilkermichael: that's the example, I haven't fully fleshed out the examples yet, but they can be made custom for each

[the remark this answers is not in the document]
wilkermichael: we can just make it more generic

api/client.go
```diff
@@ -363,18 +363,18 @@ func (s *Status) Task(name string, q *QueryParam) (map[string]TaskStatus, error)
 	return taskStatuses, nil
 }
 
-// Task can be used to query the task endpoints
-type Task struct {
+// TaskClient can be used to query the task endpoints
```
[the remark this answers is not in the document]
wilkermichael: it collides with the Task rep, also I thought TaskClient made the struct clearer. 

Also, we could make the c *Client anonymous, then you wouldn't have a stutter?

openapi.yaml
```diff
@@ -0,0 +1,511 @@
...
+      host:
+        default: 'localhost'
+      port:
+        default: '8558'
+      basePath:
+        enum:
+          - 'v1'
+        default: v1
+paths:
+  /providers:
```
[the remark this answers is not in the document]
wilkermichael: yeah, that was my intent for sure

openapi.yaml
wilkermichael: yep, right, it's not included under the task.SourceInputs either, this was copy-paste error from conditions

api/client.go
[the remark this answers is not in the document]
wilkermichael: I decided to refactor Status and Task client to be TaskClient and StatusClient respectively. The `c *Client` field has now been made anonymous so that TaskClient and StatusClient now inherit *Client

I feel that:
- This will eliminate naming collisions with representations in the future
- The names now better reflect what these structs are

api/handler.go
```diff
@@ -0,0 +1,43 @@
...
+// Make sure we conform to ServerInterface
+var _ oapigen.ServerInterface = (*Handler)(nil)
+
+const (
+	handlerSubsystemName = "handler"
+)
+
+// Handler is composed of CTS server handlers which allows
+// the handler to adhere to the generated server interface
+type Handler struct {
```
lornasong, line 20: nit: what are your thoughts on naming `Handler` and its subsystem name to something like `Handlers` (plural)? It seems like `Handler` is an aggregation of the handlers rather than a handler itself but I might be misunderstanding.

api/openapi.yaml
```diff
@@ -0,0 +1,327 @@
...
+          example: "default"
+        node_meta:
+          type: object
+          additionalProperties:
+            type: string
+          example:
+            key: value
+        source_includes_var:
+          type: boolean
+          example: false
```
lornasong, line 254: The default for `source_includes_var` is false. Would we want to include the default value for this for consistency?

api/openapi.yaml
```diff
@@ -0,0 +1,327 @@
...
+    ConsulKVCondition:
+      type: object
+      properties:
+        path:
+          type: string
+          example: "my-key"
+        recurse:
+          type: boolean
+          example: false
```
lornasong, line 263: The default for `recurse` is false. Would we want to include the default value for this for consistency?

lornasong (commented):
great to see the api docs in the swagger editor! left a couple of minor comments.


we also chatted offline and documenting it here for posterity (please correct if I misrepresent anything):

- This PR will not include configuration changes proposed for config streamlining. This will be done in a separate PR after config streamlining changes are made
- The new endpoint's handler implementation are commented out so that when they are implemented, they can be reviewed then
- Dry-run endpoint naming/path is a placeholder for now and may change
- Existing endpoints will be converted to this new way in a separate PR after we confirm the openapi implementation

api/openapi.yaml
```diff
@@ -0,0 +1,327 @@
...
+          description: unexpected error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+  /v1/tasks/{name}:
+    delete:
+      summary: Deletes a task by name
+      operationId: deleteTaskByName
+      description: deletes a single task based on the name provided
```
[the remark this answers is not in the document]
wilkermichael: looking at the Swagger Editor example, they go with capitals, so I can change to that. We can just be consistent with that.

api/openapi.yaml
```diff
@@ -0,0 +1,327 @@
...
+            min:
+              type: string
+              example: "20s"
+        condition:
+          type: object
+          properties:
+            catalog-services:
+              $ref: '#/components/schemas/CatalogServicesCondition'
+            services:
+              $ref: '#/components/schemas/ServicesSourceInput'
```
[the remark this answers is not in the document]
wilkermichael: yep, I'm okay with that, made `ServicesCondition` object

api/openapi.yaml
wilkermichael: sure, makes sense

api/openapi.yaml
```diff
@@ -0,0 +1,327 @@
...
+          example: "default"
+        source_includes_var:
+          type: boolean
+          example: false
+    ScheduleCondition:
+      type: object
+      properties:
+        cron:
+          type: string
+          example: "* * * * Mon"
```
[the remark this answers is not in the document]
wilkermichael: added, from the docs I found the following required fields:
ConsulKV Condition and SourceInput: `path`
Schedule Condition: `cron`

api/openapi.yaml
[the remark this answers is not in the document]
lornasong: Good to know! Thank you

api/handler.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/hashicorp/consul-terraform-sync/api/oapigen"
+	"github.com/hashicorp/consul-terraform-sync/logging"
+)
+
+// Make sure we conform to ServerInterface
+var _ oapigen.ServerInterface = (*Handlers)(nil)
+
+const (
+	handlerSubsystemName = "handler"
```
lornasong, line 15: Not sure how much this const is tied to `Handlers`. Does it make sense to change this to `handlersSubsystemName = "handlers"`?

api/openapi.yaml
```diff
@@ -0,0 +1,341 @@
...
+          example: false
+        datacenter:
+          type: string
+          example: "dc1"
+        namespace:
+          type: string
+          example: "default"
+        source_includes_var:
+          type: boolean
+          example: false
```
lornasong, line 280: Maybe add `default: false` here?

api/openapi.yaml
```diff
@@ -0,0 +1,341 @@
...
+          example: "web.*"
+    ConsulKVSourceInput:
+      type: object
+      properties:
+        path:
+          type: string
+          example: "my-key"
+        recurse:
+          type: boolean
+          example: false
```
lornasong, line 306: Maybe add `default: false` to recurse here?

api/openapi.yaml
```diff
@@ -0,0 +1,341 @@
+openapi: "3.0.0"
```
lornasong, line 1: While scrolling around for the review, I noticed this file is a little large for navigating. Not an action for this PR, just noting a thought  - wondering if openapi would support it if we broke the yaml into multiple files e.g. paths and components in different files. Again, not an action item for this PR but thinking down the line as we expand on this file!

lornasong (approved):
this looks good to me. it looks like the feedback from @ mkam is also addressed

approving! left a couple minor things only if it makes sense to address

api/openapi.yaml
wilkermichael: Yeah, that's something we could definitely look into, I think you can reference schemas at the very least, so could break the schemas into their own files

wilkermichael: Creating an issue to track this for the future

Issue #357, opened 2021-08-17
Support query parameters in services condition


mkam:
#### Description
Support query parameters `filter`, `datacenter`, and `namespace` in the services condition of a task. 

Note that `node_meta` (an option in the catalog-services condition block) and `tag` (an option in the services block) are not included, as they are both deprecated in the [Consul health service API](https://www.consul.io/api-docs/health#parameters-2) in favor of the filter parameter.

```hcl
 condition "services" {
     filter = "\"example\" in Service.Tags"
     datacenter  = "dc1"
     namespace = "default"
}
```

#### Use Cases
1. Task has a condition with regex and query parameters.

- The names of the services are not known beforehand, so cannot create service blocks for each service. 
- The user knows the names of the services, but does not want to have to create multiple service blocks if all of them use the same parameters.

```hcl
task {
  name = "services_condition_task"
  source = "path/to/services-condition-module"
  condition "services" {
     regexp = ".*"
     filter = "\"example\" in Service.Tags"
     datacenter  = "dc2"
     namespace = "test-ns"
}
```

2. Task has a list of services and a condition with query parameters.

- The user does not want to have to create multiple service blocks when all of them use the same parameters.

```hcl
task {
  name = "services_condition_task"
  source = "path/to/services-condition-module"
  services = ["api", "db", "web"]
  condition "services" {
     filter = "\"example\" in Service.Tags"
     datacenter  = "dc2"
     namespace = "test-ns"
}
```

3. Task has a condition with query parameters and either a regex or a list of services. Service block has a service that a service in the task.

- User has a grouping of services that all mostly require one set of parameters but has specific exceptions for certain services.
- User has a commonly used service but for the specific task, needs to use a different set of parameters.

Considerations:
- What do we prioritize, task condition parameters or service parameters? Would probably depend on which of the use cases we believe will be more common.
- Do we merge parameters or choose one entire set of parameters over the other?

Regex:
```hcl
task {
  name = "services_condition_task"
  source = "path/to/services-condition-module"
  condition "services" {
     regexp = ".*"
     filter = "\"example\" in Service.Tags"
     datacenter  = "dc2"
     namespace = "test-ns"
}

service {
   name = "web"
   datacenter = "dc1"
}
```

```hcl
task {
  name = "task_1"
  source = "path/to/module"
  services = ["api", "db", "web"]
}

task {
  name = "task_2"
  source = "path/to/module"
  services = ["web"]
}

task {
[...]

PR #523, opened 2021-11-24
Implement Services Condition & Source Input New Query Fields

lornasong:
Implement new query fields:
- Datacenter
- Namespace
- Filter

Note:
 - Configuration added in earlier PR: https://github.com/hashicorp/consul-terraform-sync/pull/520
 - New config field cts_user_defined_meta will be implemented in a separate PR

Changes:
 - Pass new query fields to ServicesMonitor templating
 - Add integration test for services-regex templating
 - Update golden test with new query fields
 - Test new query field `filter` in services condition e2e test
   - Note: manually tested services source input because source input is currently only supported with scheduled conditions. Adding testing for this to scheduled conditions e2e test felt out of scope. Will add e2e tests when source input is supported by other conditions)

Part of #357


wilkermichael (approved):
Looks good Lorna!

PR #524, opened 2021-11-24
Implement Services Condition & Source Input MetaData field

lornasong:
Implement new field `cts_user_defined_meta`

Note:
 - This is a parallel field to the existing `cts_user_defined_meta` in the `service` block
 - Configuration added in earlier PR: https://github.com/hashicorp/consul-terraform-sync/pull/520
 - New query fields implemented in earlier PR: https://github.com/hashicorp/consul-terraform-sync/pull/523

Changes:
 - Update ServicesMeta to support both `cts_user_defined_meta` in the `service` block (existing) and in `condition "services"` new
 - Update references to ServicesMeta
 - Update driver to pass both types of metadata to template
 - Test new `cts_user_defined_meta` in services condition e2e test
   - Note: manually tested for services source input because source input is currently only supported with scheduled conditions. Adding testing for this to scheduled conditions e2e test felt out of scope. Will add e2e tests when source input is supported by other conditions)

Part of #357

### DIFF
--- a/driver/terraform.go
+++ b/driver/terraform.go
@@ -615,9 +615,12 @@ func getTerraformHandlers(taskName string, providers TerraformProviderBlocks) (h
 }
 
 // getServicesMetaData helps retrieve metadata which can come from a number of
-// configuration sources: service block, condition block, source_input block.
-// Currently, it is only possible for a task to have metadata defined in one of
-// the listed sources
+// configuration sources: task.services' related service block, condition
+// "service" block, source_input "service" block.
+//
+// Currently it is only possible for a task to be configured with one of these
+// configuration sources, hence a task only has one metadata source. This is
+// validated at the configuration-level. This method relies on this assumption.
 func (tf *Terraform) getServicesMetaData() (*tmplfunc.ServicesMeta, error) {
 	servicesMeta := &tmplfunc.ServicesMeta{}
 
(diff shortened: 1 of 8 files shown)


driver/terraform.go
```diff
@@ -610,6 +609,55 @@ func getTerraformHandlers(taskName string, providers TerraformProviderBlocks) (h
 	return next, nil
 }
 
+// getServicesMetaData helps retrieve metadata which can come from a number of
+// configuration sources: service block, condition block, source_input block.
+// Currently, it is only possible for a task to have metadata defined in one of
+// the listed sources
```
mkam, line 615: Do we error if more than one is defined or does it just have a precedence order?

mkam (approved):
Looks good to me! The comments that clarified whether the code was for the old vs. new meta were super helpful!

driver/terraform.go
lornasong: Ah good question! So with some later config streamlining changes, I'm planning to error at the configuration-level so that a task cannot be configured with more than one of `task.services`, `condition "services"`, `source_input "services"`. Ideally if more than one metadata is defined, it should error before reaching this method. I should have shared this context and can update the comment to clarify this!

To actually answer your question, if somehow more than one metadata is defined, this implementation is just precedence of order. My initial thought is that it would be ok to not error here because of the configuration validation, but I might be overlooking something. Please let me know if you have thoughts or concerns!

mkam: That all makes sense to me, thanks for clarifying!

lornasong: Thanks for checking me! I will add a comment with this clarifying details

lornasong:
Rebase to latest on main

PR #526, opened 2021-11-30
Delete Task API

mkam:
Adds support for the Delete Task API.

- Add methods for deleting the driver and events
- Add deletion API to task lifecycle handler
- Add a test module that sleeps for a default 5s so we can test actions on running tasks

 Future PR will have relevant client/CLI changes.

Examples:
Happy path
```
$ curl -v  -X DELETE  http://127.0.0.1:8558/v1/tasks/test-task  | jq .
> DELETE /v1/tasks/test-task HTTP/1.1
> Host: 127.0.0.1:8558
> User-Agent: curl/7.64.1
> Accept: */*
>
< HTTP/1.1 200 OK
< Access-Control-Allow-Headers: Content-Type
< Access-Control-Allow-Methods: GET, PUT, PATCH, POST, DELETE
< Access-Control-Allow-Origin: *
< Content-Type: application/json
< Date: Tue, 30 Nov 2021 22:18:59 GMT
< Content-Length: 54
<
{
  "request_id": "9b23eea7-a435-2797-c71e-10c15766cd73"
}
```
Task does not exist
```
$ curl -v  -X DELETE  http://127.0.0.1:8558/v1/tasks/nonexistent-task  | jq .
...
< HTTP/1.1 404 Not Found
...
{
  "error": {
    "message": "a task with the name 'nonexistent-task' does not exist or has not been initialized yet"
  },
  "request_id": "d207d86d-5e93-12c8-e3c1-b6e8d6fe9d14"
}
```
Task is currently running
```
$ curl -v  -X DELETE  http://127.0.0.1:8558/v1/tasks/delayed-task  | jq .
...
< HTTP/1.1 409 Conflict
...
{
  "error": {
    "message": "task 'delayed-task' is currently running and cannot be deleted at this time"
  },
  "request_id": "46f23047-24f7-050c-42bf-3033d219d699"
}
```


api/task_delete.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/hashicorp/consul-terraform-sync/logging"
+)
+
+// deleteTask deletes an existing task and its events. Does not delete
+// if the task is active.
+func (h *taskHandler) deleteTask(w http.ResponseWriter, r *http.Request) {
```
wilkermichael: Would it be better to replace this function signature with the (h *TaskLifeCycleHandler) DeleteTaskByName(w http.ResponseWriter, r *http.Request, name string) method? Then you wouldn't need a separate call out from that method, and the name string will already be parsed for you.

api/task_delete.go
```diff
@@ -0,0 +1,60 @@
...
+	taskName, err := getTaskName(r.URL.Path, taskPath, h.version)
+	logger := logging.FromContext(r.Context()).Named(deleteTaskSubsystemName)
+	if err != nil {
+		logger.Trace("bad request", "error", err)
+		jsonErrorResponse(r.Context(), w, http.StatusBadRequest, err)
```
wilkermichael, line 17: I would look at the `api/handler.go sendError` method instead, this method conforms to the schema defined in the task lifecycle documentation we worked on and in the openapi.yaml.

I didn't change the behaviour of jsonErrorResponse because I really didn't want to touch any of the old api at this point.

wilkermichael: Also see one of the example handlers in task.go on how to extract the requestID from the context to pass to this method.

api/task_delete.go
```diff
@@ -0,0 +1,60 @@
...
+		return
+	}
+
+	if taskName == "" {
```
wilkermichael, line 21: Since the task name is part of the path, if we routed to this handler without a taskName that would actually be an error in how the programmer setup routing. Since it a programmer error, and not one a user could perform, we might want to panic instead? Or else not include this check?

api/task_delete.go
```diff
@@ -0,0 +1,60 @@
...
+		err := fmt.Errorf("no task name was included in the api request. " +
+			"Deleting a task requires the task name: '/v1/tasks/:name'")
+		logger.Trace("bad request", "error", err)
+		jsonErrorResponse(r.Context(), w, http.StatusBadRequest, err)
+		return
+	}
+	logger.Trace("deleting task", "task", taskName)
```
wilkermichael, line 28: for consistency, maybe call the field task_name since you're not logging the entire task?

wilkermichael: same comment for other logs with taskName

api/task_delete.go
```diff
@@ -0,0 +1,60 @@
...
+			"been initialized yet", taskName)
+		logger.Trace("task not found", "task", taskName, "error", err)
+		jsonErrorResponse(r.Context(), w, http.StatusNotFound, err)
+		return
+	}
+
+	// Check if task is active
+	if h.drivers.IsActive(taskName) {
+		err := fmt.Errorf("task '%s' is currently running and cannot be deleted "+
+			"at this time", taskName)
```
wilkermichael, line 43: Should we add instructions in the error to try again?

api/task_delete_test.go
```diff
@@ -0,0 +1,100 @@
...
+			taskName := "task_a"
+			drivers := driver.NewDrivers()
+			d := new(mocks.Driver)
+			drivers.Add(taskName, d)
+			if tc.active {
+				drivers.SetActive(taskName)
+			}
+
+			store := event.NewStore()
+			store.Add(event.Event{TaskName: taskName})
```
wilkermichael, line 72: Any reason we shouldn't require the error returned here is nil?

api/task_delete_test.go
```diff
@@ -0,0 +1,100 @@
+package api
```
wilkermichael, line 1: Great test coverage!

driver/drivers.go
```diff
@@ -110,3 +110,15 @@ func (d *Drivers) IsActive(name string) bool {
 	_, ok := d.active.Load(name)
 	return ok
 }
+
+func (d *Drivers) Delete(taskName string) error {
```
wilkermichael, line 114: Add a function comment here since this is a public method

event/store.go
```diff
@@ -68,3 +68,12 @@ func (s *Store) Read(taskName string) map[string][]Event {
 	}
 	return ret
 }
+
+func (s *Store) Delete(taskName string) {
```
wilkermichael, line 72: Add a function comment here since this is a public method

event/store_test.go
```diff
@@ -142,3 +142,46 @@ func TestStore_Read(t *testing.T) {
...
+			[]Event{},
+			map[string][]Event{},
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			store := NewStore()
+			for _, event := range tc.values {
+				store.Add(event)
```
wilkermichael, line 179: should we be checking the error here? 

If you decide we should, can you also update the other `store.Add()` calls in this file

controller/readwrite.go
```diff
@@ -328,6 +328,8 @@ func (rw *ReadWrite) checkApply(ctx context.Context, d driver.Driver, retry, onc
 	// new data
 	if rendered {
 		rw.logger.Info("executing task", taskNameLogKey, taskName)
+		rw.drivers.SetActive(taskName)
+		defer rw.drivers.SetInactive(taskName)
```
wilkermichael, line 332: Just verifying my understanding.

We set it to active while running, so that now, if we call `delete` via the api, we cannot delete an `active` task correct?

api/task_delete.go
wilkermichael: You also wouldn't need the version either anymore, name would parse out the name regardless of the endpoint version

api/task.go
```diff
@@ -184,14 +185,22 @@ func (h *taskHandler) updateTask(w http.ResponseWriter, r *http.Request) {
...
-func NewTaskLifeCycleHandler() *TaskLifeCycleHandler {
+func NewTaskLifeCycleHandler(store *event.Store, drivers *driver.Drivers,
+	version string) *TaskLifeCycleHandler {
 	return &TaskLifeCycleHandler{
-		//tasks: make(map[string]oapigen.Task),
+		mu:      &sync.RWMutex{},
+		store:   store,
+		drivers: drivers,
+		version: version,
+		tasks:   make(map[string]oapigen.Task),
```
wilkermichael, line 203: you can leave this line commented out, it's only for testing

api/task.go
```diff
@@ -184,14 +185,22 @@ func (h *taskHandler) updateTask(w http.ResponseWriter, r *http.Request) {
...
 type TaskLifeCycleHandler struct {
-	// TODO: replace this when implementing endpoint
-	//tasks map[string]oapigen.Task
-	//lock  sync.Mutex
+	mu *sync.RWMutex
+
+	store   *event.Store
+	drivers *driver.Drivers
+	version string
+	tasks   map[string]oapigen.Task
```
wilkermichael, line 193: you can leave this line commented out, it's only for testing

e2e/api_test.go
```diff
@@ -322,6 +322,109 @@ func TestE2E_TaskEndpoints_UpdateEnableDisable(t *testing.T) {
 	testutils.CheckDir(t, false, resourcesPath)
 }
 
+// TestE2E_TaskEndpoints_Delete tests the delete task endpoint. This
```
wilkermichael, line 325: These are great! :D

wilkermichael (commented):
Great work Melissa, really thorough covering all the bases for delete. Love the amount of testing as well!

controller/readwrite.go
mkam: Yeah, there are a lot of complications around deleting a task when it's running, such as:

- The request will hang if until the task is complete
- At the end of the task running, an event is stored, and all events for a task need to be deleted when the task is deleted
- When we add cleanup in the future, we may clean something up but the running task might just add it back

So I made it where it returns the 409 Conflict when we delete an active, running task.

api/task.go
mkam: Oh, okay! I should have asked about what it was for but I forgot! Will we need this tasks map in the future beyond for testing?

api/task_delete.go
mkam: I think if we add that, it would be more of a CLI error message than an API message.

api/task_delete_test.go
mkam: No reason, I can add

event/store_test.go
mkam: Wouldn't hurt to add a check, I can make those changes!

api/task_delete.go
mkam: That makes sense, I'll update!

api/task_delete.go
mkam: Totally forgot about request id! 🤦‍♀️ I'm thinking now too we could return the request id as part of the delete response instead of the empty JSON it currently returns.

api/task_delete.go
wilkermichael: You're right, that makes sense to me

api/task.go
wilkermichael: No I don't think it'll be needed. It was just so that I could test creating a task in an in memory store so I could then delete it using the delete endpoint. Just for some added "realism" to the testing scenario.

api/task_delete.go
mkam: Yeah, I was originally keeping it separate so that I could test it with the old handler and then easily switch it to the new handler. It doesn't feel all that necessary anymore to have it work for both, so I'll get rid of the DeleteTaskByName in task.go and replace it with this.

api/task_delete.go
mkam: Yeah, I think I'll remove the check. When I wrote the code, I had more in mind the original handler since the openapi stuff was still WIP, so that's why I needed this check! But now that it's specifically for the open api handler, it doesn't seem necessary.

wilkermichael (approved):

mkam:
@wilkermichael Thanks for all the feedback! Everything should be addressed now!

wilkermichael:
> 

Sounds good! I pre-approved it

wilkermichael:
Oh sorry, I see there are new commits, I'll do a once-over

wilkermichael (approved):
looks good!

PR #527, opened 2021-11-30
Feature 522 handle create task request to config conversion

wilkermichael:
This commit does a few things to prepare for creating a new task based on an api request:

1) Modified the openapi.yaml to separate out some of the embedded definitions into their own type definitions. Embedded type definitions are honoured in the code generation which results in embedded structs. To make the generated code more manageable, it simplifies things to separate out definitions so that structs are generated for those types rather than being embedded.

2) Created a wrapper around the generated task request so that I could add two methods:
-  `ToConfigTaskConfig` for converting the request to a config object for further processing. This has the added benefit of allowing for use of the configs excellent validate and finalize methods.
- `String` which allows for the request object to be converted to a string under the hood by the logger with all pointers resolved. This is helpful for logging.
- I opted to just marshal the string to json for the `String` method, I think this simplifies the string creation, and the object already has the correct json tags in place to support this. Let me know what you think about this

**Example of taskRequest.String() usecase**
```Go
	logger := logging.FromContext(r.Context()).Named(createTaskSubsystemName)
	
	// Decode the new task and add it to our "database"
	var req taskRequest
	requestID := requestIDFromContext(r.Context())
	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
		sendError(w, r, http.StatusBadRequest, "invalid format for create task request")
		return
	}
	
	logger.Trace("create task request", "create_task_request", req)
```

**Log With String Method**
`2021-11-30T09:45:01.261-0800 [TRACE] api.createtask: create task request: reqID=eaf7b6bd-fef9-3f38-02d0-e382f4bf2edc
  create_task_request=
  | {"buffer_period":{"enabled":true,"max":"0s","min":"0s"},"description":"Writes the service name, id, and IP address to a file","enabled":true,"name":"new-example-task","providers":["local"],"services":["api"],"source":"./example-module"}`

**Log Without String Method**
`2021-11-30T09:46:46.066-0800 [TRACE] api.createtask: create task request: reqID=0b4c2abe-ad99-5ef5-375e-53e3a3777916 create_task_request="{0xc0000b93c8 <nil> 0xc00018c2d0 0xc0002404f0 new-example-task 0xc0000b9368 0xc0000b9398 ./example-module <nil> <nil> <nil> <nil>}"`

Some background on why there is a mix of pointers and non-pointer types in the generated structs. The code generation will generate optional fields as pointer types and required fields as non-pointer types.


api/representation_test.go
```diff
@@ -0,0 +1,431 @@
...
+}
+
+func TestTaskRequest_ToConfigTaskConfig(t *testing.T) {
+	cases := []struct {
+		name               string
+		request            *taskRequest
+		taskConfigExpected *config.TaskConfig
+	}{
+		{
+			name: "minimum required only",
```
wilkermichael, line 54: I'll add underscores to be consistent with other test cases

api/representation_test.go
```diff
@@ -0,0 +1,431 @@
...
+				TFVersion:    config.String(""),
+				BufferPeriod: config.DefaultBufferPeriodConfig(),
+				Enabled:      config.Bool(true),
+				Condition:    config.DefaultConditionConfig(),
+				WorkingDir:   config.String("sync-tasks"),
+				SourceInput:  config.DefaultSourceInputConfig(),
+			},
+		},
+		{
+			name: "basic fields filled",
```
wilkermichael, line 78: same here

api/oapigen/types.go
```diff
@@ -8,6 +8,13 @@ import (
 	"fmt"
 )
 
+// BufferPeriod defines model for BufferPeriod.
+type BufferPeriod struct {
+	Enabled *bool   `json:"enabled,omitempty"`
```
findkim, line 13: What's the context behind setting `omitempty` for all fields in the API response types? I find that working with a JSON API is most predictable when the fields are present with empty values, and that `omitempty` is useful for optional relationship fields.

api/oapigen/types.go
```diff
@@ -22,6 +29,14 @@ type CatalogServicesCondition_NodeMeta struct {
 	AdditionalProperties map[string]string `json:"-"`
 }
 
+// Condition defines model for Condition.
+type Condition struct {
+	CatalogServices *CatalogServicesCondition `json:"catalog-services,omitempty"`
+	ConsulKv        *ConsulKVCondition        `json:"consul-kv,omitempty"`
```
findkim, line 35:
```suggestion
	ConsulKV        *ConsulKVCondition        `json:"consul-kv,omitempty"`
```

Can you update the other occurrences to have `Kv` meet go idiomatic casing for initials?

api/representation.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	return tc, nil
+}
+
+// String writes out the task request in an easily readable way
+// useful for logging
+func (tr taskRequest) String() string {
+	data, _ := json.Marshal(tr)
+	return fmt.Sprintf("%s\n", data)
```
findkim, line 121: The stringer functions in go don't have opinionated new lines, can you remove it?

api/representation_test.go
```diff
@@ -0,0 +1,431 @@
...
+)
+
+func TestTaskRequest_String(t *testing.T) {
+	var req taskRequest
+
+	err := json.Unmarshal([]byte(taskReq), &req)
+	require.NoError(t, err)
+
+	actual := fmt.Sprintf("%s", req)
+	expected := "{\"buffer_period\":{\"enabled\":true,\"max\":\"0s\",\"min\":\"0s\"},\"description\":\"Writes the service name, id, and IP address to a file\",\"enabled\":true,\"name\":\"new-example-task\",\"providers\":[\"local\"],\"services\":[\"api\"],\"source\":\"./example-module\"}\n"
```
findkim, line 43: if possible, it's easier to read if it's wrapped around string literal ` instead of double quotes which require escaping each quote, like the above taskReq

api/oapigen/types.go
findkim: Or are these objects only used as request types and not related to response types?

wilkermichael: So the code generation will make any field that's not specifically 'required' as part of the schema a pointer. It will also omit empty pointers. It would be up to us to initialize the pointers and return them if we want them to be included in responses, otherwise you would see.

```Json
{
   "BufferPeriod"{
     "enabled": null
     . . .
  }
}
```

So omitempty in this case just stops us from returning nil values

```Go
taskRequest{
  Name:       "test-name",
  Source:     "test-source",
  Services:   &[]string{"api", "web"},
  WorkingDir: config.String("sync-tasks"),
  BufferPeriod: &oapigen.BufferPeriod{
	  Enabled: config.Bool(false),
	  Max:     config.String(""),
	  Min:     config.String(""),
  },
}
```
If I created a task with a bufferperiod like the above max and min would both be returned in the json as empty strings: `{"buffer_period":{"enabled":false,"max":"","min":""},"name":"test-name","services":["api","web"],"source":"test-source","working_dir":"sync-tasks"}`

else if I didn't initialize Min and Max:
```Go
taskRequest{
  Name:       "test-name",
  Source:     "test-source",
  Services:   &[]string{"api", "web"},
  WorkingDir: config.String("sync-tasks"),
  BufferPeriod: &oapigen.BufferPeriod{
	  Enabled: config.Bool(false),
  },
}
```

I would get: `{"buffer_period":{"enabled":false},"name":"test-name","services":["api","web"],"source":"test-source","working_dir":"sync-tasks"}`

api/oapigen/types.go
wilkermichael: Unfortunately the code generation isn't smart enough to figure out abbreviations like KV and ID, so we get `Id` and `Kv`.

I'll look into it some more, but nothing has come up so far in my research for a workaround

api/representation.go
wilkermichael: done!

api/representation_test.go
wilkermichael: Some background, the reason I included it like this, is because this is how `require.Equal(t, expected, actual)` returns the expected matching string when you don't match, so it was easy to copy and paste the expected match after verifying it visually.

However, that's not a good enough reason not to change it, so I've made your suggested change!

api/oapigen/types.go
wilkermichael: Per your question about response types, yes these objects would also be part of the response type. If we feel that the response types `require` min and max to be returned, we can also create different response objects with those requirements, it would just result in duplicated code.

api/oapigen/types.go
findkim: Ohh this is auto-generated, this is fine as-is then! sorry i'm not very familiar with the code gen. Thanks for the context!

api/representation_test.go
findkim: I see! didn't realize the perk here, given that this is a nit and would be happy either way

api/oapigen/types.go
wilkermichael: All good! I mean, it looks surprisingly authentic for auto-generated code, so I don't blame you

api/oapigen/types.go
findkim: Thanks for the info! Cool, yeah definitely useful for request types and reflecting required values. Response types is tricky since it doesn't seem to play nice with request types with this framework. No suggestion here just yet in how to handle it, though it's not totally uncommon to have different objects for request vs response. We'll see where this takes us, happy to move forward as-is. Thanks for navigating the code-gen stuff

findkim (approved):

api/oapigen/types.go
wilkermichael: No problem, happy to help

PR #533, opened 2021-12-02
Delete Task CLI

mkam:
Adds the client and the CLI command for deleting a task.

**Example Outputs**
Happy path/approved
```
$ consul-terraform-sync task delete test-task

Enter a value: yes

==> Deleting task 'test-task'...

==> Deleted task 'test-task'
```
Denied
```
$ consul-terraform-sync task delete test-task

Enter a value: no

    Cancelled deleting task 'test-task'
```
Task does not exist
```
$ consul-terraform-sync task delete nonexistent
==> Do you want to delete 'nonexistent'?

Enter a value: yes

==> Deleting task 'nonexistent'...

==> Error: unable to delete 'nonexistent'
    request returned 404 status code with error: a task with the name
'nonexistent' does not exist or has not been initialized yet
```
Task is currently running
```
$ consul-terraform-sync task delete delayed-task
==> Do you want to delete 'delayed-task'?

Enter a value: yes

==> Deleting task 'delayed-task'...

==> Error: unable to delete 'delayed-task'
    request returned 409 status code with error: task 'delayed-task' is currently
running and cannot be deleted at this time
```

### DIFF
--- a/command/task_delete.go
+++ b/command/task_delete.go
@@ -35,7 +35,7 @@ func (c *taskDeleteCommand) Help() string {
 	helpText := fmt.Sprintf(`
 Usage: consul-terraform-sync task delete [options] <task name>
 
-  Task Delete is used to delete existing tasks. Will not delete a task
+  Task Delete is used to delete an existing task. Will not delete a task
   if the task is currently running.
 
 Options:
@@ -57,7 +57,7 @@ Example:
 
 // Synopsis is a short one-line synopsis of the command
 func (c *taskDeleteCommand) Synopsis() string {
-	return "Deletes existing tasks."
+	return "Deletes an existing task."
 }
 
 // Run runs the command
--- a/command/task_delete.go
+++ b/command/task_delete.go
@@ -1,0 +1,106 @@
+package command
+
+import (
+	"flag"
+	"fmt"
+	"strings"
+
+	"github.com/mitchellh/go-wordwrap"
+)
+
+const cmdTaskDeleteName = "task delete"
+
+// TaskDeleteCommand handles the `task delete` command
+type taskDeleteCommand struct {
+	meta
+
+	flags *flag.FlagSet
+}
+
+func newTaskDeleteCommand(m meta) *taskDeleteCommand {
+	flags := m.defaultFlagSet(cmdTaskDeleteName)
+	return &taskDeleteCommand{
+		meta:  m,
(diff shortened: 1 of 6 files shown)


command/task_delete.go
```diff
@@ -0,0 +1,106 @@
...
+func (c taskDeleteCommand) Name() string {
+	return cmdTaskDeleteName
+}
+
+// Help returns the command's usage, list of flags, and examples
+func (c *taskDeleteCommand) Help() string {
+	helpText := fmt.Sprintf(`
+Usage: consul-terraform-sync task delete [options] <task name>
+
+  Task Delete is used to delete existing tasks. Will not delete a task
```
findkim, line 38:
```suggestion
  Task Delete is used to delete an existing task. Will not delete a task
```

Since we're not supporting bulk delete at this time, it might be more clear to just keep "task" consistently singular

command/task_delete.go
```diff
@@ -0,0 +1,106 @@
...
+  if the task is currently running.
```
findkim, line 39: How do you think the delete error due to task running play at scale? I imagine a bad case when someone needs to delete a task because it's flapping terribly but would have to coincidentally catch the gap when it's not actively running. Would we be able to offer a queued delete that would occur right after the current run completes?

mkam: I'm still thinking through all the details for offering queued deletes but I think we can do it!

I mostly want to make sure that a run can't be queued up while a delete is in progress. The biggest issue with a run after deletion is that it creates an event that does not have an associated driver, which messes up the overall task status API. The secondary/less important problem is when we add cleanup support, the deleted resources would potentially get recreated. Again, I think we can do it, my brain is just still in catch up mode so I've got to think it through some more.

A workaround we could offer currently is to have the user disable the task, which will wait for the task to complete, and then follow up with a delete. Not an ideal workflow but just throwing it out there in case the queued deletes takes longer to implement than I anticipate!

command/task_delete.go
mkam: I've committed your suggestion and will add all the changes for plural to singular!

command/task_delete.go
findkim: Those are great cases to consider for marking deletion. Thanks for sharing your thoughts there and it's not blocking for this PR since it's CLI focused!

findkim (approved):

Issue #529, opened 2021-11-30
Conditions using catalog services endpoint are not using blocking queries


findkim:
### Describe the bug
Consul logs when running CTS with `services.regexp` or `catalog-services` conditions reveal that the `/v1/catalog/services` endpoint usage is not optimizing with blocking queries.

```
    2021-11-30T16:26:04.807-0600 [DEBUG] agent.http: Request finished: method=GET url=/v1/catalog/services from=127.0.0.1:49463 latency=152.816µs
    2021-11-30T16:26:04.926-0600 [DEBUG] agent.http: Request finished: method=GET url=/v1/catalog/services from=127.0.0.1:49463 latency=149.189µs
    2021-11-30T16:26:05.033-0600 [DEBUG] agent.http: Request finished: method=GET url=/v1/catalog/services from=127.0.0.1:49463 latency=140.917µs
    2021-11-30T16:26:05.137-0600 [DEBUG] agent.http: Request finished: method=GET url=/v1/catalog/services from=127.0.0.1:49463 latency=134.581µs
    2021-11-30T16:26:05.252-0600 [DEBUG] agent.http: Request finished: method=GET url=/v1/catalog/services from=127.0.0.1:49463 latency=94.975µs
```

### Versions
**Consul Terraform Sync**
```
consul-terraform-sync v0.4.2 (bc2b2a0)
Compatible with Terraform >= 0.13.0, < 1.1.0
```

**Consul Version**
```
Consul v1.9.5
```

### Configuration File(s)
```
task {
  name     = "catalog_task"
  services = ["api", "web"]
  source   = "./e2e/test_modules/local_tags_file"
  condition "catalog-services" {
    source_includes_var = true
    regexp              = "^api$"
  }
}
```

```
task {
  name   = "services_regex_task"
  source = "findkim/print/cts"
  condition "services" {
    regexp = "api.*"
  }
}
```

### Expected Behavior
Fewer requests from CTS to Consul for these conditions. The `index=XX` query parameter for the `/v1/catalog/services` requests would be present in Consul logs

### Actual Behavior
Constant API calls to Consul client

### Steps to Reproduce
1. Run CTS with either tasks above
2. View Consul logs

### Additional Context
The fix was listed in 0.4.2 ([changelog](https://github.com/hashicorp/consul-terraform-sync/blob/main/CHANGELOG.md#042-november-22-2021)) with https://github.com/hashicorp/consul-terraform-sync/pull/460 but I'm seeing different behavior

PR #535, opened 2021-12-02
Fix query types to use blocking queries

findkim:
### Bug 1
hcat would update the query index for blocking queries as expected since https://github.com/hashicorp/hcat/pull/73, but CTS was not utilizing it in the next API request to Consul and instead using a clean `hcat.QueryOptions` that didn't have reference of the iterating index.

Changes now use the updated query options by using it as the base and merging in the other query parameters. This was my bad missing this detail when I had attempted to fix it earlier https://github.com/hashicorp/consul-terraform-sync/pull/460

Resolves https://github.com/hashicorp/consul-terraform-sync/issues/529

### Bug 2
Fixing the blocking query exposed an existing race condition whether the API request for the health service or catalog-services would return within a [microsecond](https://github.com/hashicorp/hcat/blob/master/watcher.go#L211-L226).

To fix the race condition for multiple API calls pertaining to the same Consul service, there is an added conditional delay to account for service propagation time in a large Consul cluster.

Changes
* Refactored CTS to use new proposed watcher.Watch which provides template information and continuously processes new data.
* CTS now needs context of which hcat template IDs correspond to which task
* CTS checks template rendering more efficiently and only attempts when notified by watcher.Watch
* Removed unnecessary watcher.Watching calls which I mistakenly thought it accepted template IDs but actually expected view IDs which are irrelevant to the caller.
* Remove unnecessary references of `&sync.RWMutex{}` which made updating tests annoying.

### Bug 3
* ServicesRegexQuery was not properly setting the indexes to be used for blocking queries
* ServicesRegexQuery was returning non-passing health services

### DIFF
--- a/controller/readwrite.go
+++ b/controller/readwrite.go
@@ -76,10 +76,14 @@ func (rw *ReadWrite) Run(ctx context.Context) error {
 		rw.watcherCh = make(chan string, rw.drivers.Len()+2)
 	}
 	go func() {
-		err := rw.watcher.Watch(ctx, rw.watcherCh)
-		if err != nil {
-			rw.logger.Error("error watching template dependencies", "error", err)
-			errCh <- err
+		for {
+			rw.logger.Trace("starting template dependency monitoring")
+			err := rw.watcher.Watch(ctx, rw.watcherCh)
+			if err == nil || err == context.Canceled {
+				rw.logger.Info("stopping dependency monitoring")
+				return
+			}
+			rw.logger.Error("error monitoring template dependencies", "error", err)
 		}
 	}()
 
--- a/controller/readwrite.go
+++ b/controller/readwrite.go
@@ -69,26 +69,33 @@ func (rw *ReadWrite) Run(ctx context.Context) error {
 		}
 	}
 
+	errCh := make(chan error)
+	tmplCh := make(chan string, rw.drivers.Len()*2)
+	go func() {
+		err := rw.watcher.Watch(ctx, tmplCh)
+		if err != nil {
+			rw.logger.Error("error watching template dependencies", "error", err)
+			errCh <- err
+		}
+	}()
+
 	for i := int64(1); ; i++ {
-		// Blocking on Wait is first as we just ran in Once mode so we want
-		// to wait for updates before re-running. Doing it the other way is
-		// basically a noop as it checks if templates have been changed but
-		// the logs read weird. Revisit after async work is done.
 		select {
-		case err := <-rw.watcher.WaitCh(ctx):
-			if err != nil {
-				rw.logger.Error("error watching template dependencies", "error", err)
-				return err
+		case tmplID := <-tmplCh:
+			taskName, ok := rw.drivers.GetTask(tmplID)
+			if !ok {
+				rw.logger.Debug("template was notified for update but the template ID does not match any task", "template_id", tmplID)
+				continue
 			}
+
+			go rw.runTask(ctx, taskName) // errors are logged for now
+
+		case err := <-errCh:
+			return err
 
 		case <-ctx.Done():
 			rw.logger.Info("stopping controller")
 			return ctx.Err()
-		}
-
-		for err := range rw.runDynamicTasks(ctx) {
-			// aggregate collected errors and just log everything for now
-			rw.logger.Error("error running tasks", "error", err)
 		}
 
 		rw.logDepSize(50, i)
@@ -135,6 +142,46 @@ func (rw *ReadWrite) runDynamicTasks(ctx context.Context) chan error {
 	}()
 
 	return errCh
+}
+
+func (rw *ReadWrite) runTask(ctx context.Context, taskName string) error {
+	logger := rw.logger.With(taskNameLogKey, taskName)
+
+	d, ok := rw.drivers.Get(taskName)
+	if !ok {
+		return fmt.Errorf("driver for task %q does not exist to run", taskName)
+	}
+
+	if rw.drivers.IsActive(taskName) {
+		// The driver is currently active with the task, initiated by an ad-hoc run.
+		logger.Trace("task is active")
+		return nil
+	}
+
+	switch d.Task().Condition().(type) {
+	case *config.CatalogServicesConditionConfig:
+		// Catalog services condition has multiple API calls it depends on for
+		// updates. This adds an additional delay for hcat to process any new
+		// updates in the background that may be related to this task. 1 second is
+		// used to account for Consul cluster propagation of the change at scale.
+		// https://www.hashicorp.com/blog/hashicorp-consul-global-scale-benchmark
+		<-time.After(time.Second)
+	case *config.ScheduleConditionConfig:
+		// Schedule tasks are not dynamic and run in a different process
+		// In the future we may want to run a scheduled task ad-hoc
+		return nil
+	}
+
+	complete, err := rw.checkApply(ctx, d, true, false)
+	if err != nil {
+		logger.Error("error running task", "error", err)
+		return err
+	}
+
+	if rw.taskNotify != nil && complete {
+		rw.taskNotify <- taskName
+	}
+	return nil
 }
 
 // runScheduledTask starts up a go-routine for a given scheduled task/driver.
--- a/controller/readwrite.go
+++ b/controller/readwrite.go
@@ -86,13 +86,13 @@ func (rw *ReadWrite) Run(ctx context.Context) error {
 	for i := int64(1); ; i++ {
 		select {
 		case tmplID := <-rw.watcherCh:
-			taskName, ok := rw.drivers.GetTask(tmplID)
+			d, ok := rw.drivers.GetTaskByTemplate(tmplID)
 			if !ok {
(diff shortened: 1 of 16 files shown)


findkim:
Once the catalog services template is using blocking queries, it seems now to race against which dependencies are processed. This is the behavior I'm now seeing:
* If each services dep return before the cs dep, the behavior is as expected.
* If the cs dep does not return last, the template will render w/o some updates to services.

This edge case is most apparent when the catalog service overlaps with the input services. For example
```hcl
task {
  services = ["api", "web"]
    regexp = "api.+"
  }
}
```

and the default case
```hcl
task {
  services = ["api", "web"]
  condition "catalog-services" {} // where default value is based on task.services -- regexp = "^api$|^web$"
}
```

For the service `api` in the above examples, there are two API requests from CTS: catalog-services and health-service.

Apparently this race condition has existed but the end behavior was not common because of the fall-through catch of hcat processing additional data updates in microsecond batches ([source code](https://github.com/hashicorp/hcat/blob/master/watcher.go#L211-L226)). This was enough to catch the health service and catalog service update into one template render.

Blocking queries seem to now expose this race condition because the return and processing of the data runs at a different cadence. My local repro consistently has the cs return first and CTS detects that and goes through the template rendering + task execution, then it receives the health-service update. So the trigger doesn't include that simultaneous health-service update. The CircleCI test env seems to be nondeterministic causing flakey tests for `TestCondition_CatalogServices_SuppressTriggers`.

findkim:
The PR Isn't complete just yet, but I'm going to go ahead and mark it ready for review to get more eyes / initial feedback on the approach. If agreed, I can spend more time on test coverage

eikenb:
> * If each services dep return before the cs dep, the behavior is as expected.
> 
> * If the cs dep does not return last, the template will render w/o some updates to services.

I didn't think this could happen. The resolver should not return complete until all dependencies are fetched and used at least once. Order shouldn't matter. I'm sure I'm missing something here (working on consul-template again has its model more in my mind).. would you mind going a bit more into how this happens?

findkim:
Hey @eikenb the first time a template renders, the second case does not occur. The bug for this edge case is on subsequent  renders.

CTS has one template file that contains 2 dependencies -- catalog services, and health services. After both have already returned at least once, any update to either dep would be considered a change by the Watcher, and then the resolver would return complete.

Though where CTS behaves differently than pure-hcat/consul-template, is that CTS has a custom solution to actually suppress the notification from the Watcher for the health-service related changes, and have the Watcher only notify when the catalog-services dep changes. The code works as expected, but conceptually it doesn't play well with Consul in practice.

For a given service "web", CTS would have hcat make API requests to two endpoints `/v1/catalog/services` and `/v1/health/services`. They function independently and could return at different times based on how the cluster propagates changes or the change of health status. "web" can appear in the catalog first, but may not return from the `/v1/health/services` if we're filtering for healthy instances only. The behavior I'm trying to prevent for this CTS template is triggering when "web" appears in the catalog but without web's health information if the second request hasn't returned yet.

^ this is my current understanding of the bug and behavior. Lmk if this sparks any ideas if there's an alternate design we could move forward with.

mkam (approved):
These changes all make sense to me! I'll keep an eye out for any future changes you might add to this PR, though.

eikenb:
> They function independently and could return at different times based on how the cluster propagates changes or the change of health status. "web" can appear in the catalog first, but may not return from the `/v1/health/services` if we're filtering for healthy instances only. The behavior I'm trying to prevent for this CTS template is triggering when "web" appears in the catalog but without web's health information if the second request hasn't returned yet.

2 thoughts...

1) If the health is filtered to return healthy instances only and returns no healthy instances, how is that different from it getting the "web" service before it gets the health status (as both would seem to manifest as no instances)?

2) This really sounds like a dependency between the service and health. I think it'd be possible to model that dependency.. maybe have a containing Notifier object that wrapped the 2 and only notified when both had been updated or something like that.

findkim:
These are good questions, and I'm happy to discuss synchronously too to help build context or brainstorm as well.

The service making it to the catalog could appear at a different time (almost insignificant) than when the healthy service is available. But it seems to sometimes be greater than 1 microsecond because the hcat drain is not catching it consistently. I believe it appears in the catalog once the registration propagates throughout the cluster and Consul servers have consensus, and is not dependent on the health of it. There is [no health filtering support for that endpoint](https://www.consul.io/api-docs/catalog#list-services), which makes sense. For the healthy service to be available would depend on the health status and the propagation to that Consul agent CTS is querying, since consensus does not guarantee that 100% of the agents have that data, but the majority. This is my peripheral understanding of Consul, but it's a complex tool so I'm happy to learn more if there's a misunderstanding somewhere.

If that extra second results in no healthy services, then trigger away and result in no change. The difference is that with <1 second delay there could be information to update infrastructure with that would otherwise be silently missed.

> Notifier object that wrapped the 2 and only notified when both had been updated or something like that.

That would be a good solution if we wanted to limit this feature, but it's not always the case where health and service overlap. You could trigger on service of A and only want service details of B (this case is not effected by the bug). 

And then going back to your point about no healthy instances, yeah that's a valid scenario and in that case triggering w/o any healthy service would result in no change. This case would prevent us from coupling the dependencies like that.

go.mod
```diff
@@ -31,7 +31,7 @@ require (
 	github.com/hashicorp/go-syslog v1.0.0
 	github.com/hashicorp/go-uuid v1.0.2
 	github.com/hashicorp/go-version v1.3.0
-	github.com/hashicorp/hcat v0.2.1-0.20220105214302-4b15127c5629
+	github.com/hashicorp/hcat v0.2.1-0.20220112200956-444bf3805a7a
```
findkim, line 34: This is currently pointed to this branch https://github.com/hashicorp/hcat/pull/84 that was recently rebased

controller/readwrite.go
```diff
@@ -69,72 +70,70 @@ func (rw *ReadWrite) Run(ctx context.Context) error {
...
-		// the logs read weird. Revisit after async work is done.
 		select {
-		case err := <-rw.watcher.WaitCh(ctx):
-			if err != nil {
-				rw.logger.Error("error watching template dependencies", "error", err)
-				return err
+		case tmplID := <-rw.watcherCh:
+			d, ok := rw.drivers.GetTaskByTemplate(tmplID)
+			if !ok {
+				rw.logger.Debug("template was notified for update but the template ID does not match any task", "template_id", tmplID)
```
mkam, line 94: When would this scenario occur?

controller/readwrite.go
```diff
@@ -76,10 +76,14 @@ func (rw *ReadWrite) Run(ctx context.Context) error {
...
 	}
 	go func() {
-		err := rw.watcher.Watch(ctx, rw.watcherCh)
-		if err != nil {
-			rw.logger.Error("error watching template dependencies", "error", err)
-			errCh <- err
+		for {
+			rw.logger.Trace("starting template dependency monitoring")
+			err := rw.watcher.Watch(ctx, rw.watcherCh)
+			if err == nil || err == context.Canceled {
```
mkam, line 81: Just to make sure I understand, `rw.watcher.Watch()` would, in the typical case, not return unless there's an error because it's constantly watching the dependencies. If there is a non context canceled error, we want to instead of retry watching instead of returning. But if there is a case where we do want to stop watching the dependency or we're going through a shutdown where the context was canceled, then we exit. Is that accurate?

mkam (approved):
Looks good to me, left a few questions to better my understanding, though!

controller/readwrite.go
findkim: Yep! the idea is for it to not return in the typical case.

For the retry: Retry handling is [set by CTS](https://github.com/hashicorp/consul-terraform-sync/blob/main/controller/watcher.go#L70) and handled by hcat on a per-dependency case. This PR/main-as-is will have CTS retry until it maxes out the retry attempts ~15mins, and then the dependency would probably sit idle(?) and the tasks no longer trigger on it. I think the ideal case here would be for us to move towards high availability is to have indefinite retries with a max wait duration (instead of exponential).

Shut down where the context is canceled: yep!

Stop watching dependency: kind of, the context portion here would be the main program shutting down, and then would relay to the watcher to stop, stopping all dependencies ([source](https://github.com/hashicorp/consul-terraform-sync/blob/main/controller/controller.go#L76-L78)).

Stop watching individual dependencies: CTS task -> Terraform template -> multiple hcat dependencies; and different templates can have the same dependencies as another template. If we want to stop monitoring a subset of dependencies, CTS will need to remove the task's template from hcat, and then hcat uses that to determine if there are any other templates relying on it. If none, only then would hcat stop them, otherwise they will keep watching. I think it's the [`watcher.Sweep` method](https://pkg.go.dev/github.com/hashicorp/hcat#Watcher.Sweep)? I haven't looked into this too carefully yet since the delete task is a new feature.

Hope this helps! Let me know if there are other thoughts around this -- more questions or ideas

mkam: That was very helpful, thanks for explaining with such detail!

go.mod
findkim: Updated to master + updated Watcher interface https://github.com/hashicorp/consul-terraform-sync/pull/535/commits/8b18d90e96e875e3041fae26b8131b356d849454

Issue #522, opened 2021-11-24
Task Lifecycle API Enhancement


wilkermichael:
#### Description
Support new dynamic task lifecycle features including

- Create a new task
- Delete a task
- Execute a dry run/inspect run

PR #537, opened 2021-12-06
Feature 522 create task handler endpoint

wilkermichael:
This PR covers some of the work for creating a new task

### Change summary:
- Controller/controller.go does a lot of initial setup, to allow for this setup to be utilized later in the span of CTS's running lifespan, a new method is required which can create a task driver based on a task config. `createNewTaskDriver`
- Refactor controller to use `createNewTaskDriver` when initializing the controller when CTS starts up
- Create a handler for creating a task `POST /v1/tasks`. Upon receiving a request, the handler will:
1) Create a new task driver using the controllers `createNewTaskDriver` method
2) Initialize the task
3) Render the template
4) Apply the task if RunOption = "now"

### Examples:

**JSON request:**

```json
{
    "enabled": true,
    "name": "api-task",
    "providers": [
        "local"
    ],
    "services": [
        "api"
    ],
    "source": "./example-module"
}
```

**CURL Request**
```
curl --location --request POST 'http://localhost:8558/v1/tasks' \
--header 'Content-Type: application/json' \
--data-raw '{
    "enabled": true,
    "name": "api-task",
    "providers": [
        "local"
    ],
    "services": [
        "api"
    ],
    "source": "./example-module"
}'
```

**CURL Request With Run=now**
```
curl --location --request POST 'http://localhost:8558/v1/tasks?run=now' \
--header 'Content-Type: application/json' \
--data-raw '{
    "enabled": true,
    "name": "api-task",
    "providers": [
        "local"
    ],
    "services": [
        "api"
    ],
    "source": "./example-module"
}'
```

**Response**
```JSON
{
    "request_id": "5756036a-681c-df73-e6b9-ffb8d9b4ec0c",
    "task": {
        "buffer_period": {
            "enabled": true,
            "max": "25s",
            "min": "5s"
        },
        "condition": {
            "services": {
                "regexp": ""
            }
        },
        "enabled": true,
        "name": "api-task",
        "providers": [
            "local"
        ],
        "services": [
            "api"
        ],
        "source": "./example-module",
        "source_input": {
            "services": {
                "regexp": ""
            }
        },
        "version": "",
        "working_dir": "create-task/api-task"
    }
}
```

### Future Work for this release

**Handling Buffer Periods**
- Currently calling the drivers `SetBufferPeriod` method is not behaving as expected.
- Currently tasks created through the `/v1/tasks` will trigger on all changes and do not honour even the global buffer period

**Handling Variables**
- Design of create task endpoint requires `variables` field of type `map[string]string` be an optional field
- Currently this field will not be supported as the driver only allows for a variables files, not strings or maps of strings

part of #522

### DIFF
--- a/api/representation.go
+++ b/api/representation.go
@@ -128,3 +128,88 @@ func (tr taskRequest) String() string {
 	data, _ := json.Marshal(tr)
 	return fmt.Sprintf("%s", data)
 }
+
+type taskResponse oapigen.TaskResponse
+
+func taskResponseFromConfigTaskConfig(tc *config.TaskConfig, requestID oapigen.RequestID) taskResponse {
+	task := oapigen.Task{
+		Description: tc.Description,
+		Name:        *tc.Name,
+		Source:      *tc.Source,
+		Version:     tc.Version,
+		Enabled:     tc.Enabled,
+		WorkingDir:  tc.WorkingDir,
+	}
+
+	if tc.Providers != nil {
+		task.Providers = &tc.Providers
+	}
+
+	if tc.Services != nil {
+		task.Services = &tc.Services
+	}
+
+	task.SourceInput = new(oapigen.SourceInput)
+	switch si := tc.SourceInput.(type) {
+	case *config.ServicesSourceInputConfig:
+		task.SourceInput.Services = &oapigen.ServicesSourceInput{
+			Regexp: si.Regexp,
+		}
+	case *config.ConsulKVSourceInputConfig:
+		task.SourceInput.ConsulKv = &oapigen.ConsulKVSourceInput{
+			Datacenter: si.Datacenter,
+			Recurse:    si.Recurse,
+			Path:       *si.Path,
+			Namespace:  si.Namespace,
+		}
+	}
+
+	task.Condition = new(oapigen.Condition)
+	switch cond := tc.Condition.(type) {
+	case *config.ServicesConditionConfig:
+		task.Condition.Services = &oapigen.ServicesCondition{
+			Regexp: cond.Regexp,
+		}
+	case *config.CatalogServicesConditionConfig:
+		task.Condition.CatalogServices = &oapigen.CatalogServicesCondition{
+			Regexp:            cond.Regexp,
+			SourceIncludesVar: cond.SourceIncludesVar,
+			Datacenter:        cond.Datacenter,
+			Namespace:         cond.Namespace,
+			NodeMeta: &oapigen.CatalogServicesCondition_NodeMeta{
+				AdditionalProperties: cond.NodeMeta,
+			},
+		}
+	case *config.ConsulKVConditionConfig:
+		task.Condition.ConsulKv = &oapigen.ConsulKVCondition{
+			Datacenter:        cond.Datacenter,
+			Recurse:           cond.Recurse,
+			Path:              *cond.Path,
+			Namespace:         cond.Namespace,
+			SourceIncludesVar: cond.SourceIncludesVar,
+		}
+	case *config.ScheduleConditionConfig:
+		task.Condition.Schedule = &oapigen.ScheduleCondition{
+			Cron: *cond.Cron,
+		}
+	}
+
+	max := config.TimeDurationVal(tc.BufferPeriod.Max).String()
+	min := config.TimeDurationVal(tc.BufferPeriod.Min).String()
+	task.BufferPeriod = &oapigen.BufferPeriod{
+		Enabled: tc.BufferPeriod.Enabled,
+		Max:     &max,
+		Min:     &min,
+	}
+
+	tr := taskResponse{
+		RequestId: requestID,
+		Task:      &task,
+	}
+	return tr
+}
+
+func (tresp taskResponse) String() string {
+	data, _ := json.Marshal(tresp)
+	return fmt.Sprintf("%s", data)
+}
--- a/api/representation_test.go
+++ b/api/representation_test.go
@@ -429,3 +429,483 @@ func TestTaskRequest_ToConfigTaskConfig_Error(t *testing.T) {
 		})
 	}
 }
+
+func TestTaskResponse_String(t *testing.T) {
+	resp := taskResponse{
+		RequestId: "e9926514-79b8-a8fc-8761-9b6aaccf1e15",
+		Task: &oapigen.Task{
+			Name:    "task",
+			Source:  "test-source",
+			Version: config.String(""),
+			BufferPeriod: &oapigen.BufferPeriod{
+				Enabled: config.Bool(false),
+				Max:     config.String("0s"),
+				Min:     config.String("0s"),
+			},
+			Enabled: config.Bool(true),
+			Condition: &oapigen.Condition{
+				CatalogServices: &oapigen.CatalogServicesCondition{
+					Regexp:            config.String(".*"),
+					SourceIncludesVar: config.Bool(true),
+					Datacenter:        config.String("dc2"),
+					Namespace:         config.String("ns2"),
+					NodeMeta: &oapigen.CatalogServicesCondition_NodeMeta{
+						AdditionalProperties: map[string]string{
+							"key1": "value1",
+							"key2": "value2",
+						},
+					},
+				},
+			},
+			SourceInput: &oapigen.SourceInput{
+				Services: &oapigen.ServicesSourceInput{
+					Regexp: config.String(""),
+				},
+			},
+			WorkingDir: config.String("sync-tasks/task"),
+		},
+	}
+
+	actual := fmt.Sprintf("%s", resp)
+	expected := `{"request_id":"e9926514-79b8-a8fc-8761-9b6aaccf1e15",` +
+		`"task":{"buffer_period":{"enabled":false,"max":"0s","min":"0s"},` +
+		`"condition":{"catalog_services":{"datacenter":"dc2","namespace":"ns2",` +
+		`"node_meta":{"key1":"value1","key2":"value2"},"regexp":".*",` +
+		`"source_includes_var":true}},"enabled":true,"name":"task","source":"test-source",` +
+		`"source_input":{"services":{"regexp":""}},"version":"","working_dir":"sync-tasks/task"}}`
+	require.Equal(t, expected, actual)
+}
+
+func TestTaskResponse_taskResponseFromConfigTaskConfig(t *testing.T) {
+	cases := []struct {
+		name             string
+		taskConfig       *config.TaskConfig
+		expectedResponse taskResponse
+	}{
+		{
+			name: "minimum_required_only",
+			taskConfig: &config.TaskConfig{
+				Description:  config.String(""),
+				Name:         config.String("test-name"),
+				Providers:    []string{},
+				Services:     []string{"api", "web"},
+				Source:       config.String("test-source"),
+				VarFiles:     []string{},
+				Version:      config.String(""),
+				BufferPeriod: config.DefaultBufferPeriodConfig(),
+				Enabled:      config.Bool(true),
+				Condition:    config.DefaultConditionConfig(),
+				WorkingDir:   config.String("sync-tasks"),
+				SourceInput:  config.DefaultSourceInputConfig(),
+			},
+			expectedResponse: taskResponse{
+				RequestId: "e9926514-79b8-a8fc-8761-9b6aaccf1e15",
+				Task: &oapigen.Task{
+					Name:        "test-name",
+					Source:      "test-source",
+					Version:     config.String(""),
+					Description: config.String(""),
+					BufferPeriod: &oapigen.BufferPeriod{
+						Enabled: config.Bool(true),
+						Max:     config.String("20s"),
+						Min:     config.String("5s"),
+					},
+					Enabled: config.Bool(true),
+					Condition: &oapigen.Condition{
+						Services: &oapigen.ServicesCondition{
+							Regexp: config.String(""),
+						},
+					},
+					SourceInput: &oapigen.SourceInput{
+						Services: &oapigen.ServicesSourceInput{
+							Regexp: config.String(""),
+						},
+					},
+					WorkingDir: config.String("sync-tasks"),
+					Services:   &[]string{"api", "web"},
+					Providers:  &[]string{},
+				},
+			},
+		},
+		{
+			name: "basic_fields_filled",
+			taskConfig: &config.TaskConfig{
+				Description:  config.String("test-description"),
+				Name:         config.String("test-name"),
+				Providers:    []string{"test-provider-1", "test-provider-2"},
+				Services:     []string{"api", "web"},
+				Source:       config.String("test-source"),
+				VarFiles:     []string{},
+				TFVersion:    config.String(""),
+				Version:      config.String("test-version"),
+				BufferPeriod: config.DefaultBufferPeriodConfig(),
+				Enabled:      config.Bool(true),
+				Condition:    config.DefaultConditionConfig(),
+				WorkingDir:   config.String("sync-tasks"),
+				SourceInput:  config.DefaultSourceInputConfig(),
+			},
+			expectedResponse: taskResponse{
+				RequestId: "e9926514-79b8-a8fc-8761-9b6aaccf1e15",
+				Task: &oapigen.Task{
+					Name:        "test-name",
+					Source:      "test-source",
+					Version:     config.String("test-version"),
+					Description: config.String("test-description"),
+					BufferPeriod: &oapigen.BufferPeriod{
+						Enabled: config.Bool(true),
+						Max:     config.String("20s"),
+						Min:     config.String("5s"),
+					},
+					Enabled: config.Bool(true),
+					Condition: &oapigen.Condition{
+						Services: &oapigen.ServicesCondition{
+							Regexp: config.String(""),
+						},
+					},
+					SourceInput: &oapigen.SourceInput{
+						Services: &oapigen.ServicesSourceInput{
+							Regexp: config.String(""),
+						},
+					},
+					WorkingDir: config.String("sync-tasks"),
+					Services:   &[]string{"api", "web"},
+					Providers:  &[]string{"test-provider-1", "test-provider-2"},
+				},
+			},
+		},
+		{
+			name: "with_services_condition",
+			taskConfig: &config.TaskConfig{
+				Description:  config.String(""),
+				Name:         config.String("task"),
+				Providers:    []string{},
+				Services:     []string{},
+				Source:       config.String("test-source"),
+				VarFiles:     []string{},
+				Version:      config.String(""),
+				TFVersion:    config.String(""),
+				BufferPeriod: config.DefaultBufferPeriodConfig(),
+				Enabled:      config.Bool(true),
+				Condition: &config.ServicesConditionConfig{
+					config.ServicesMonitorConfig{
+						Regexp:             config.String("^web.*"),
+						Datacenter:         config.String(""),
+						Namespace:          config.String(""),
+						Filter:             config.String(""),
+						CTSUserDefinedMeta: map[string]string{},
+					},
+				},
+				WorkingDir:  config.String("sync-tasks"),
+				SourceInput: config.DefaultSourceInputConfig(),
+			},
+			expectedResponse: taskResponse{
+				RequestId: "e9926514-79b8-a8fc-8761-9b6aaccf1e15",
+				Task: &oapigen.Task{
+					Name:        "task",
+					Source:      "test-source",
+					Version:     config.String(""),
+					Description: config.String(""),
+					BufferPeriod: &oapigen.BufferPeriod{
+						Enabled: config.Bool(true),
+						Max:     config.String("20s"),
+						Min:     config.String("5s"),
+					},
+					Enabled: config.Bool(true),
+					Condition: &oapigen.Condition{
+						Services: &oapigen.ServicesCondition{
+							Regexp: config.String("^web.*"),
+						},
+					},
+					SourceInput: &oapigen.SourceInput{
+						Services: &oapigen.ServicesSourceInput{
+							Regexp: config.String(""),
+						},
+					},
+					WorkingDir: config.String("sync-tasks"),
+					Services:   &[]string{},
+					Providers:  &[]string{},
+				},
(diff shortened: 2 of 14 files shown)


api/task_create_test.go
```diff
@@ -0,0 +1,293 @@
...
+		return d, nil
+	}
+
+	c := TaskLifeCycleHandlerConfig{
+		store:               store,
+		drivers:             drivers,
+		bufferPeriod:        bufferPeriod,
+		workingDir:          workingDir,
+		createNewTaskDriver: createNewTaskDriver,
+	}
```
lornasong, line 259: optional suggestion - the dependencies created above are pretty straightforward such that the comments don't add much additional information. would you consider something like below? i find that's still fairly readable and reduces line count


```
	d := new(mocks.Driver) // this is the same
  	createNewTaskDriver := func(taskConfig config.TaskConfig) (driver.Driver, error) {
  		return d, nil
  	}
	c := TaskLifeCycleHandlerConfig{
  		store:               event.NewStore(),
  		drivers:             driver.NewDrivers(),
  		bufferPeriod:        config.DefaultBufferPeriodConfig(),
  		workingDir:          "sync-task",
  		createNewTaskDriver: createNewTaskDriver,
  	}
```

e2e/api_test.go
```diff
@@ -425,6 +441,180 @@ func TestE2E_TaskEndpoints_Delete_Conflict(t *testing.T) {
...
+	// 2. Create infrastructure change that would trigger new task
+	// 3. Create a new task with run=now
+	// 4. Check that the new task exists
+	// 5. Check that new task ran based on (2) infrastructure change
+
+	srv := testutils.NewTestConsulServer(t, testutils.TestConsulServerConfig{
+		HTTPSRelPath: "../testutils",
+	})
+	defer srv.Stop()
+	tempDir := fmt.Sprintf("%s%s", tempDirPrefix, "create_task")
```
lornasong, line 529: looks like the tempDir for the above task test is also `create_task` - we try to keep the tempDir-s unique

e2e/api_test.go
```diff
@@ -425,6 +441,180 @@ func TestE2E_TaskEndpoints_Delete_Conflict(t *testing.T) {
...
+		ID:      serviceName,
+		Name:    serviceName,
+		Address: "5.6.7.8",
+		Port:    8080,
+	}
+	testutils.RegisterConsulService(t, srv, service, defaultWaitForRegistration)
+
+	// Create task
+	u := fmt.Sprintf("http://localhost:%d/v1/tasks",
+		cts.Port())
```
lornasong, line 479: formatting nit: it looks like port is on the next line though it'll likely fit on this line. is there a reason?

e2e/api_test.go
```diff
@@ -425,6 +441,180 @@ func TestE2E_TaskEndpoints_Delete_Conflict(t *testing.T) {
...
+
+	createTaskRequest := fmt.Sprintf(createTestTaskTemplate, taskName, serviceName)
+
+	resp := testutils.RequestHTTP(t, http.MethodPost, u, createTaskRequest)
```
lornasong, line 483: this utils returns the response and expects the caller to close. you'll want to add a `defer resp.Body.Close()`. there's a couple others that below as well that I won't leave a comment on

e2e/api_test.go
```diff
@@ -425,6 +441,180 @@ func TestE2E_TaskEndpoints_Delete_Conflict(t *testing.T) {
...
+		Name:    serviceName,
+		Address: "5.6.7.9",
+		Port:    8080,
+	}
+	testutils.RegisterConsulService(t, srv, service, defaultWaitForRegistration)
+	time.Sleep(defaultWaitForNoEvent)
+	resp = testutils.RequestHTTP(t, http.MethodGet, s, "")
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+	resourcesPath = filepath.Join(tempDir, taskName, resourcesDir)
+	validateServices(t, true, []string{service.ID}, resourcesPath)
```
lornasong, line 510: The comment above says "Make a change that triggers the new task, expect an event" - it looks like the code here has an additional sleep "defaultWaitForNoEvent" and checks the resources (vs. the event count).

e2e/api_test.go
```diff
@@ -425,6 +441,180 @@ func TestE2E_TaskEndpoints_Delete_Conflict(t *testing.T) {
...
+		ID:      serviceName,
+		Name:    serviceName,
+		Address: "5.6.7.8",
+		Port:    8080,
+	}
+	testutils.RegisterConsulService(t, srv, service, defaultWaitForRegistration)
+
+	// Create task
+	u := fmt.Sprintf("http://localhost:%d/v1/tasks",
+		cts.Port())
```
lornasong, line 547: similarly here, would you mind sharing the reason for adding port on a separate line?

e2e/api_test.go
```diff
@@ -425,6 +441,180 @@ func TestE2E_TaskEndpoints_Delete_Conflict(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, http.StatusOK, resp.StatusCode, "response body %s", string(bodyBytes))
+
+	// Check that the task has been created
+	s := fmt.Sprintf("http://localhost:%d/%s/status/tasks/%s",
+		cts.Port(), "v1", taskName)
+	resp = testutils.RequestHTTP(t, http.MethodGet, s, "")
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+
+	// Verify that there was no update
```
lornasong, line 494: I initially took this comment to mean that the created task should not have been updated and was confused since we haven't updated the task yet. Maybe something like `// Verify that the task did not run` to clarify "update"?

e2e/api_test.go
```diff
@@ -425,6 +441,180 @@ func TestE2E_TaskEndpoints_Delete_Conflict(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, http.StatusOK, resp.StatusCode, "response body %s", string(bodyBytes))
+
+	// Check that the task has been created
+	s := fmt.Sprintf("http://localhost:%d/%s/status/tasks/%s",
+		cts.Port(), "v1", taskName)
+	resp = testutils.RequestHTTP(t, http.MethodGet, s, "")
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+
+	// Verify that there was an update
```
lornasong, line 563: Similarly here I initially misinterpreted this comment.  Maybe `// Verify that the task did run`?

e2e/api_test.go
```diff
@@ -425,6 +441,180 @@ func TestE2E_TaskEndpoints_Delete_Conflict(t *testing.T) {
...
+	resourcesPath := filepath.Join(tempDir, taskName, resourcesDir)
+	validateServices(t, true, []string{service.ID}, resourcesPath)
+}
+
+// TestE2E_TaskEndpoints_InvalidSchema tests the create task endpoint with an invalid schema, no task
+// should be created. This runs a Consul server and the CTS binary in daemon mode.
+//	POST /v1/tasks
+func TestE2E_TaskEndpoints_InvalidSchema(t *testing.T) {
```
lornasong, line 571: it looks like this test is mainly exercising the code that parses the request and checking that a task isn't created. could this possibly be done as a unit/integration test?

api/task_create.go
```diff
@@ -0,0 +1,110 @@
...
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	logger := logging.FromContext(r.Context()).Named(createTaskSubsystemName)
+	logger.Trace("create task request received, reading request")
+
+	// Decode the task request
+	var req taskRequest
+	requestID := requestIDFromContext(r.Context())
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		logger.Trace("bad request", "create_task_request", r.Body)
```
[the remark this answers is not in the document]
lornasong: Does the json error returned at line 26 have helpful information in it? If so, I think it would be useful to include in the log and error

api/task_create_test.go
```diff
@@ -0,0 +1,293 @@
...
+
+func TestTaskLifeCycleHandler_CreateTask_InternalError(t *testing.T) {
+	t.Parallel()
+	cases := []struct {
+		name       string
+		taskName   string
+		request    string
+		statusCode int
+		message    string
+		response   taskResponse
```
lornasong, line 164: Similarly here, `response` does not seem consumed by test cases

lornasong (commented):
excited for create task endpoint! left some questions and suggestions. if there are changes you agree with and make, would you mind initially adding it as a separate commit so that it will be easier for me to see what has changed?

one separate thought with regards to the first commit for converting the config - supporting openapi seems to be generating a lot of code to maintain. no action item, just noting a thought to keep track of.

api/representation.go
```diff
@@ -128,3 +128,88 @@ func (tr taskRequest) String() string {
...
+
+	if tc.Services != nil {
+		task.Services = &tc.Services
+	}
+
+	task.SourceInput = new(oapigen.SourceInput)
+	switch si := tc.SourceInput.(type) {
+	case *config.ServicesSourceInputConfig:
+		task.SourceInput.Services = &oapigen.ServicesSourceInput{
+			Regexp: si.Regexp,
```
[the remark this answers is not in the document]
wilkermichael: Yes, I noticed this after this work, I'll submit a separate PR for this

api/representation.go
```diff
@@ -128,3 +128,88 @@ func (tr taskRequest) String() string {
...
+		}
+	}
+
+	max := config.TimeDurationVal(tc.BufferPeriod.Max).String()
+	min := config.TimeDurationVal(tc.BufferPeriod.Min).String()
+	task.BufferPeriod = &oapigen.BufferPeriod{
+		Enabled: tc.BufferPeriod.Enabled,
+		Max:     &max,
+		Min:     &min,
+	}
```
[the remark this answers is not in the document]
wilkermichael: Yes correct, the plan is to revisit buffer period before the release. If the worst happens and we can't include it, then we will handle it at the schema level to reject bufferPeriods.

api/representation_test.go
```diff
@@ -429,3 +429,478 @@ func TestTaskRequest_ToConfigTaskConfig_Error(t *testing.T) {
...
+				Services: &oapigen.ServicesSourceInput{
+					Regexp: config.String(""),
+				},
+			},
+			WorkingDir: config.String("sync-tasks/task"),
+		},
+	}
+
+	actual := fmt.Sprintf("%s", resp)
+	expected := `{"request_id":"e9926514-79b8-a8fc-8761-9b6aaccf1e15","task":{"buffer_period":{"enabled":false,"max":"0s","min":"0s"},"condition":{"catalog_services":{"datacenter":"dc2","namespace":"ns2","node_meta":{"key1":"value1","key2":"value2"},"regexp":".*","source_includes_var":true}},"enabled":true,"name":"task","source":"test-source","source_input":{"services":{"regexp":""}},"version":"","working_dir":"sync-tasks/task"}}`
```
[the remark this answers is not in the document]
wilkermichael: yeah I can change this

e2e/api_test.go
wilkermichael: No reason, was because I copy/pasted from another test and removed a couple of the params. I'll fix this

api/representation.go
```diff
@@ -128,3 +128,88 @@ func (tr taskRequest) String() string {
...
 	return fmt.Sprintf("%s", data)
 }
+
+type taskResponse oapigen.TaskResponse
+
+func taskResponseFromConfigTaskConfig(tc *config.TaskConfig, requestID oapigen.RequestID) taskResponse {
+	task := oapigen.Task{
+		Description: tc.Description,
+		Name:        *tc.Name,
+		Source:      *tc.Source,
```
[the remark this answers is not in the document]
wilkermichael: I don't understand, aren't they different in a task?

```hcl
task {
  name           = "taskA"
  description    = ""
  enabled        = true,
  providers      = []
  services       = ["web", "api"]
  source         = "org/example/module"
  version        = "1.0.0"
  variable_files = []
    regexp = ".*"
  }
}
```

wilkermichael:
> one separate thought with regards to the first commit for converting the config - supporting openapi seems to be generating a lot of code to maintain. no action item, just noting a thought to keep track of.

I'm not sure if this actually adds extra overhead, for the current flow of creating a task, we'd have to convert a create task representation into a task config, no matter the source of the representation. 

An alternative is to add struct tags to the config object and unmarshall to that instead... but from experience, this becomes bloated and harder to manage as you juggle the configuration requirements and the endpoint requirements.

api/task.go
```diff
@@ -269,11 +235,58 @@ func decodeBody(body []byte) (UpdateTaskConfig, error) {
...
+		}
+	}
+
+	return nil
 }
 
-// runOption returns a run option for updating the task
-func runOption(r *http.Request) (string, error) {
+// getRunOption returns a run option for updating the task
+func getRunOption(r *http.Request) (string, error) {
```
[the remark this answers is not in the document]
wilkermichael: sure I can change it to `parseRunOption`. I changed this so it wouldn't conflict/be confused with the runOption variable

api/task_create.go
```diff
@@ -0,0 +1,110 @@
...
+		sendError(w, r, http.StatusBadRequest, fmt.Sprintf("error decoding the request"))
+		return
+	}
+	logger.Trace("create task request", "create_task_request", req)
+
+	// Check if task exists, if it does, do not create again
+	if _, ok := h.drivers.Get(req.Name); ok {
+		logger.Trace("task already exists", "task_name", req.Name)
```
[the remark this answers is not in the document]
wilkermichael: yeah, so my reasoning was that in this example, this isn't actually an error in CTS, this is an error with the request. So to me it wouldn't really make sense for CTS to log an error, as it did nothing wrong.

That's the reasoning I gave between logging a trace and error, basically if CTS returned an error ten I would log an error. Let me know what you think though, I'm flexible on this one.

api/task_create.go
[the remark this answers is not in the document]
wilkermichael: same comment as above

api/task_create.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	// Return the task response
+	resp := taskResponseFromConfigTaskConfig(tc, requestID)
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	err = json.NewEncoder(w).Encode(resp)
+	if err != nil {
+		logger.Error("error encoding json", "error", err, "execute_dryrun_response", resp)
```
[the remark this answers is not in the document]
wilkermichael: yes, you're right, this was leftover from the commented out code I used as the base for this. I'll change this.

api/task_create_test.go
```diff
@@ -0,0 +1,293 @@
...
+
+func TestTaskLifeCycleHandler_CreateTask(t *testing.T) {
+	t.Parallel()
+	cases := []struct {
+		name       string
+		taskName   string
+		request    string
+		run        string
+		statusCode int
+		response   taskResponse
```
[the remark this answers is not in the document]
wilkermichael: removing, didn't end up needing it, thanks for catching that

api/task_create_test.go
```diff
@@ -0,0 +1,293 @@
...
+			run:        "",
+			statusCode: http.StatusOK,
+		},
+		{
+			name:       "happy_path_run_now",
+			taskName:   testTaskName,
+			request:    testCreateTaskRequest,
+			run:        driver.RunOptionNow,
+			statusCode: http.StatusOK,
+		},
```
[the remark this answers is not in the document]
wilkermichael: Sure I can add this

api/task_create_test.go
```diff
@@ -0,0 +1,293 @@
...
+
+func TestTaskLifeCycleHandler_CreateTask_BadRequest(t *testing.T) {
+	t.Parallel()
+	cases := []struct {
+		name       string
+		taskName   string
+		request    string
+		statusCode int
+		message    string
+		response   taskResponse
```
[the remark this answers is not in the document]
wilkermichael: I'll remove

api/task_create_test.go
wilkermichael: yep, that's fine. This was more a checklist for me when I was creating this, but you're right, now that it's done it makes sense to refactor it.

e2e/api_test.go
wilkermichael: copy paste from another test, I'll fix these

e2e/api_test.go
wilkermichael: The invalid schema is checked by the generated code which requires setting up the routing, that's why I included it as an integration test

wilkermichael:
@lornasong I looked over your comments and added some replies, let me know what you think

I'll start working on fixes.

api/representation.go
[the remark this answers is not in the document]
lornasong: Sorry - this is a comment I left to myself and didn't catch I should delete when reviewing my comments. (It originally was a note to myself why the pointer was dereferenced whereas the other fields are not).

api/representation.go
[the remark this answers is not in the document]
lornasong: Gotcha - I'll jot these down and share them with you

api/representation.go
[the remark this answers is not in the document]
lornasong: To confirm my understanding of "reject bufferPeriods" does that cover not allowing buffer periods in the request as well as the response?

api/task.go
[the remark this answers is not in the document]
lornasong: Gotcha, makes sense

api/task_create.go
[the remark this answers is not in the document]
lornasong: Thanks for sharing. That's helpful to know. Your reason makes sense to me. I'm ok with keeping these at trace

api/task_create.go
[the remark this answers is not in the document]
lornasong: Thanks for the clarification on the trace-level. I had a second comment there as well regarding the json error - not sure what your thoughts were

e2e/api_test.go
lornasong: Not sure if you were planning to respond to this comment - to clarify, given the comment, it looks like the sleep() isn't needed and there could be a check for the events. Are you intending to check for events or is this just checking the updated resources?

e2e/api_test.go
lornasong: To clarify, you mean that it is why you made it an e2e test? Or do you mean that it can be an integration test?

lornasong:
@wilkermichael I replied or emoji-ed your responses. There was one comment not addressed - wasn't sure if it was intentional or not but left a follow up question on it.


> I'm not sure if this actually adds extra overhead. For the current flow of creating a task, we'd have to convert a create task representation into a task config, no matter the source of the representation.
> An alternative is to add struct tags to the config object and unmarshall to that instead... but from experience, this becomes bloated and harder to manage as you juggle the configuration requirements and the endpoint requirements.

I see your point about having to convert if we do api docs generation. I was weighing if we used config.TaskConfig as the api request. Just considering possibilities. I was wondering about struct tags, so it's helpful to hear your thoughts thoughts on that.

The source of my concern is that I'm planning to add a new field to condition/source_input services. Without reading your PR, I'm not sure I would have realized that there is config code in the api directory that I would have to update. However, as you're saying, this wouldn't be solved by using an alternative to `openapi` if we don't use struct tags.

api/representation.go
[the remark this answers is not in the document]
wilkermichael: I was thinking the request, but yes you're right we'd have to take out the buffer period in the response as well... 

Regardless, my intention is to have the buffer period be part of this release

e2e/api_test.go
wilkermichael: Sorry, yes, this could potentially be an integration test, my experience with e2e and integration tests in the past were that they could be synonymous (which they aren't here) so I was confused. I'll look into potentially making this an integration test.

api/task_create.go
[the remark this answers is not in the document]
wilkermichael: It could I guess, this one could be logged as an error

api/representation.go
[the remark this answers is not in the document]
lornasong: Sounds good. Thanks for confirming

e2e/api_test.go
lornasong: Gotcha. Yea, within CTS repo, we have some integration tests that aren't in our e2e directory and only exercise a targeted piece of code rather than running the binary. If the integration test isn't feasible or doesn't have less code/complexity, then don't worry about it

e2e/api_test.go
wilkermichael: I also did this for the other tests in this file which weren't calling close on the response bodies

e2e/api_test.go
wilkermichael: I was looking at the delete tests that were just written and misunderstood the sleep, I will fix this

e2e/api_test.go
wilkermichael: Thinking about this some more, to get this working as an integration test would require I setup the routing and handling in the test itself. Since the test is really to make sure that everything is setup properly (including the middleware that would reject this invalid request) the integration test wouldn't really verify that. It would just be testing that the generated code does what it's supposed to do, which isn't what my goal was. I'm going to leave this as an e2e test.

wilkermichael:
@lornasong this is ready for another pass, thanks again for all the feedback!

wilkermichael:
@lornasong I also realised that the create task endpoint should actually be returning a 201 (Status Created) instead of a 200. I've updated that with an additional commit

api/task_create_test.go
```diff
@@ -170,21 +170,29 @@ func TestTaskLifeCycleHandler_CreateTask_InternalError(t *testing.T) {
 			message:    "error initializing new task, mock error",
 			statusCode: http.StatusInternalServerError,
 		},
+		{
+			name:       "happy_path_run_now",
```
lornasong, line 174: the test name does not seem to reflect the test case (looks unhappy)

lornasong (approved):
looks good to me. thanks for all the changes. the separate commits made reviewing easier

api/task_create_test.go
wilkermichael: it is very unhappy

PR #538, opened 2021-12-06
Replace default condition / source_input block with empty

lornasong:
This set of changes is in preparation for new `names` field for condition / source_input services

**Context**
Currently we allow
1. a task to not be configured with a condition / source_input block, which defaults to a empty condition / source_input services block where `regexp = ""`
2. a task to be configured with an empty condition / source_input services where `regexp = ""`.

However, we will want to update condition / source_input services with a new field `names` such that either `regexp` or `names` must be configured. Therefore:
- The concept of a default condition / source_input no longer makes sense
- condition / source_input services can no longer be unconfigured

**Changes**
Commit 1: Introduce structs to configure an empty condition / source_input. Replace DefaultConditionConfig() & DefaultSourceInputConfig() with related Empty*()
Commit 2-5: Consume new Empty*() and update tests


config/condition_no.go
```diff
@@ -0,0 +1,39 @@
...
+// Validate validates the values and required options. This method is recommended
+// to run after Finalize() to ensure the configuration is safe to proceed.
+func (c *NoConditionConfig) Validate() error {
+	return nil
+}
+
+// GoString defines the printable version of this struct.
+func (c *NoConditionConfig) GoString() string {
+	return ""
+}
```
wilkermichael, line 39: If we leave `""` then condition and source input defaults print like this: `Condition:SourceInput:` seems a bit weird, maybe `"&Condition{&NoCondition{}}"` or print empty braces `{}`?

summary: return `"{}"` or `"&Condition{&NoCondition{}}"`

config/source_input_no.go
```diff
@@ -0,0 +1,39 @@
+package config
```
wilkermichael, line 1: Just a thought, since there aren't really any functional differences between source_input_no.go and condition_no.go, how would you feel about creating a struct called `NoMonitorConfig` instead? Since both SourceInput and Condition are composed of the Monitor type, you could set the empty values of both to `NoMonitorConfig{}`

eg.
```GO
func EmptySourceInputConfig() SourceInputConfig {
	return &NoMonitorConfig{}
}

func isSourceInputEmpty(c SourceInputConfig) bool {
	_, ok := c.(*NoMonitorConfig)
	return ok
}
```

```GO
func EmptyConditionConfig() ConditionConfig {
	return &NoMonitorConfig{}
}
```

config/testdata/long.json
```diff
@@ -103,11 +103,6 @@
             "key2": "value2"
           }
         }
-      },
-      "source_input": {
-        "services": {
-          "regexp": ""
-        }
```
wilkermichael, line 110: Good catch!

wilkermichael (commented):
Great work Lorna, really thorough. I like that you cleaned up the test cases that didn't make sense!

Just a few comments, and a suggested refactor, let me know what you think

config/condition_no.go
lornasong: Good point, I like the suggestion to return the empty braces. Thanks!

config/source_input_no.go
lornasong: Yea good point. I think the NoMonitorConfig is a great idea - reduces code to maintain

lornasong:
Added a commit with your feedback. Ready for re-review - let me know if you have any other thoughts. I'll merge this PR after your PR since it looks like I removed some functions that it uses

config/source_input.go
```diff
@@ -20,13 +20,13 @@ type SourceInputConfig interface {
 // EmptySourceInputConfig sets un-configured source inputs with a non-null
 // value
 func EmptySourceInputConfig() SourceInputConfig {
-	return &NoSourceInputConfig{}
+	return &NoMonitorConfig{}
 }
 
 // isSourceInputEmpty returns true if the provided SourceInputConfig `c` is
 // of type NoSourceInputConfig
```
wilkermichael, line 27: change comment to be type `NoMonitorConfig`

wilkermichael (approved):
Thanks for making those changes Lorna, only one very minor comment. Otherwise, looks great!

config/source_input.go
lornasong: Ah, thanks for catching!

lornasong:
@wilkermichael I added an additional commit to resolve latest on main in `representation_test.go`. Would you be able to review https://github.com/hashicorp/consul-terraform-sync/pull/538/commits/3c8fe95750924f629b2c1a7d0ee6048cbb42777c?

wilkermichael (approved):
Looks good Lorna

PR #539, opened 2021-12-08
Wait for event before checking files

mkam:
Passing Consul Compatibility: https://app.circleci.com/pipelines/github/hashicorp/consul-terraform-sync/2527/workflows/af50a61b-f781-4b2b-b190-e418cec2bca5/jobs/5609

### DIFF
--- a/e2e/compatibility/consul_test.go
+++ b/e2e/compatibility/consul_test.go
@@ -1,3 +1,4 @@
+//go:build e2e
 // +build e2e
 
 // $ go test -run TestCompatibility_Consul ./e2e/compatibility -tags=e2e -v
@@ -30,7 +31,8 @@ const (
 
 	nullTaskName = "null_task"
 
-	defaultWaitForAPI = 30 * time.Second
+	defaultWaitForAPI   = 30 * time.Second
+	defaultWaitForEvent = 8 * time.Second
 )
 
 func TestCompatibility_Consul(t *testing.T) {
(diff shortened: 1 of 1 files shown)


e2e/compatibility/consul_test.go
```diff
@@ -161,22 +163,30 @@ func testServiceInstanceCompatibility(t *testing.T, tempDir string, port int) {
 	testutils.CheckDir(t, false, webResourcesPath)
 
 	// 1. register db service instances
+	beforeRegister := time.Now()
```
wilkermichael, line 166: Just wondering if there is a big difference between just calling 
api.WaitForEvent(t, cts, "db_task", time.Now(), defaultWaitForEvent) rather than storing the time before doing the registration? Since we're not doing any validation on how long it takes, maybe it makes it a bit simpler?

wilkermichael (approved):
Looks good Melissa, one minor comment, whatever you decide is good!

e2e/compatibility/consul_test.go
```diff
@@ -161,22 +163,30 @@ func testServiceInstanceCompatibility(t *testing.T, tempDir string, port int) {
...
 	registerService(t, &capi.AgentServiceRegistration{ID: "db1", Name: "db"}, port)
 	testutils.CheckFile(t, true, dbResourcesPath, "db1.txt")
+	api.WaitForEvent(t, cts, "db_task", beforeRegister, defaultWaitForEvent)
```
mkam, line 169: Just realized this is in the wrong place, it should be before the CheckFile!

e2e/compatibility/consul_test.go
mkam: Not a big difference, it's all down to timing. That value is used by WaitForEvent for checking for an event that starts after the given time. To me, it's safer to have this now value set before the Consul registration on the off chance an event starts between the `registerService()` and the `api.WaitForEvent()`. But you could also argue that we don't want an event to happen between the `time.Now()` and when the service is finished registering in Consul!

Either way, I don't think we've ever had any issue with the placement of time.Now(), so I'm fine to change it as an inline variable to simplify things!

mkam: I looked at the other tests and for the most part, we do the check beforehand, so I've left it as is!

PR #543, opened 2021-12-08
Add `names` config field to `condition "services"` and `source_input "services"`

lornasong:
Add new `names` config. Examples:

```
source_input "services" {
  names = ["api", "web"]
}
```
```
condition "services" {
  names = ["db"]
}
```

Changes:
- Add support regexp="" as a valid regex (needed to determine if regexp is configured or not)
- Add names field
- Update `condition "services"` and `source_input "services"` relationship with `task.services` list such that condition nor source_input can be configured when services list is configured
  - This is backwards breaking but in a non-meaningful way
  - Users were previously not permitted to configure condition/source_input services with a regexp along with task.services
  - With this change, users can no longer configure an empty condition/source_input services
     ```
     task {
       name = "task"
       description = "example of empty cond with services - no longer permitted"
       services = ["api"]
       condition "services {
       }
     }
     ```

### DIFF
--- a/config/monitor_services.go
+++ b/config/monitor_services.go
@@ -102,15 +102,14 @@ func (c *ServicesMonitorConfig) Merge(o MonitorConfig) MonitorConfig {
 	return r2
 }
 
-// Finalize ensures there no nil pointers.
+// Finalize ensures there no nil pointers. with the _exception_ of Regexp. There
+// is a need to distinguish betweeen nil regex (unconfigured regex) and empty
+// string regex ("" regex pattern) at Validate()
 func (c *ServicesMonitorConfig) Finalize([]string) {
 	if c == nil { // config not required, return early
 		return
 	}
 
-	if c.Regexp == nil {
-		c.Regexp = String("")
-	}
 	if c.Datacenter == nil {
 		c.Datacenter = String("")
 	}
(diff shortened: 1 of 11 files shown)


config/monitor_services.go
```diff
@@ -102,14 +112,16 @@ func (c *ServicesMonitorConfig) Merge(o MonitorConfig) MonitorConfig {
 	return r2
 }
 
-// Finalize ensures there no nil pointers.
+// Finalize ensures there no nil pointers. with the _exception_ of Regexp. There
+// is a need to distinguish betweeen nil regex (unconfigured regex) and empty
+// string regex ("" regex pattern) at Validate()
```
findkim, line 117: When would a user want to use `""` as a regular expression for service names? It seems to only match empty string service names, which Consul doesn't support.

lornasong: The empty string regular expression acts similar as a wildcard, it matches on everything. You can try it out [here](https://go.dev/play/p/MsTenXQeQnw)

findkim: Ahh that's good to know, thank you for checking me!

findkim (approved):

config/monitor_services.go
findkim: TIL! I've been doing `.*` 😂

lornasong: Same here! The actual behavior of "" isn't really what I would have guessed

lornasong:
Rebased on main. Resolved a conflict in `representation_test.go`

PR #544, opened 2021-12-09
Refactor templates directory to use Template struct

lornasong:
Current state of templating:
 - Organized by how the templating was configured i.e. `condition` block, `source_input` block  , and `services` list
 - Example: we have `condition_services.go` and `source_input_services.go`. Their shared logic is in `monitor_services.go`
   - `source_input_services.go` is essentially the same as `monitor_services.go`
   - `condition_services.go` is similar to `monitor_services.go` but handles additional logic around source_includes_var

Context for refactor:
 - `condition_services.go` has templating logic for the `{{ servicesRegex }}` template.
 - we will be supporting `names` field in `condition "services"` which is very different from `{{ servicesRegex }}`. It will be more similar to `services` list
 - It doesn't make sense to have `names` and `regexp` templating handled by the same structure just because they are configured by the same type of condition block

Work:
 - This PR: refactor files to consolidate by template-type (e.g. refactor condition* and source_input* into monitor*)
 - Future work:
   - Rename `monitor_services` to `monitor_services_regex` (since it currently handles servicesRegex tmpl)
   - Add a new `monitor_services` to handle `{{ services }}` template used by services list and names

Commits:
 - 1: Remove source_input_* files
 - 2-4: Consolidate condition_* files into monitor_* files
 - 5: Rename a directory to remove "condition"
 - 6: Update + refactor for new usage of monitor
 - 7: Update tests


templates/tftmpl/variables.go
```diff
@@ -57,26 +57,24 @@ func newVariablesTF(w io.Writer, filename string, input *RootModuleInputData) er
...
 	}
 
-	if input.Condition != nil && input.Condition.SourceIncludesVariable() {
-		if err = input.Condition.appendVariable(w); err != nil {
-			return err
-		}
-	}
+	// append a variable for each monitored object
+	// note: assumes monitored objects all have a unique type. otherwise would
+	// need to check to avoid appending duplicate variables
```
mkam, line 67: Thanks for the callout here!

templates/tftmpl/variables.go
```diff
@@ -57,26 +57,24 @@ func newVariablesTF(w io.Writer, filename string, input *RootModuleInputData) er
...
+	for _, monitor := range input.Monitors {
+		if monitor.SourceIncludesVariable() {
+			if monitor.ServicesAppended() {
+				// services variable is already appended earlier. skip
+				continue
```
mkam, line 72: Just want to make sure I understand this properly, this loop isn't just for the `services` variable, but if the monitor does happen to have `ServicesAppended` set to true, that means the monitor cares about `services` and no further action needs to be taken since we've already handled `services`. Is that correct?

mkam (approved):
This looks good to me! PR and commit descriptions were very helpful, and I think this is a really clean refactor.

I did want to throw out this question to which I don't have an opinion one way or another! Are we still good with the term `monitor`? Just wanted a gut check on it since the functionality has changed slightly, though I also don't have any good alternatives to offer up.

lornasong:
> Are we still good with the term monitor? Just wanted a gut check on it since the functionality has changed slightly

Good question! I was thinking that the way I would describe the (currently named) Monitor* structures is something like "the structs that handle templating-related things for the info that CTS monitors". So I felt that "Monitor" still kind of fit. Agree though - this departs from the way monitor is used in the config. I considered "Template" as well but the struct does a little more than append templates. Let me know if you have thoughts!

That reminds me - I'll also update the comment on the interface and structures according to what we discuss

templates/tftmpl/variables.go
lornasong: Yes, your understanding is correct! I think the name ServicesAppended() is a little confusing - ServicesAppended() will only returns true for MonitorServices (and never for the others). I will update the comment for the loop to clarify the `services` variable exception. Maybe I can also rename `ServicesAppended` to something like `IsServicesVar` to help clarify? Let me know if you have thoughts!

mkam: Ooh, I think `IsServicesVar` is a good rename, that helps a lot!

lornasong:
Hmmm. I thought about your "Monitor" naming questions some more. My description isn't quite right since we'll end up with a ServicesMonitor and ServicesRegexMonitor. I think the description should be "the structs that handle templating-related things for a given template function e.g. `{{ services }}` and `{{ servicesRegex }}`"

We've been abbreviating template function as "tmplfunc". What are your thoughts on "TmplFunc" e.g. "ServicesTmplFunc", "ServicesRegexTmplFunc"?

mkam:
TmplFunc also makes sense in this context! Will to be confusing, though, with what we have in [templates/tftmpl/tmplfunc](https://github.com/hashicorp/consul-terraform-sync/tree/main/templates/tftmpl/tmplfunc)?

lornasong:
🤔 good point - I agree it would be confusing

My latest thought is to go with "Template" and refactor off `appendModuleAttribute()` in a separate PR. It's the one non-"template" function and I think it will help some TFC-side template refactoring I'm working on. With that, I think "Template" will be more accurate. Then the description could be "Template handles templates for different types of template functions and variables". Let me know if you have thoughts!

mkam:
That sounds good to me! And feel free to add that refactor to this PR!

lornasong:
Appreciate all this discussion with me @mkam !

I added the commit: https://github.com/hashicorp/consul-terraform-sync/pull/544/commits/b31fa6ed7c9ca7d67927f871b320c2ec6d2b1079 which renames "Monitor" to "Template". It's ready for review now. Thank you 🙏 

Sleeping on it, I decided not to do the refactor to get rid of the `appendModuleAttribute`. I realized there is also an `appendVariables` and it seemed more straight forward to find all the "append" methods on the Template struct even though these are only template-related methods. If you have concerns, let me know!

mkam:
Did you want to rename the files as well to have template over monitor? If not, feel free to merge!

PR #545, opened 2021-12-10
Enable Security Scanning (CRT)

claire-labry:
What this PR does is it enables security scanning (owned by the Security team) for this repository. This is a part of the Common Release Tooling (CRT) effort which `enables a consistent workflow for security scanning that can be used across products and services. It provides a single abstraction point to allow us integrate our own scanning capabilities, as well as leverage external tools and APIs from vendors, open source, or free-to-use projects.` Learn more about HashiCorp's security-scanner [here](https://github.com/hashicorp/security-scanner/).

While running the security scan tests, I ran into several [false positives](https://github.com/hashicorp/crt-workflows-common/runs/4472723819?check_suite_focus=true) with `secrets` in the [binary](https://github.com/hashicorp/consul-terraform-sync/blob/51414c677cbe369db24043e8d2f84ba502b7a58e/.release/security-scan.hcl#L8) and [container](https://github.com/hashicorp/consul-terraform-sync/blob/51414c677cbe369db24043e8d2f84ba502b7a58e/.release/security-scan.hcl#L4) stanzas in the security-scan.hcl file. This is a known issue with Security and they suggested me to set these to `false` aka off for the time being. Once Security fixes this, I will test it again and raise a new PR to turn those back on to `true`

You'll also see that I had to upgrade the Dockerfile to use `alpine:3.15.0` as the older version caused the security-scanner to pick up some vulnerabilities, CVEs to be specific. See error outputs [here](https://github.com/hashicorp/crt-workflows-common/runs/4473388427?check_suite_focus=true). We worked with Security and they scanned different versions of the alpine base image and concluded that 3.15.0 was clean and suggested CTS to upgrade to that version to avoid any friction and potential vulnerabilities in the future and simply tidy up the container. 

With that being said, with the alpine version being updated and the secret scanning turned off (for now) the tests with security-scan are [passing](https://github.com/hashicorp/crt-workflows-common/actions/runs/1560728556) at the verify step which is the last workflow event. 

If anyone needs any further clarification or if nothing makes sense, feel free to comment or reach out to me! ❤️

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,7 +3,7 @@ name: build
 on:
   push:
     # Sequence of patterns matched against refs/heads
-    branches:   
+    branches:
       # Push events on main branch
       - main
       # Push events to branches matching refs/heads/release/**
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,11 +3,13 @@ name: build
 on:
   push:
     # Sequence of patterns matched against refs/heads
-    branches:    
-      # Push events on main branch
-      - main
-      # Push events to branches matching refs/heads/release/**
-      - 'release/**'
+    branches:   
+      # push to enable-security-scan for testing
+      - enable-security-scan 
+      # # Push events on main branch
+      # - main
+      # # Push events to branches matching refs/heads/release/**
+      # - 'release/**'
 
 env:
   PKG_NAME: "consul-terraform-sync"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,12 +4,10 @@ on:
   push:
     # Sequence of patterns matched against refs/heads
     branches:   
-      # push to enable-security-scan for testing
-      - enable-security-scan 
-      # # Push events on main branch
-      # - main
-      # # Push events to branches matching refs/heads/release/**
-      # - 'release/**'
+      # Push events on main branch
+      - main
+      # Push events to branches matching refs/heads/release/**
+      - 'release/**'
 
 env:
   PKG_NAME: "consul-terraform-sync"
--- a/.release/ci.hcl
+++ b/.release/ci.hcl
@@ -4,7 +4,7 @@ project "consul-terraform-sync" {
   team = "consul api tooling"
   slack {
     # feed-consul-api-gh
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.13 as default
+FROM alpine:3.15.0 as default
```
eikenb, line 1: A few months ago we converted all our team's docker images to use `FROM alpine:3` so they'd always use the latest stable release (prompted by this same issue). I'd suggest copying that pattern unless there is a reason not to... and if there is I'd be curious to know so as to evaluate whether it applies to the other projects. Thanks.

sarahethompson (approved):

.github/workflows/build.yml
```diff
@@ -3,7 +3,7 @@ name: build
 on:
   push:
     # Sequence of patterns matched against refs/heads
-    branches:    
+    branches:
```
wilkermichael, line 6: Can you undo the change to this file since it's just a spacing change?

.release/ci.hcl
```diff
@@ -4,7 +4,7 @@ project "consul-terraform-sync" {
   team = "consul api tooling"
   slack {
     # feed-consul-api-gh
-    notification_channel = "C026W707YHJ"
+    notification_channel = "C026W707YHJ"
```
wilkermichael, line 7: Same as above, we can just remove this change since it's just a spacing change

wilkermichael (approved):
Thanks for doing this, all good from our team

Dockerfile
claire-labry: Applied that change! Thank you for recommending that improvement 💯

claire-labry:
I'm getting ready to merge this PR but noticed the `ci/circleci: e2e_tests` is failing (has been re-ran a couple of times) and I'm not sure if I know enough about it to understand why it might be failing. Some guidance in helping me better understand what this is for would be really helpful as I don't want to merge in this PR if this is something that needs to be fixed on my end. @wilkermichael

wilkermichael:
@claire-labry I suspected their might have been a fix for the e2e tests already, so just wanted to confirm there was nothing else going on in this branch.

I rebased onto the current main branch, everything looks good now

PR #547, opened 2021-12-10
Feature 522 handle create task variables

wilkermichael:
With the addition of the create task endpoint, there is now two ways to provide variables to CTS for a given task
1) variable_files of type list(string) (as part of the config)
2) variables of type map[string]string (from a create task request)

Example request with variables for context:
```json
{
    "enabled": true,
    "name": "api-task",
    "providers": [
        "local"
    ],
    "services": [
        "api"
    ],
    "variables":{
        "filename": "\"set.txt\"",
        "object": "{user_meta = {name = \"michael\"\n age = \"32\"\n day_of_the_week = \"Friday\"}\n tags = [\"happy\", \"glad\"]}"
    },
    "source": "./variables-module"
}
```

This PR serves to provide a common processing path for these variables sources so that they can be used by the terraform driver. All variables are processed and rendered to a `variables.tfvars` file in the working directory, and this file is then provided to the terraform driver.

I've performed manual tests and everything seems to be working, but haven't written any e2e tests. I didn't see tests for variables currently (correct me if I'm wrong), so might be a bigger task to write the e2e tests. Can do this in a future PR.

part of #522


driver/task.go
```diff
@@ -249,16 +265,19 @@ func (t *Task) Source() string {
...
+// for the task's module as a list.
 func (t *Task) VariableFiles() []string {
 	t.mu.RLock()
 	defer t.mu.RUnlock()
 
-	varFiles := make([]string, len(t.varFiles))
-	for i, vf := range t.varFiles {
-		varFiles[i] = vf
+	// In the future may be worth refactoring for Terraform CLI to only expect a
+	// single varFile, for now this is useful
```
findkim, line 275: Changing existing behavior that Terraform CLI and CTS supports would be breaking change that doesn't offer additional value. Could the API support multiple files instead?

It provides users more flexibility to organize their variables, like one file contains more sensitive information, and the other could act more like parameters.

templates/tftmpl/root.go
```diff
@@ -41,11 +41,15 @@ const (
 	// variable is written to.
 	TFVarsFilename = "terraform.tfvars"
 
+	// VarsTFVarsFileName is the file name for a tfvars file which is generated and contains
+	// variables provided as part of the task configuration
+	VarsTFVarsFileName = "variables.tfvars"
```
findkim, line 46: What's the context behind "variables.tfvars" filename? Could the filename be preserved from what was provided from the config/payload for debugging purposes?

driver/task.go
wilkermichael: So CTS still supports multiple file input, it just writes out a single file to the working directory, and that single file is fed to terraform CLI. So from a user perspective they can still organize the files any way they want eg.

var_file_1  ------>
var_file_2 ------>       variables.tfvars
var_file_3 ------>

templates/tftmpl/root.go
wilkermichael: If we want to support multiple variables files in the working directory then we can do that, otherwise the name came from `variables.module.tf` and `variables.tf` which are related to this file.

driver/task.go
wilkermichael: So I don't see this as a breaking change I guess, nothing changes in the configuration or input to CTS

findkim: Based on the example request in the PR summary, it seems like this is the beginning of diverging UX between config file and API where only one file is supported, and this comment leads to suggest breaking change. Which way is it?

```json
     "variables":{
        "filename": "\"set.txt\"",
        "object": "..."
      }
```

templates/tftmpl/root.go
findkim: What's the thought behind collapsing user input variables into one file?

driver/task.go
wilkermichael: maybe a poor example choice on my part, I had built an example module on top of the simple example module that's part of the CTS repo. Filename in this context doesn't indicate a file for input, it's a key-value (it's the name of the file that the terraform module creates and write service information to)

The API just support key-values, it doesn't have any concept of files

templates/tftmpl/root.go
wilkermichael: So for TFC, the variables for all files just get loaded into a single list of parsed key values, and this list gets fed to TFC. My thought was to mirror this behaviour, where instead of feeding variables to the terraform module, I instead took that same list of variables and wrote a file to the working space. This file would then be used as input to the terraform driver, much in the same way the list is used as input to the TFC driver

driver/task.go
findkim: Okay that example was pretty confusing especially in addition to changing the `Task.varFiles` to a singular file. Have you double checked if this change behaves correctly for the existing use-case for loading a [task from the config file that contains multiple variable files](https://www.consul.io/docs/nia/configuration#variable_files). This now gets ignored https://github.com/hashicorp/consul-terraform-sync/blob/main/driver/task.go#L82

findkim: Chatted and this use-case is covered with the existing code path that loads variables from the files during initialization. A follow-up non-blocking refactor would be to look into how the `variables.module.tf` file is generated and couple it to this new direction for its `.tfvars`. Maybe the `TaskConfig.varFiles`  could be removed completely at this point.

templates/tftmpl/root.go
findkim: Cool thank for sharing. A single file maybe simpler to manage for future task updates too

templates/tftmpl/tfvars.go
```diff
@@ -79,6 +79,26 @@ func newProvidersTFVars(w io.Writer, filename string, input *RootModuleInputData
 	return err
 }
 
+// newVariablesTFVars writes input variables for configured variables.
+func newVariablesTFVars(w io.Writer, filename string, input *RootModuleInputData) error {
```
findkim, line 83: A full on e2e isn't required, but unit tests should be added. Here are some [existing unit tests with golden files that could be used as reference](https://github.com/hashicorp/consul-terraform-sync/blob/64f9d1d9f9990617bfb05c3000f46de0388287fa/templates/tftmpl/golden_test.go#L501 )

templates/tftmpl/root.go
findkim: Not sure if this was already considered, would the `*auto.tfvars` naming scheme be useful for this feature? https://www.terraform.io/docs/language/values/variables.html#variable-definition-precedence

wilkermichael: I hadn't come across this, I can add the *.auto* to the tfvars file if that's the convention.

templates/tftmpl/golden_test.go
```diff
@@ -458,13 +477,83 @@ func TestNewFiles(t *testing.T) {
...
+// not suffice.
+func TestVariablesTFVarsDetailed(t *testing.T) {
+	task := Task{
+		Description: "user description for task named 'test'",
+		Name:        "test",
+		Source:      "namespace/consul-terraform-sync/consul//modules/test",
+		Version:     "0.0.0",
+	}
+
+	testCases := []struct {
```
findkim, line 499: Does this have to be in a table?

templates/tftmpl/root.go
```diff
@@ -41,11 +41,15 @@ const (
 	// variable is written to.
 	TFVarsFilename = "terraform.tfvars"
 
+	// VarsTFVarsFileName is the file name for a tfvars file which is generated and contains
+	// variables provided as part of the task configuration
+	VarsTFVarsFileName = "variables.auto.tfvars"
```
findkim, line 46: If `auto.tfvars` is used as the filename format now, we may not need to even track `varFile` anymore and pass that through to the TF CLI command since it should automatically discover it in the working directory.

findkim (approved):
Thanks for adding tests!

wilkermichael:
@findkim I added three new commits.
- Addressed review comment
- Added golden test
- terraform driver/CLI refactor

I decided to do the terraform driver/cli refactor to take a single file rather than a list. I figured it would be easier for review since you have the most context now, but let me know if you'd like me to throw it in a separate PR

templates/tftmpl/root.go
wilkermichael: Oh, interesting, I will give this a try

templates/tftmpl/golden_test.go
wilkermichael: I ended up removing this test now that I sort the keys I can do the check in the TestNewFiles

client/terraform_cli.go
```diff
@@ -86,27 +85,27 @@ func NewTerraformCLI(config *TerraformCLIConfig) (*TerraformCLI, error) {
 	}
 
 	// Expand any relative paths for variable files to absolute paths
-	varFiles := make([]string, 0, len(config.VarFiles))
-	for _, vf := range config.VarFiles {
-		if !filepath.IsAbs(vf) {
+
+	var varFile string
```
wilkermichael: I removed this in the last commit

client/terraform_cli.go
```diff
@@ -43,7 +42,7 @@ type TerraformCLIConfig struct {
 	ExecPath   string
 	WorkingDir string
 	Workspace  string
-	VarFiles   []string
+	VarFile    string
```
wilkermichael, line 43: This can be removed

client/terraform_cli.go
```diff
@@ -86,27 +85,27 @@ func NewTerraformCLI(config *TerraformCLIConfig) (*TerraformCLI, error) {
...
 			}
-			vfAbs := filepath.Join(wd, vf)
-			varFiles = append(varFiles, vfAbs)
+			vfAbs := filepath.Join(wd, config.VarFile)
+			varFile = vfAbs
 		} else {
-			varFiles = append(varFiles, filepath.Clean(vf))
+			varFile = filepath.Clean(config.VarFile)
 		}
 	}
```
wilkermichael: I removed this in the last commit

findkim (approved):
After talking, updating the `providers.tfvars` to `providers.auto.tfvars` cleanup works to remove the need for track var files completely.

The changes look good overall and going ahead and re-approving so you can merge whenever the tests are passing again ✅

wilkermichael:
@findkim thanks Kim for the review, I appreciate the time you put into it. I know there was a lot of back and forth on it.

PR #549, opened 2021-12-15
Rename ServicesTemplate to ServicesRegexTemplate

lornasong:
The current ServicesTemplate supports the template function `{{ servicesRegex }}`.

I plan to create a Template that will support the template function `{{ service }}` for new `names` field for `condition "services"` (and refactor services list to use it too)

Therefore, renaming the current ServicesTemplate to ServicesRegexTemplate in this PR. Upcoming PR to add back ServicesTemplate but for `{{ service }}` tmplfunc


wilkermichael (approved):
Looks good @lornasong

PR #551, opened 2021-12-15
Update ECR tag in CRT Builds

smacfarlane:
This updates the ECR tag to the correct naming convention. 

Signed-off-by: Scott Macfarlane <smacfarlane@hashicorp.com>


lornasong (approved):
thanks @smacfarlane!

PR #552, opened 2021-12-16
Feature 522 generate client refactor delete command

wilkermichael:
Scope of this feature is just to refactor delete command to use new generated client.

Part of #522


api/task_lifecycle_client.go
```diff
@@ -0,0 +1,127 @@
...
+// Do performs the http request and parses the response to report any errors
+func (d *TaskLifecycleHTTPClient) Do(req *http.Request) (*http.Response, error) {
+	resp, err := d.http.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	// defer resp.Body.Close() not called for happy path, only called for
+	// unhappy path. caller of this method will close if returned err == nil.
+
+	if resp.StatusCode != http.StatusOK {
```
mkam, line 109: Could we have something more general, like checking that the resp.StatusCode is in the 2XX range? That way we don't have to update the if statement for different possible happy path response codes.

api/task_lifecycle_client.go
```diff
@@ -0,0 +1,127 @@
...
+		defer resp.Body.Close()
+
+		var errResp ErrorResponse
+		decoder := json.NewDecoder(resp.Body)
+		if err = decoder.Decode(&errResp); err != nil {
```
mkam, line 114: I noticed when there's a schema error (like not providing a required field), there is a plaintext response like:
```
request body has an error: doesn't match the schema: Error at "/name": property "name" is missing
```
I'm thinking that will cause a problem here and obstruct the error because we there'll be a decoding issue. Ideally it would be nice if we could wrap these schema issues into a JSON form, but not sure if that support is available.

api/task_lifecycle_client.go
wilkermichael: good point! As we talked about this is copied directly from https://github.com/hashicorp/consul-terraform-sync/blob/7592b852f526c0b7e95a56e275f12cecdaf65d02/api/client.go#L230 

For this PR I'm thinking I'll just submit the refactor of delete, so I'll leave it as is. When I work on my create command feature however, I'll have to make the change since it's returning a 201

api/task_lifecycle_client.go
wilkermichael: Yeah this is provided by the openapi swagger schema validator middleware... I'll look into if there is a way to wrap it. We can always remove this middleware as well if we find it's more harmful than good.

wilkermichael:
@mkam I'm going to bump this up for review now

api/task_lifecycle_client.go
wilkermichael: Will tackle in a different feature

mkam (approved):
Looks good to me, thanks for updating the README with the clarification!

Issue #425, opened 2021-10-07
Fix plan output readability for TFC driver


lornasong:
### Describe the bug

When a disabled task is re-enabled through the Enable CLI or CTS is executed in inspect mode, the CLI outputs information from a Terraform Plan to inform the user of the changes that CTS will make when the task becomes enabled.

For the TFC driver, the current CLI output looks different from when it was originally implemented if the task has a Terraform version of `0.15.2` or greater. This current output includes json blobs and is less readable and user-friendly.

Workaround: at the bottom of the outputted plan, there should be a URL e.g. "Inspection results can also be viewed at: <url\>". This URL will bring you to the TFC UI that will show the plan in a much more readable way.

Sample
```
==> Inspecting changes to resource(s) if enabling 'task_serv'...

    Generating plan that Consul Terraform Sync will use Terraform to execute

    Terraform v1.0.8
on linux_amd64
Configuring remote state backend...
Initializing Terraform configuration...
{"@level":"info","@message":"Terraform 1.0.8","@module":"terraform.ui","@timestamp":"2021-10-07T17:56:16.701312Z","terraform":"1.0.8","type":"version","ui":"0.1.0"}
{"@level":"info","@message":"module.task_serv.local_file.address[\"consul.lornasong.dc1\"]: Plan to create","@module":"terraform.ui","@timestamp":"2021-10-07T17:56:18.960698Z","change":{"resource":{"addr":"module.task_serv.local_file.address[\"consul.lornasong.dc1\"]","module":"module.task_serv","resource":"local_file.address[\"consul.lornasong.dc1\"]","implied_provider":"local","resource_type":"local_file","resource_name":"address","resource_key":"consul.lornasong.dc1"},"action":"create"},"type":"planned_change"}
{"@level":"info","@message":"module.task_serv.local_file.address[\"web2.lornasong.dc1\"]: Plan to create","@module":"terraform.ui","@timestamp":"2021-10-07T17:56:18.960794Z","change":{"resource":{"addr":"module.task_serv.local_file.address[\"web2.lornasong.dc1\"]","module":"module.task_serv","resource":"local_file.address[\"web2.lornasong.dc1\"]","implied_provider":"local","resource_type":"local_file","resource_name":"address","resource_key":"web2.lornasong.dc1"},"action":"create"},"type":"planned_change"}
[...]

PR #557, opened 2021-12-16
Add tfc_run_url to dry run response

mkam:
Will only be returned with CTS enterprise using Terraform Driver. Chose the
prefix "tfc_" to differentiate a TFC run from a CTS run.

Will be used to output the link to the speculative plan in the CLI. The URL will be especially helpful since the terraform plan readability bug is still open (https://github.com/hashicorp/consul-terraform-sync/issues/425).


lornasong (approved):
Looks good to me! One thought - it could be nice for users if we label this property as "Enterprise only".

I briefly looked around the openapi docs but couldn't tell if a "description" field or something was supported for properties. If you knew, might be worth adding if you agree. Otherwise, no action needed!

mkam:
@lornasong That's a great point about the description, I added it now and double checked that the description rendered in 
a swagger doc.

PR #559, opened 2021-12-17
re-generated client code

wilkermichael:
There was an error with the committed generated code, fixing that


wilkermichael:
@lornasong I'm just fixing a problem after the rebase. There isn't any code to review, I just regenerated the code based on already committed changes.

PR #560, opened 2021-12-20
New ServicesTemplate for services list & condition/source_input services names

lornasong:
Create a new ServicesTemplate which manages `{{ service }}` template function. Tasks configured with `services` list (existing) and condition/source_input "services" `names` field (new) will both use the new ServicesTemplate.

Changes:
 - Create new ServicesTemplate. Move existing logic for `{{ service }}` template function into ServicesTemplate
 - Support condition/source_input "services" `names` field with ServicesTemplate
 - Update `services` list to use ServicesTemplate

Notes:
 - e2e test diff (commit # 3) is more easily viewed without whitespace: https://github.com/hashicorp/consul-terraform-sync/pull/560/commits/e2c53e6f1cd044f53fba0617df60af6039231591?w=1
 - new e2e test is only for condition "services" `names` field. source_input was manually tested. e2e tests will be created for all source_input once it is supported by non-schedule conditions


driver/terraform.go
```diff
@@ -558,7 +558,15 @@ func (tf *Terraform) initTaskTemplate() error {
...
 
-func (tf *Terraform) setNotifier(tmpl templates.Template, serviceCount int) {
+func (tf *Terraform) setNotifier(tmpl templates.Template) {
+	serviceCount := len(tf.task.Services())
+	if cond, ok := tf.task.Condition().(*config.ServicesConditionConfig); ok {
+		serviceCount += len(cond.Names)
+	}
+	if si, ok := tf.task.SourceInput().(*config.ServicesSourceInputConfig); ok {
+		serviceCount += len(si.Names)
+	}
```
findkim, line 568: I'm losing track of the possible combinations with the deprecation. Can you double check me if I'm understanding the intent correctly?

`task.services` cannot be configured alongside [services condition](https://github.com/hashicorp/consul-terraform-sync/blob/f2b798f835da2f43a5e7416b7c00b56177a34c6f/config/task.go#L523) nor [services source input](https://github.com/hashicorp/consul-terraform-sync/blob/f2b798f835da2f43a5e7416b7c00b56177a34c6f/config/task.go#L552). If this is true, then the `+=`s are can be confusing without this context and can be interpreted that overlap is possible, and potentially double counting service names.

Could you replace it with overriding `serviceCount` instead?

```suggestion
	serviceCount := len(tf.task.Services())
	if cond, ok := tf.task.Condition().(*config.ServicesConditionConfig); ok {
		serviceCount = len(cond.Names)
	}
	if si, ok := tf.task.SourceInput().(*config.ServicesSourceInputConfig); ok {
		serviceCount = len(si.Names)
	}
```

e2e/condition_service_test.go
```diff
@@ -15,99 +15,130 @@ import (
...
+					}
+				}
+			}`,
+		},
+	}
+
+	for _, tc := range cases {
+		tc := tc // rebind tc into this lexical scope for parallel use
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel() // Run table tests in parallel as they can take a lot of time
```
findkim, line 66: Nice scoping! (no-op for the PR but to keep in mind) This flags to me we may have this bite us elsewhere in our tests 🤔

templates/tftmpl/template_services.go
```diff
@@ -0,0 +1,161 @@
...
+
+// concatServiceTemplate sorts the services alphabetically and concatenates
+// a template for each service
+func (t ServicesTemplate) concatServiceTemplates() (string, error) {
+	// double-check that service query parameter is configured in only one way
+	// the current way or the deprecated way
+	isCurrent := t.Datacenter != "" || t.Namespace != "" || t.Filter != ""
+	isDeprecated := t.Services != nil
+
+	if isCurrent && isDeprecated {
```
findkim, line 79: Appreciate the readability here! Just double checking, this should never occur with validation done earlier in the code path?

findkim (approved):
Just realized the festive green and red, and enjoying the more red in this PR.

templates/tftmpl/template_services.go
lornasong: Yes, this is not possible for the user due to config validation - added a comment to hopefully clarify. It is possible if ServicesTemplate is manually configured with both e.g. in a  test-case

e2e/condition_service_test.go
lornasong: Can't take credit for this! I borrowed this pattern introduced earlier.

Yea, the issue that I know of is if someone tries to runs e2e test containing t.Parallel (or even some a subset of e2e tests) locally, it will have a race issue when the CTS starts in parallel (I think around downloading TF?!). Last I know for running all tasks (as in circleci), we have one test that runs first to avoid this race. You might have something else in mind though.

driver/terraform.go
lornasong: I was thinking about how we've talked about potentially supporting users to configure both `source_input "services"` and `condition "services"` down the line. This was trying to leave open this possibility but its over-simplified and, to your point, it could potentially could lead to double-counting of services.

Thanks for catching, I updated to your suggestion and added a comment.

templates/tftmpl/template_services.go
findkim: Thanks for the info! Oh sweet I see a new comment added for context too.

e2e/condition_service_test.go
findkim: I see. Nah I haven't sifted through the code to have a specific case to update, but it could be something I wasn't keen on myself in one of my tests. I'll keep my eye out if I come across this pattern for loop + parallel.

driver/terraform.go
findkim: Ahh that convo is ringing a bell now, thanks for the reminder. 👍 on changing it based on the double count.

PR #563, opened 2021-12-21
Support condition/source_input "services" regexp empty string

lornasong:
Previously updated condition/source_input "services" regexp configuration to
support an empty string (which is behaves the same as a wildcard)
Commit: https://github.com/hashicorp/consul-terraform-sync/pull/543/commits/89785a332c572fc24c248927cd978814226e7175

Update templating, which did a check for `regexp == ""`, to allow template and hcat
query to be appended


wilkermichael (approved):
@lornasong looks good to me!

lornasong:
rebased to latest on main

PR #564, opened 2021-12-21
Add condition/source_input "services" names field empty string validation

lornasong:
Check that services names in `names` field are not empty strings

### DIFF
--- a/config/monitor_services.go
+++ b/config/monitor_services.go
@@ -157,12 +157,23 @@ func (c *ServicesMonitorConfig) Validate() error {
 		return fmt.Errorf("either the regexp or names field must be configured")
 	}
 
+	// Validate regex
 	if regexConfigured {
 		if _, err := regexp.Compile(StringVal(c.Regexp)); err != nil {
 			return fmt.Errorf("unable to compile services regexp: %s", err)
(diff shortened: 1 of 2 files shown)


config/monitor_services.go
```diff
@@ -165,6 +166,21 @@ func (c *ServicesMonitorConfig) Validate() error {
...
+		for _, name := range c.Names {
+			if name == "" {
+				emptyName = true
+				break
+			}
+		}
+		if emptyName {
+			return fmt.Errorf("names field includes empty string(s). services " +
+				"names cannot be empty")
+		}
```
findkim, line 181: It seems like the goal is to error if there are empty names, which could be returned at the first occurrence instead of explicitly breaking out of the for loop
```suggestion
		for _, name := range c.Names {
			if name == "" {
				return fmt.Errorf("names field includes empty string(s). services " +
				"names cannot be empty")
			}
		}
```

findkim (approved):
🙌 for more input validation

config/monitor_services.go
lornasong: Thanks for the suggestion! Saves a variable and much more compact

PR #571, opened 2021-12-22
API controller refactor

findkim:
* Add Controller Server interface to API server 
  * Replace driver & event details for create/delete with Controller Server methods 
  * This sets up the `api` package to be a go module and can be used as a client library and is setup for server mock testing
  * Rescoped API package testing for create/delete to test API response codes and objects
* Moved core logic around task creation and deletion out of the `api` package to the `controller` package to collate task related functionality together -- task initialization, template render, task execution
  * [x] TODO: add controller-level tests for the moved code
* Simplify `sendError` function signature since all the use cases had to convert error type to string
* Remove intermediate objects in the api package that wrapped oapigen objects 
  * `type taskRequestConfig struct`
  * `taskResponse oapigen.TaskResponse`
  * The wrapper seemed to be added to allow setting Terraform module variables via a map in the create task API. To reduce the # of task objects the project currently has (4), I opted to add the variable map to the config task object. It's currently not accessible via config file though, but that can be added outside of this refactor effort.
* Remove references of controller-level config, like buffer period and working directory, from task API request handling and use defaults later during task initialization by the controller

Status and Task endpoints are not refactored yet as a part of this PR. Stopgap to keep this PR small and for tests to pass was to add getters for the controller's Store and Drivers.


api/api_test.go
```diff
@@ -107,11 +102,59 @@ func TestServe(t *testing.T) {
 	}
 }
 
+func TestServe_refactored(t *testing.T) {
+	t.Parallel()
+
+	port := testutils.FreePort(t)
+	cases := []struct {
```
findkim, line 109: Add create task unit test

api/task_create_test.go
```diff
@@ -150,24 +141,23 @@ func TestTaskLifeCycleHandler_CreateTask_BadRequest(t *testing.T) {
...
-			err := c.drivers.Add(tc.taskName, d)
-			require.NoError(t, err)
-
-			resp := runTestCreateTask(t, c, "", tc.statusCode, tc.request)
+			resp := runTestCreateTask(t, handler, "", tc.statusCode, tc.request)
 
 			// Task should be added to the drivers list
-			_, ok := c.drivers.Get(tc.taskName)
-			require.True(t, ok)
+			// _, ok := c.drivers.Get(tc.taskName)
```
findkim, line 151: Cleanup

api/task_delete_test.go
```diff
@@ -6,40 +6,49 @@ import (
...
 func TestTaskDelete_DeleteTaskByName(t *testing.T) {
 	t.Parallel()
-	existingTask := "task_a"
+	taskName := "task"
 	cases := []struct {
 		name       string
-		taskName   string
+		mock       func(*mocks.Server)
 		active     bool
 		deleted    bool
```
findkim, line 23: Can probably be cleaned up

controller/readwrite.go
```diff
@@ -352,6 +334,97 @@ func (rw *ReadWrite) checkApply(ctx context.Context, d driver.Driver, retry, onc
...
+			// Once template rendering is finished, break
+			break
+		}
+	}
+
+	// TODO: Set the buffer period
+	// d.SetBufferPeriod()
+
+	// Add the task driver to the driver list
+	err = rw.drivers.Add(taskName, d)
```
findkim, line 386: Only add to drivers on successful create (and run) -- move to TaskCreate and TaskCreateAndRun

findkim:
I started this refactor a while back and there has been many PRs merged since then. I have this PR pointed to a refactor branch. After this is reviewed and merged, we can work on collaborating to resolve the conflicts with main and all of its recent changes.

Missing from this PR is creating a task with inspect option, which will be added back in after the refactor work is rebased to pull in recent changes

mkam (approved):
Looks good to me! I double checked that the action items we had discussed previously have been addressed. I'm really excited for these changes, we'll be able to have a lot more control over the API going forward!

api/task_create.go
```diff
@@ -21,89 +21,49 @@ func (h *TaskLifeCycleHandler) CreateTask(w http.ResponseWriter, r *http.Request
 
 	// Decode the task request
 	var req taskRequest
-	requestID := requestIDFromContext(r.Context())
+	ctx := r.Context()
+	requestID := requestIDFromContext(ctx) // TODO: log with request ID
```
wilkermichael, line 26: @findkim the logger should already log with the request ID. The middleware adds it to the context logger which is extracted on line 19. Let me know if this is what you were looking to do?

findkim: thanks for taking a look and double checking here! Sorry for the vague TODO comment, I was hoping to capture whether the request ID logging is propagated throughout the server-side beyond the handler. Do you know if this is already the case?

wilkermichael: No worries, I'm going through it now to understand the changes since they'll effect the remainder of the task lifecycle work. They look great so far! I can tell a lot of thought and effort went into it and I'm excited to use these changes in my Create Task CLI :D

I see what you mean, no they aren't propagated down yet, I imagine we could pass the logger via context throughout the server side though, which would be even easier now with your refactor. I recall seeing a comment somewhere in the consul code that initializing logger via context was a TODO for their code base as well.

PR #572, opened 2021-12-22
Refactor dry run API to be a part of create tasks API

mkam:
POST `/dryrun_tasks` ->  POST `/tasks?run=inspect`

- Adds same dry run functionality to task creation, removing some duplicated setup work
- Update test cases to have the same cases covered in the dry run tests
- Remove the old code for the separate dry run API
- Remove a check for the run option which is already checked by the openapi validation. A run option that isn't now or inspect does not cause any errors anyway, so didn't seem like it was needed in the handler.


lornasong (approved):
makes sense to me! nice catch removing the redundant check on invalid run options

PR #573, opened 2021-12-22
Only store events on task creation if the task ran successfully

mkam:
Previously, we were always creating an event. If the task was not created in the end,
we would then have an oprhaned event that had no driver associated with it. This caused
an error when getting the overall event statuses.

For a task that is created and not run, the status will be unknown until the first run
is triggered.

Changes:
- Moved where we call the task apply out of the initNewTask method and into the main creation method
- Update storing event to only happen if the task was applied successfully with `?run=now`
- Moved the initialization to happen before checking whether `?run=inspect`

### DIFF
--- a/api/task_create.go
+++ b/api/task_create.go
@@ -55,6 +55,15 @@ func (h *TaskLifeCycleHandler) CreateTask(w http.ResponseWriter, r *http.Request
 		return
 	}
 
+	// Initialize the new task
+	err = initNewTask(r.Context(), d)
+	if err != nil {
+		err = fmt.Errorf("error initializing new task: %s", err)
+		logger.Error("error creating task", "error", err)
+		sendError(w, r, http.StatusBadRequest, err.Error())
+		return
+	}
+
 	// Create a dry run task if run parameter is set to inspect
 	var run string
 	if params.Run != nil {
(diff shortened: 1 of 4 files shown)


api/task_create.go
```diff
@@ -55,6 +55,15 @@ func (h *TaskLifeCycleHandler) CreateTask(w http.ResponseWriter, r *http.Request
...
 
+	// Initialize the new task
+	err = initNewTask(r.Context(), d)
+	if err != nil {
+		err = fmt.Errorf("error initializing new task: %s", err)
+		logger.Error("error creating task", "error", err)
+		sendError(w, r, http.StatusBadRequest, err.Error())
+		return
+	}
+
```
lornasong, line 66: I like that this got refactored out to be shared! Minor naming suggestion: What are your thoughts on the name for method `createDryRunTask()` now that it doesn't call `initNewTask()`? It seems as though it no longer creates the task and only does inspect. Maybe naming it to something like `inspectTask()`?

api/task_create.go
```diff
@@ -55,47 +55,60 @@ func (h *TaskLifeCycleHandler) CreateTask(w http.ResponseWriter, r *http.Request
...
+		// Apply task
+		err = d.ApplyTask(r.Context())
+		if err != nil {
+			err = fmt.Errorf("error applying new task: %s", err)
+			logger.Error("error applying task", "error", err)
+			sendError(w, r, http.StatusBadRequest, err.Error())
+			return
+		}
+		// Store event if apply was successful and task will be created
+		ev.End(err)
```
lornasong, line 106: It seems as though if the task is created but ApplyTask() errored, it would be useful to store an event to notify that the apply was unsuccessful. That is, as I understand, the task is still created (and not rolled back) and ran; it was just unsuccessful. I think I might be missing something. Let me know!

lornasong (commented):
makes sense to not store an event if the task is not created. left a question to check my understanding for the case when the task is created and it ran but unsuccessfully

api/task_create.go
mkam: That's a good point, I'll rename it!

api/task_create.go
mkam: In its current state, the task is rolled back if an apply errors. However, I've been thinking of changing the create and run request to be asynchronous since a task run could take an arbitrary amount of time. In this asynchronous workflow, I think we wouldn't do any roll back and then store the failed events so that the user has feedback on the run.

lornasong: Gotcha! Thanks for checking me. I see what you're saying - the new task's driver doesn't get added to the map of drivers until we know that the apply was successful. Sounds good

lornasong (approved):
looks good to me!

PR #576, opened 2022-01-03
Auto-approve option for delete and enable commands

mkam:
Adds `-auto-approve` as an option to skip the interactive approval step when enabling a task that has changes or deleting a task. The flag name was based on [a similar option for Terraform apply](https://www.terraform.io/cli/commands/apply#automatic-plan-mode).

For enabling tasks, if there are changes, CTS will still inspect the task and output the Terraform plan so that users are able to see what changes will happen when the apply is run.

This same option will be added to task creation once the CLI command has been implemented.

### DIFF
--- a/command/meta.go
+++ b/command/meta.go
@@ -45,6 +45,8 @@ const (
 	FlagClientCert = "client-cert"
 	FlagClientKey  = "client-key"
 	FlagSSLVerify  = "ssl-verify"
+
+	FlagAutoApprove = "auto-approve"
 )
 
 func (m *meta) defaultFlagSet(name string) *flag.FlagSet {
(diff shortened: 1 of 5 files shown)


mkam:
Merging but continuing the discussion for follow ups!

I'm thinking that the short term could be returning the apply output, which I can see coming in handy down the line anyway. 

With just the apply output, we still have the issue where in approval required situations, there could be a difference between the plan outputted and what is eventually applied after approval. But I think it'll be trickier to pass the plan to the apply since we'll have to create a uniquely named output file of the TF plan, pass that plan file name back to the client, and then send it back to the API on the enable and run. 

I'm also not sure if Terraform Cloud has the option to create a run based on a speculative plan. 🤔  I couldn't find any information on a quick look through the docs.

PR #578, opened 2022-01-04
Change condition 'source_includes_var' default from false to true

lornasong:
**Breaking Change for `condition "catalog-services"` and `condition "consul-kv"`**

Currently, the `source_includes_var` field of a `condition` block defaults to false when unconfigured. This PR changes the default value of `source_includes_var` from false to true for all conditions.

User impact: users that do not have `source_includes_var` configured in a `condition "catalog-services"` and `condition "consul-kv"` block will now need to explicitly configure `source_includes_var = false` rather than relying on the default behavior.

```diff
task {
  name = "task_a"
  services = ["x", "y", "z"] 
  condition "consul-kv" {
    path = "my_path"
+   source_includes_var = false // explicitly configure false value now
  }
}
```

Reason: Currently `condition "services"` does not support a `source_includes_var` field and always passes the source input. We plan to add `source_includes_var` field to `condition "services"` block. If `source_includes_var` default remains false, this would create a breaking change for `condition "services"` block. Updating this default value will avoid a breaking change for `condition "services"` which is the most frequent use-case for CTS.


findkim (approved):
Appreciate the in-depth PR summary that includes the rationale. Changes look good and are quite clean thanks to your refactor work.

lornasong:
rebased to latest on main

PR #579, opened 2022-01-04
Feature 522 create client cli

wilkermichael:
This PR is for a new CLI client for creating a new task

### Summary:

- create task command allows a user to create a new task using a provided `.hcl` file containing a single task (and only a single task)
- currently doesn't support adding terraform variables or terraform variable files, this will need to be added in a future PR
- the CLI output for `create task` is a WIP, this is the first pass. One area that could be improved (possible in a future PR) is replacing the current JSON task output with a customized human readable output as discussed with @findkim 

### Known Issue
- CLI output when connecting to a server with the wrong scheme is not displaying the correct message. This is a known issue and is being tracked. This issue only effects the CLI output, TLS/mTLS is functionally working as intended.

### New Supported Commands:
`consul-terraform-sync task create`
`consul-terraform-sync task create help`

### How to use `task create` 
`task create help` output
```
Usage: consul-terraform-sync task create [options] --task-file=<task config>

  Task Create is used to create a new task. It is not to be used for updating a task, it will not create a task if the
  task name already exists.

Options:
  auto-approve false
    Skip interactive approval of inspect plan

  ca-cert 
    Path to a CA file to use for TLS when communicating with Consul-Terraform-Sync. This can also be specified using the CTS_CACERT environment variable.

  ca-path 
    Path to a directory of CA certificates to use for TLS when communicating with Consul-Terraform-Sync. This can also be specified using the CTS_CAPATH environment variable.

  client-cert 
    Path to a client cert file to use for TLS when verify_incoming is enabled. This can also be specified using the CTS_CLIENT_CERT environment variable.

  client-key 
    Path to a client key file to use for TLS when verify_incoming is enabled. This can also be specified using the CTS_CLIENT_KEY environment variable.

  http-addr http://localhost:8558
    The `address` and port of the CTS daemon. The value can be an IP address or DNS address, but it must also include the port. This can also be specified via the CTS_ADDRESS environment variable. The default value is http://localhost:8558. The scheme can also be set to HTTPS by including https in the provided address (eg. https://127.0.0.1:8558)

  port 8558
    The port to use for the Consul Terraform Sync API server, it is preferred to use the http-addr field instead

  ssl-verify true
    Boolean to verify SSL or not. Set to true to verify SSL. This can also be specified using the CTS_SSL_VERIFY environment variable.


Example:

  $ consul-terraform-sync task create --task-file="task.hcl"
  ==> Inspecting changes to resource if creating task 'my_task'...

  // ... inspection details

  ==> Creating the task will perform the actions described above.
      Do you want to perform these actions for 'my_task'?
       - Consul Terraform Sync cannot guarantee that these exact actions will be
             performed if monitored services have changed.

  Enter a value: yes
  
  // ... output continues
```

#### Sample task.hcl
```hcl
task {
  name           = "api-task"
  description    = "Writes the service name, id, and IP address to a file"
  module         = "./example-module"
  providers      = ["local"]
  services       = ["api"]
  enabled = true
}
```

#### Sample run
```
consul-terraform-sync task create --task-file="task.hcl"

==> Inspecting changes to resource if creating task 'api-task'...

    Generating plan that Consul Terraform Sync will use Terraform to execute

    Request ID: e5957b14-7350-23b6-d0f2-b71ca61ef370
    Task: {
      "enabled": true,
      "module": "./example-module",
      "name": "api-task",
      "providers": [
        "local"
      ],
      "services": [
        "api"
      ],
      "variables": {}
    }

    Plan: 

Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
  + create

  # module.api-task.local_file.consul_services will be created
  + resource "local_file" "consul_services" {
      + content              = <<-EOT
            api api 127.0.0.1 
        EOT
      + directory_permission = "0777"
      + file_permission      = "0777"
      + filename             = "test.txt"
      + id                   = (known after apply)
    }

Plan: 1 to add, 0 to change, 0 to destroy.

==> Creating the task will perform the actions described above.
    Do you want to perform these actions for 'api-task'?
     - Consul Terraform Sync cannot guarantee Terraform will perform
       these exact actions if monitored services have changed.

Enter a value: yes

==> Task 'api-task' created
    Request ID: '20e901a3-a1d1-ea4c-938e-1fafd1808419'
```

part of #522

### DIFF
--- a/api/task_lifecycle_client.go
+++ b/api/task_lifecycle_client.go
@@ -102,8 +102,6 @@ func (c *TaskLifecycleClient) CreateTask(ctx context.Context, runOption string, 
 		defer resp.Body.Close()
 	}
 	if err != nil {
-		err = fmt.Errorf("error creating task %s, %w", req.Name, err)
-
 		return TaskResponse{}, err
 	}
 
--- a/api/task_lifecycle_client.go
+++ b/api/task_lifecycle_client.go
@@ -11,11 +11,6 @@ import (
 	"github.com/hashicorp/consul-terraform-sync/config"
 )
 
-const (
-	RunOptionInspect = "inspect"
-	RunOptionNow     = "now"
-)
-
 // TaskLifecycleClient defines a client for task lifecycle requests
 // Currently non task lifecycle requests use the client in api/client.go, but eventually all endpoint
 // may use this new client. In that case TaskLifecycleClient should be renamed
--- a/api/task_lifecycle_client.go
+++ b/api/task_lifecycle_client.go
@@ -1,12 +1,19 @@
 package api
 
 import (
+	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 
 	"github.com/hashicorp/consul-terraform-sync/api/oapigen"
 	"github.com/hashicorp/consul-terraform-sync/config"
+)
+
+const (
+	RunOptionInspect = "inspect"
+	RunOptionNow     = "now"
 )
 
 // TaskLifecycleClient defines a client for task lifecycle requests
@@ -79,6 +86,42 @@ func (c *TaskLifecycleClient) FullAddress() string {
 // Scheme returns the scheme being used by the client
 func (c *TaskLifecycleClient) Scheme() string {
 	return fmt.Sprintf(c.scheme)
+}
+
+// CreateTask takes a task request and run option and sends this information to the client. It then returns
+// a task response object and any errors to the caller.
+func (c *TaskLifecycleClient) CreateTask(ctx context.Context, runOption string, req TaskRequest) (TaskResponse, error) {
+	var run oapigen.CreateTaskParamsRun
+	switch runOption {
+	case RunOptionInspect:
+		run = RunOptionInspect
+	case RunOptionNow:
+		run = RunOptionNow
+	default:
+		err := errors.New("invalid run option provided")
+		return TaskResponse{}, err
+	}
+
+	resp, err := c.Client.CreateTask(ctx, &oapigen.CreateTaskParams{Run: &run}, oapigen.CreateTaskJSONRequestBody(req))
+	if resp != nil {
+		defer resp.Body.Close()
+	}
+	if err != nil {
+		err = fmt.Errorf("error creating task %s, %w", req.Name, err)
+
+		return TaskResponse{}, err
+	}
+
+	var taskResp TaskResponse
+	decoder := json.NewDecoder(resp.Body)
+	err = decoder.Decode(&taskResp)
+	if err != nil {
+		err = fmt.Errorf("invalid response for task %s, %w", req.Name, err)
+
+		return TaskResponse{}, err
+	}
+
+	return taskResp, nil
 }
 
 var _ httpClient = (*TaskLifecycleHTTPClient)(nil)
--- a/api/task_lifecycle_client.go
+++ b/api/task_lifecycle_client.go
@@ -106,7 +106,8 @@ func (d *TaskLifecycleHTTPClient) Do(req *http.Request) (*http.Response, error) 
 	// defer resp.Body.Close() not called for happy path, only called for
 	// unhappy path. caller of this method will close if returned err == nil.
 
-	if resp.StatusCode != http.StatusOK {
+	if checkStatusCodeCategory(ClientErrorResponseCategory, resp.StatusCode) ||
+		checkStatusCodeCategory(ServerErrorResponseCategory, resp.StatusCode) {
 		defer resp.Body.Close()
 
 		var errResp ErrorResponse
--- a/command/meta.go
+++ b/command/meta.go
@@ -54,7 +54,7 @@ func (m *meta) defaultFlagSet(name string) *flag.FlagSet {
 
 	// Values provide both default values, and documentation for the default value when -help is used
 	m.port = m.flags.Int(FlagPort, config.DefaultPort,
-		fmt.Sprintf("The port to use for the Consul Terraform Sync API server, it is preferred to use the %s field instead", FlagHTTPAddr))
+		fmt.Sprintf("The port to use for the Consul-Terraform-Sync API server, it is preferred to use the %s field instead", FlagHTTPAddr))
 	m.addr = m.flags.String(FlagHTTPAddr, api.DefaultAddress, fmt.Sprintf("The `address` and port of the CTS daemon. The value can be an IP "+
 		"address or DNS address, but it must also include the port. This can "+
 		"also be specified via the %s environment variable. The "+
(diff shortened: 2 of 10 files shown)


api/task_lifecycle_client.go
```diff
@@ -1,14 +1,21 @@
...
 
 	"github.com/hashicorp/consul-terraform-sync/api/oapigen"
 	"github.com/hashicorp/consul-terraform-sync/config"
 )
 
+const (
+	RunOptionInspect = "inspect"
+	RunOptionNow     = "now"
+)
+
```
findkim, line 18: heh great minds think alike! Just a heads up that these constants were recently added to the [`api` package](https://github.com/hashicorp/consul-terraform-sync/blob/main/api/task.go#L22-L23).

It may be that it was recently merged to main, it's not included in your branch, but I suspect it'll cause a compile error if merged in. Could you rebase your branch?

api/task_lifecycle_client.go
```diff
@@ -81,6 +88,42 @@ func (c *TaskLifecycleClient) Scheme() string {
...
+		err := errors.New("invalid run option provided")
+		return TaskResponse{}, err
+	}
+
+	resp, err := c.Client.CreateTask(ctx, &oapigen.CreateTaskParams{Run: &run}, oapigen.CreateTaskJSONRequestBody(req))
+	if resp != nil {
+		defer resp.Body.Close()
+	}
+	if err != nil {
+		err = fmt.Errorf("error creating task %s, %w", req.Name, err)
```
findkim, line 110: What do these errors end up looking like with the wrapper? I'm cautious that this is redundant for the client and would be the last bubbled up error with very similar wraps.

command/meta.go
```diff
@@ -210,6 +210,18 @@ func (m *meta) requestUserApprovalDelete(taskName string) (int, bool) {
...
 }
 
+// requestUserApprovalCreate prints a prompt for user approval of deleting a task
+// and waits for the user input. It returns an exit code and boolean describing
+// if the user approved.
+func (m *meta) requestUserApprovalCreate(taskName string) (int, bool) {
+	m.UI.Info("Creating the task will perform the actions described above.")
+	m.UI.Output(fmt.Sprintf("Do you want to perform these actions for '%s'?", taskName))
+	m.UI.Output(" - This action cannot be undone.")
+	m.UI.Output(" - Consul Terraform Sync cannot guarantee Terraform will perform")
```
findkim, line 220: I know you didn't introduce this, but I think it's worthwhile to rename (here and others) to `Consul-Terraform-Sync` to match our documentation!

command/task_create.go
```diff
@@ -0,0 +1,181 @@
...
+// Synopsis is a short one-line synopsis of the command
+func (c *taskCreateCommand) Synopsis() string {
+	return "Creates a new task."
+}
+
+// Run runs the command
+func (c *taskCreateCommand) Run(args []string) int {
+	c.meta.setFlagsUsage(c.flags, args, c.Help())
+	var taskFile string
+	c.flags.StringVar(&taskFile, "task-file", "", "A file containing the hcl definition of a task")
```
findkim, line 82: Our configs support both HCL and JSON, do you know if there is json limitation for the CLI feature?

command/task_create.go
```diff
@@ -0,0 +1,181 @@
...
+
+		return ExitCodeError
+	}
+
+	c.UI.Output(fmt.Sprintf("Request ID: %s", taskResp.RequestId))
+	b, _ := json.MarshalIndent(taskResp.Task, "    ", "  ")
+	c.UI.Output(fmt.Sprintf("Task: %s\n", string(b)))
+	c.UI.Output(fmt.Sprintf("Plan: \n%s", *taskResp.Run.Plan))
+
+	if !*c.autoApprove {
```
findkim, line 158: I mentioned this in the recent `auto-approve` PR and sharing the discussion here. Not necessary asking for changes to be made for this PR, but an enhancement for the `auto-approve` feature would be to skip the `RunOptionInspect` portion and just run now since auto-approve doesn't expect human interaction to begin with.

findkim (approved):
Really appreciate the thorough PR summary and callouts on the current state of the effort. The UX documented made following the PR straightforward.

Looking forward to using this myself!

api/task_lifecycle_client.go
wilkermichael: I just threw in a dummy error to get this to trip, this is what it looks like

```
==> Inspecting changes to resource if creating task 'api-task'...

    Generating plan that Consul Terraform Sync will use Terraform to execute

==> Error: unable to generate plan for 'api-task'
    error creating task api-task, this is an error
```

wilkermichael: maybe this doesn't need to be wrapped, it does seem a bit redundant

command/task_create.go
wilkermichael: nope you're right, I'll fix this description

api/task_lifecycle_client.go
findkim: I've seen the wrapping multiple times, so i don't think the dummy error is representative of what the nested ones would surface.

command/task_create.go
mkam: For reference, here's the conversation around this: https://github.com/hashicorp/consul-terraform-sync/pull/576#pullrequestreview-843784086

api/task_lifecycle_client.go
wilkermichael: Ah I misunderstood your first statement then, you're thinking that the server might bubble up a similar `fmt.Errorf("error creating task %s, %w", req.Name, err)` so we might get the same error written multiple times.

Well then, maybe `task client error` adds the context without being redundant?

wilkermichael: hmm but then there isn't really an error with the client... I think it would be better to just return the error without wrapping in this case

command/meta.go
wilkermichael: Changed this and other instances in this file

command/task_create.go
wilkermichael: Great, I'll leave it as is for now, and we can tackle this enhancement along with the other locations of auto-approve

api/task_lifecycle_client.go
wilkermichael: Just manually going through it, I think this is what the errors end up looking like:

```
==> Inspecting changes to resource if creating task 'api-task'...

    Generating plan that Consul Terraform Sync will use Terraform to execute

==> Error: unable to generate plan for 'api-task'
    error creating task api-task, request returned 400 status code with error: error decoding the request: some error

==> Error: unable to generate plan for 'api-task'
    error creating task api-task, request returned 400 status code with error:  task with name api-task already exists

==> Error: unable to generate plan for 'api-task'
    error creating task api-task, request returned 400 status code with error:  error with task configuration: some error

==> Error: unable to generate plan for 'api-task'
    error creating task api-task, request returned 400 status code with error:  this is an error
```

wilkermichael: I don't think it's any less clear without the wrapping

```
==> Error: unable to generate plan for 'api-task'
        request returned 400 status code with error: error decoding the request: some error

==> Error: unable to generate plan for 'api-task'
        request returned 400 status code with error:  task with name api-task already exists

==> Error: unable to generate plan for 'api-task'
        request returned 400 status code with error:  error with task configuration: some error

==> Error: unable to generate plan for 'api-task'
        request returned 400 status code with error:  this is an error
```

findkim: Ah ok, I wasn't sure how many and whether the other error wraps would bubble up. Looks like this path is clear, but sounds good that it's clear with and without it.

PR #570, opened 2021-12-22
Deprecate user-facing task configs "source" field for new "module" field

lornasong:
Closes: https://github.com/hashicorp/consul-terraform-sync/issues/566

Example:
```diff
task {
  name   = "task_a"
- source = "path/to/module"
+ module = "path/to/module"
  condition "services" {
    names = ["api"]
  }
}
```

Note: this PR is for user-facing "source". A separate PR will be opened for backend instances of "source" e.g. driver.Task.Source

PR #580, opened 2022-01-05
API refactor rebased

findkim:
This is the follow-up PR that extends https://github.com/hashicorp/consul-terraform-sync/pull/571 to be rebased from `main`. The differences between the two PRs:
1. Adds inspect/dryrun task creation back in, which was omitted in #571
2. Updates inspect/dryrun create api to use new controller refactor
3. Updates all references of "config.task.source" to new `module` name #570
4. Updates some unit tests from `task.services` config to `task.condition.services.names` option #565
5. Update `config.TaskConfig` conversion to oapigen type to account for required `condition.catalog-services.regexp` #574
6. Accounts for dryrun changes (refactor to create endpoint, only store events on creation+run)


api/representation.go
```diff
@@ -13,14 +12,9 @@ import (
...
 
-type taskRequestConfig struct {
-	config.TaskConfig
-	variables map[string]string
-}
-
 // ToTaskRequestConfig converts a taskRequest object to a Config TaskConfig object. It takes as arguments a buffer period,
 // and a working directory which are required to finalize the task config.
-func (tr taskRequest) ToTaskRequestConfig(bp *config.BufferPeriodConfig, wd string) (taskRequestConfig, error) {
+func (tr taskRequest) ToTaskRequestConfig() (config.TaskConfig, error) {
```
lornasong, line 17: Not necessary for this PR since I know this is the rebase - I was wondering what your thoughts are on renaming `ToTaskRequestConfig` since  it's now returning a regular task config? Maybe something like `ToTaskConfig`?

api/representation.go
```diff
@@ -126,120 +120,108 @@ func (tr taskRequest) ToTaskRequestConfig(bp *config.BufferPeriodConfig, wd stri
...
 	}
 
-	task.SourceInput = new(oapigen.SourceInput)
-	switch si := trc.SourceInput.(type) {
-	case *config.ServicesSourceInputConfig:
-		if len(si.Names) > 0 {
-			task.SourceInput.Services = &oapigen.ServicesSourceInput{
-				Names: &si.Names,
-			}
-		} else {
```
lornasong, line 187: Want to double check if we should handle `names` for `source_input "services"` (there's a similar handling for `condition "services"`)

api/representation_test.go
```diff
@@ -546,474 +374,3 @@ func TestTaskResponse_String(t *testing.T) {
 		`"source_input":{"services":{"regexp":""}},"version":"","working_dir":"sync-tasks/task"}}`
 	require.Equal(t, expected, actual)
 }
-
-func TestTaskResponse_taskResponseFromConfigTaskConfig(t *testing.T) {
```
lornasong, line 550: Not necessary for this PR since I know this is the rebase - I'm probably missing some context. Looks like this set of tests got deleted but it looks like this function still exists (renamed to `taskResponseFromTaskConfig`)?

api/representation.go
```diff
@@ -126,120 +120,108 @@ func (tr taskRequest) ToTaskRequestConfig(bp *config.BufferPeriodConfig, wd stri
...
+	if tc.SourceInput != nil {
+		task.SourceInput = new(oapigen.SourceInput)
+		switch si := tc.SourceInput.(type) {
+		case *config.ServicesSourceInputConfig:
 			task.SourceInput.Services = &oapigen.ServicesSourceInput{
 				Regexp: si.Regexp,
 			}
-		}
```
findkim, line 191: I think this was a bad rebase -- looking if this needs to be added back

lornasong (commented):
Super excited for this refactor 🎉  I'm particularly enjoying the cleanup in the tests for fields that are not necessary to set.

I did a scan through mostly focusing on rebase changes caused by my prs and had only one question (others are non-blocking/related). Things look good from my side of changes!

api/representation.go
lornasong: Nevermind! Just saw your comment. Looks like you're on it

api/representation.go
findkim: good catch, I'll make the update.

mkam (approved):
Rebase and changes look good to me! We discussed this separately, but wanted to leave a note here that the TaskInspect() signature may require changes to account for the TFC run URL used in CTS Enterprise. I'll look into this as a follow up.

api/representation_test.go
findkim: Thank you for the review. Added it back + updated it! The corresponding function was removed early in the refactor effort but was later added back for `config.TaskConfig` and overlooked testing for it.

findkim:
Thank you @lornasong and @mkam for the set or eyes on this PR series! 🙇

lornasong (approved):
thanks for the changes! looks good to me ✨

wilkermichael (commented):
This is great! The structure make way more sense now.

controller/server.go
```diff
@@ -11,17 +12,48 @@ func (rw *ReadWrite) Config() config.Config {
...
-	return nil
+	d, err := rw.createTask(ctx, taskConfig)
+	if err != nil {
+		return err
+	}
+
+	return rw.runTaskOnce(ctx, d)
 }
 
 func (rw *ReadWrite) TaskDelete(ctx context.Context, name string) error {
```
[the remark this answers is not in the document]
wilkermichael: All good, I see this was merged while I was going through it. No worries about this.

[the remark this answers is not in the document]
findkim: Oh no sorry about that @wilkermichael ! Appreciate your time and review, especially since this is on the larger side. 
+1 to documenting public methods. I missed this and can do it the next PR i make.

PR #583, opened 2022-01-06
Update TaskInspect to return an additional string

mkam:
This is needed in CTS enterprise to return the run URL of the TFC speculative plan. Also fixed the benchmark tests which were still expecting the Controller interface instead of a ReadWriteController.

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -16,5 +16,5 @@ type Server interface {
 	TaskCreate(context.Context, config.TaskConfig) (config.TaskConfig, error)
 	TaskCreateAndRun(context.Context, config.TaskConfig) (config.TaskConfig, error)
 	TaskDelete(ctx context.Context, taskName string) error
-	TaskInspect(context.Context, config.TaskConfig) (bool, string, error)
+	TaskInspect(context.Context, config.TaskConfig) (bool, string, string, error)
 }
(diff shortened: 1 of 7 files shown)


api/server.go
```diff
@@ -16,5 +16,5 @@ type Server interface {
 	TaskCreate(context.Context, config.TaskConfig) (config.TaskConfig, error)
 	TaskCreateAndRun(context.Context, config.TaskConfig) (config.TaskConfig, error)
 	TaskDelete(ctx context.Context, taskName string) error
-	TaskInspect(context.Context, config.TaskConfig) (bool, string, error)
+	TaskInspect(context.Context, config.TaskConfig) (bool, string, string, error)
```
wilkermichael, line 19: This interface now is returning 4 things which is getting fairly long. I wonder if at this point it might be better to define a new object to return? What do you think?

wilkermichael (approved):
Looks good @mkam, just one comment about a small refactor

api/server.go
mkam: Thanks for calling this out, I should have mentioned this somewhere in the PR! It was originally an object that was in the driver package, but we didn't want the API to be importing code from the driver. So definitely something to revisit like you said, I think we just have to figure out where the object will live.

wilkermichael: All good Melissa, Kim added a TODO in her current refactor PR

https://github.com/hashicorp/consul-terraform-sync/blob/090f6114a3cc3e4883a9690f33081d01008ca032/api/server.go#L19

PR #584, opened 2022-01-06
Add support for 'source_includes_var' field on 'condition "services"'

lornasong:
Allows `condition "services"` to be configured with `source_includes_var` field which enables users to include the services information as module input. Similar to other types of dynamic conditions: `catalog-services` and `consul-kv`

`source_includes_var` defaults to true when unset. This aligns with the current behavior of `condition "services"` where services information is by default passed to the module

Reason: this will allow for consistency such that all dynamic conditions support the `source_includes_var` field

Example:
```
task {
  name        = "task_a"
  description = "trigger on api and web but don't pass info to module"
  source      = "path/to/module"
  condition "services" {
    names               = ["api", "web"]
    source_includes_var = false
  }
}
```


api/representation_test.go
```diff
@@ -132,7 +134,8 @@ func TestTaskRequest_ToRequestTaskConfig(t *testing.T) {
 				Enabled: config.Bool(true),
 				Condition: &oapigen.Condition{
 					Services: &oapigen.ServicesCondition{
-						Names: &[]string{"api", "web"},
+						Names:             &[]string{"api", "web"},
+						SourceIncludesVar: config.Bool(false),
```
findkim, line 138: Can you change one of theses tests to test the default behavior for `true`?

findkim (approved):

api/representation_test.go
lornasong: I can change one of the test cases to test when the request has SourceIncludesVar = true, then the config should have SourceIncludesVar = true. Is this what you mean?

Asking since I'm not sure if you might mean the case where the request does not have SourceIncludesVar set, then the config defaults to SourceIncludesVar = true. Since it looks like `ToTaskConfig` no longer calls Finalize() and Validate(), we can't assigning default values here (i.e. the config still has SourceIncludesVar = nil). Let me know if I'm misunderstanding you.

findkim: Oh I see, this is just the conversion method and the defaults aren't handled here. Nah it's ok! Thank you for your explanation.

PR #585, opened 2022-01-06
Return changes_present in run response and message for disabled tasks

mkam:
Added back `changes_present` to the run object returned in the response for `POST /v1/tasks?run=inspect`. This is to keep parity with our [existing API for enabling tasks](https://www.consul.io/docs/nia/api/tasks#changes_present), which will eventually return the same run object.

Also, instead of returning an empty string as the plan when a task is set to disabled, CTS will now return a message explicitly stating that the plan was skipped.

**Example Response**
```
{
    "request_id": "9073d033-e8fd-ba89-7920-dbc1916233db",
    "run": {
        "changes_present": false,
        "plan": "Task is disabled, inspection was skipped."
    },
    "task": {
        "enabled": false,
        "module": "mkam/hello/cts",
        "name": "mkam_inspected_task",
        "providers": [
            "local"
        ],
        "services": [
            "api"
        ]
    }
}
```


lornasong (commented):
looks good to me!

PR #586, opened 2022-01-06
Remove working_directory from create API

mkam:
- There are no restrictions on what a user can specify for the working directory, so it
is a security risk to allow this to be set remotely.
- The working directory is more of a concern to the operator running CTS and not the API user who
may have no insight of the file structure on the host running CTS.


findkim (approved):
Love how clean the removal turned out to be. Looks good

PR #587, opened 2022-01-06
Refactor task update

findkim:
Refactored the handler for task update to remove dependencies on `event` and `driver` to the new controller server interface.
* Moved core handler logic to `ReadWrite.UpdateTask`
* Added method signature for update task to capture current behavior with the flexibility to support future update values
* Moved and updated `api.Test_Task_Update` to `controller.TestServer_TaskUpdate`

Minor update
* Added run option constants to the `api` package. It is now in 2 places (`driver` and `api` package, but it allows the `api` package to move away from the driver package dependency.


config/task_test.go
```diff
@@ -901,6 +901,19 @@ func TestTaskConfig_FinalizeValidate(t *testing.T) {
...
+				Name:   String("task_a"),
+				Module: String("path"),
+				Condition: &ServicesConditionConfig{
+					ServicesMonitorConfig: ServicesMonitorConfig{
+						Names: []string{"serviceA"},
+					},
+				},
+			},
+			true,
+		},
```
findkim, line 916: The tests currently fail because I'm trying to add the functionality where it's possible to get a task and use its config as the base for updating the task (pseudocode)

```go
conf, err := ctrl.Task(name)
conf.Enabled = false
_, err = ctrl.UpdateTask(conf)
```

```
error validating `condition "services"` block: regexp and names fields cannot both be unconfigured. If both are needed, consider including the list of names as part of the regex or creating separate tasks
```

However, what I found is that the config returned is considered invalid for `condition "services"` because the first validation sets `regexp` to an empty string, and the second will error that `names` and `regexp` are set.

findkim: cc @lornasong i think it's worthwhile now to allow the rexep value to be nil :/

wilkermichael:
@findkim I gave a first pass on the first two commits, I can take a second look when the integration tests pass. However, based on your description that shouldn't effect the first two commits.

Great work, steadily making your way through this refactor!

config/task_test.go
lornasong: @findkim thanks for letting me know. I see what you mean about the behavior being not-idempotent. In case it would be useful, I went ahead and made a draft PR to keep `regexp` set as nil: https://github.com/hashicorp/consul-terraform-sync/pull/590. If it's not what you had in mind, no worries. Let me know the best way to proceed.

findkim: Thanks @lornasong for going ahead and making those changes! It looks great and happy to move forward with your PR so it can be idempotent but I don't see it as blocking this PR -- thoughts on this below:

Thinking through the `TaskUpdate` method signature more, I think what we currently have is the better approach with the update config instead of passing the same config object since it's likely we will not want to allow updating all options. I left a TODO note for us to determine in the future where (which package) that update config object should be added to. And for the sake of the refactor to not introduce new things, it'll use `config.TaskConfig` for now and it's now the variable is just conventionally name "updateConf"

findkim:
Thanks @wilkermichael for the early review! Updated the tests and approach, and ready for any additional comments/suggestions

config/task_test.go
lornasong: Sounds good. I'll go ahead with the PR to set `regexp` as nil.

Thanks for the context. Agree, I originally created the `UpdateTaskConfig` struct with the thought of preventing limiting what can be updated. Your plan to use the `config.TaskConfig` for now makes sense to me

controller/server.go
```diff
@@ -104,16 +105,67 @@ func (rw *ReadWrite) TaskInspect(ctx context.Context, taskConfig config.TaskConf
...
+		return false, "", "", fmt.Errorf("task name is required")
+	}
+
+	taskName := *taskConfig.Name
+	logger := rw.logger.With(taskNameLogKey, taskName)
+	logger.Trace("updating task")
+	if rw.drivers.IsActive(taskName) {
+		return false, "", "", fmt.Errorf("task '%s' is active and cannot be updated at this time", taskName)
+	}
+	rw.drivers.SetActive(taskName)
```
wilkermichael, line 124: This gets set to active as sort of a "non-blocking" mutex I guess?

wilkermichael (approved):

wilkermichael:
A couple of my comments were from my previous review, I didn't hit submit :S

Looks good, only one minor comment

api/task.go
```diff
@@ -65,11 +61,17 @@ type UpdateTaskConfig struct {
...
+type InspectPlan struct {
+	ChangesPresent bool   `json:"changes_present"`
+	Plan           string `json:"plan"`
 }
 
 // updateTask does a patch update to an existing task
 func (h *taskHandler) updateTask(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
 	taskName, err := getTaskName(r.URL.Path, taskPath, h.version)
 	logger := logging.FromContext(r.Context()).Named(updateTaskSubsystemName)
```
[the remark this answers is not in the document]
findkim: thanks! I had missed this one

controller/server.go
findkim: yeah the idea was to have a non-blocking mutex, mostly to preserve the original intention of delete behavior before the refactor.

controller/server_test.go
```diff
@@ -261,3 +261,98 @@ func TestServer_TaskDelete(t *testing.T) {
...
+		updateConf.Enabled = config.Bool(true)
+		d.On("UpdateTask", mock.Anything, driver.PatchTask{Enabled: true}).
+			Return(driver.InspectPlan{ChangesPresent: false, Plan: ""}, nil)
+		_, plan, _, err = ctrl.TaskUpdate(ctx, updateConf, "")
+		require.NoError(t, err)
+		assert.Empty(t, plan)
+
+		// No events since the task did not run
+		events := ctrl.store.Read(taskName)
+		assert.Empty(t, events)
```
lornasong, line 322: I liked that you checked for events here. Suggestion to add the events check to the other test cases since events implementation and testing was moved out of the api package.

controller/server_test.go
```diff
@@ -261,3 +261,98 @@ func TestServer_TaskDelete(t *testing.T) {
...
+			Name:    config.String("task_b"),
+			Enabled: config.Bool(true),
+		}
+
+		changed, plan, _, err := ctrl.TaskUpdate(ctx, updateConf, driver.RunOptionInspect)
+
+		require.NoError(t, err)
+		assert.Equal(t, expectedPlan.Plan, plan)
+		assert.Equal(t, expectedPlan.ChangesPresent, changed)
+	})
```
lornasong, line 357: Suggestion to add a new test case for "run" option to confirm that events are added appropriately.

lornasong (approved):
looks good to me! great to see events and driver moved out of the api package. left some non-blocking question/comments

controller/server.go
```diff
@@ -105,6 +106,68 @@ func (rw *ReadWrite) TaskInspect(ctx context.Context, taskConfig config.TaskConf
 	return plan.ChangesPresent, plan.Plan, "", err
 }
 
+func (rw *ReadWrite) TaskUpdate(ctx context.Context, updateConf config.TaskConfig, runOp string) (bool, string, string, error) {
+	// Only enabled changes are supported at this time
+	if updateConf.Enabled == nil {
+		return false, "", "", nil
+	}
```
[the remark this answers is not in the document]
findkim: There is not, I just tested the current behavior and `enabled` is not required in the request -- `{}` is acceptable with 200. But happy to add a check in to return an informative error.

```json
{
  "error": {
    "message": "/v1/tasks/:task_name currently only supports the 'enabled' field. Missing 'enabled' from the request body."
  }
}
```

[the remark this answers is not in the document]
findkim: Though if any other attributes are set in the request body, the expected 400 error does return

```json
{
  "error": {
    "message": "request body's JSON contains unsupported keys: name"
  }
}
```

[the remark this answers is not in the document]
lornasong: For the use-case I'm thinking of your second comment regarding the unsupported keys covers it! Thanks for looking into this and letting me know

controller/server_test.go
findkim: Sounds good, added!

controller/server_test.go
findkim: Thanks, added one for `run=now`

Issue #556, opened 2021-12-16
CTS support for Terraform 1.1 release


devarshishah3:
#### Description
Terraform just released v1.1. CTS is not compatible with it

PR #588, opened 2022-01-07
Support for Terraform  v1.1

mkam:
- Updates constraint to allow v1.1
- Upgrades terraform-exec

Closes https://github.com/hashicorp/consul-terraform-sync/issues/556


wilkermichael (approved):
Looks good @mkam, looked through looks like you caught all of the string changes necessary

PR #590, opened 2022-01-07
Keep `regexp` nil when unset for ServicesMonitor config

lornasong:
Make an exception to allow `regexp` config field for ServicesMonitor to remain
nil even after Finalize() and Validate()

Reason: we may want to validate a config multiple times
Source: https://github.com/hashicorp/consul-terraform-sync/pull/587#discussion_r779917409

### DIFF
--- a/config/monitor_services_test.go
+++ b/config/monitor_services_test.go
@@ -453,3 +453,27 @@ func TestServicesMonitorConfig_GoString(t *testing.T) {
 		})
 	}
 }
+
+// TestServicesMonitorConfig_RegexpNil tests the exception that when `Regexp` is
+// unset, it retains nil value after Finalize() and Validate(). Tests it is
+// idempotent
+func TestServicesMonitorConfig_RegexpNil(t *testing.T) {
+	t.Parallel()
+
+	conf := &ServicesMonitorConfig{
+		Names: []string{"api"},
+		// Regexp unset
+	}
+
+	// Confirm `Regexp` nil
+	conf.Finalize()
+	err := conf.Validate()
+	assert.NoError(t, err)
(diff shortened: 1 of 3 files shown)


config/monitor_services_test.go
```diff
@@ -453,3 +453,27 @@ func TestServicesMonitorConfig_GoString(t *testing.T) {
 		})
 	}
 }
+
+// TestServicesMonitorConfig_RegexpNil tests the exception that when `Regexp` is
+// unset, it retains nil value after Finalize() and Validate(). Tests that
+// that is it idempotent
```
wilkermichael, line 459: nit: change `Tests that that is it idempotent` to `Tests that Regexp is idempotent`

wilkermichael (commented):
Is there potential for these lines to cause issues with de-referencing?

https://github.com/hashicorp/consul-terraform-sync/blob/8d9b55d6056cc42e2d64bc835baa35ba2c3d833a/driver/task.go#L403

https://github.com/hashicorp/consul-terraform-sync/blob/8d9b55d6056cc42e2d64bc835baa35ba2c3d833a/driver/task.go#L363

lornasong:
Currently contextually, those lines won't panic because our config validation errors if `names` is empty and `regexp` is nil. But I will update them to check for `regexp` nil instead to be safe or in case the context changes down the line

wilkermichael:
> Currently contextually, those lines won't panic because our config validation errors if `names` is empty and `regexp` is nil. But I will update them to check for `regexp` nil instead to be safe or in case the context changes down the line

Sounds good, thanks!

wilkermichael (approved):
Great work Lorna, review complete

lornasong:
rebased to latest on main

PR #592, opened 2022-01-10
Rename 'source_input' to 'module_input' config backend changes (pt1)

lornasong:
More "source_input" / "SourceInput" renaming primarily focused on the api, configuration, e2e packages.

Separate, future PR will be opened for for: driver, templating, controller, etc changes

Commits batch together changes to hopefully make things a little less painful to review.


findkim (approved):

api/representation.go
```diff
@@ -33,7 +33,7 @@ func (tr taskRequest) ToTaskConfig() (config.TaskConfig, error) {
 	// Convert source input
 	if tr.ModuleInput != nil {
 		if tr.ModuleInput.Services != nil {
-			si := &config.ServicesSourceInputConfig{
+			si := &config.ServicesModuleInputConfig{
```
wilkermichael, line 36: Caught this in another review, consider renaming `si` instances to `mi` or something else that better reflects `ModuleInput` rather than `SourceInput`

lornasong: Sure, I actually thought the "s" in "si" stood for "services". I ended up going with "input" since it was most readable for me.

PR #599, opened 2022-01-13
Feature 522 add terraform variable file create task cli

wilkermichael:
This PR covers the following enhancements and fixes:

1) configFromDriverTask would panic when a variable was provided that was not of type string:

https://github.com/hashicorp/consul-terraform-sync/blob/fe685da4621d396c46fa44fea27db0aca35f1573/controller/server.go#L172-L175

This was because the value could be of any type, not just a string. If a non string type is used with `.AsString()` this function will panic.

2) Create task CLI should handle the conversion of terraform variable files included in a task definition into 
map[string]string for remote requests.

### Example

**Example Task with Terraform varfiles:**

``` hcl
task {
  name           = "api-task"
  description    = "Writes the service name, id, and IP address to a file"
  module         = "./variables-module"
  providers      = ["local"]
  services       = ["web"]
  variable_files = ["/Users/mwilkerson/dev/michael-cts-example/filename.tfvars","/Users/mwilkerson/dev/michael-cts-example/obj.tfvars"]
  enabled = true
}
```

**Example Terraform varfiles:**
``` hcl
# fielname.tfvars
filename = "michael.txt"
```

```hcl
#obj.tfvars
sample = {
  user_meta = {
    name            = "amanda"
    age             = "32"
    day_of_the_week = "Monday"
  }
  tags = ["cool", "great"]
}
````

**Example Converted Request**
``` json
    Task: {
      "enabled": true,
      "module": "./variables-module",
      "name": "api-task",
      "providers": [
        "local"
      ],
      "services": [
        "web"
      ],
      "variables": {
        "filename": "\"michael.txt\"",
        "sample": "{\"tags\":[\"cool\",\"great\"],\"user_meta\":{\"age\":\"32\",\"day_of_the_week\":\"Monday\",\"name\":\"amanda\"}}"
      }
    }
```

part of #522

### DIFF
--- a/api/representation.go
+++ b/api/representation.go
@@ -2,6 +2,8 @@ package api
 
 import (
 	"encoding/json"
+	"io"
+	"os"
 	"time"
 
 	"github.com/hashicorp/consul-terraform-sync/api/oapigen"
@@ -18,31 +20,48 @@ type TaskRequest oapigen.TaskRequest
 // TaskRequestFromTaskConfig converts a taskRequest object to a Config TaskConfig object.
 func TaskRequestFromTaskConfig(tc config.TaskConfig) (TaskRequest, error) {
 	if len(tc.VarFiles) > 0 {
-		// Load all variables from passed in variable files before
-		// converting to map[string]string
-		loadedVars := make(hcltmpl.Variables)
+		tc.Variables = make(map[string]string)
 		for _, vf := range tc.VarFiles {
-			tfvars, err := tftmpl.LoadModuleVariables(vf)
+			f, err := os.Open(vf)
 			if err != nil {
 				return TaskRequest{}, err
 			}
 
-			for k, v := range tfvars {
-				loadedVars[k] = v
+			err = readToVariablesMap(vf, f, tc.Variables)
+			if err != nil {
+				return TaskRequest{}, err
 			}
-		}
-
-		// Value can be anything so marshal it to equivalent json
-		// and store json as the string value in the map
-		tc.Variables = make(map[string]string)
-		for k, v := range loadedVars {
-			b, _ := ctyjson.Marshal(v, v.Type())
-			tc.Variables[k] = string(b)
 		}
 	}
 
 	tr := oapigenTaskFromConfigTask(tc)
 	return TaskRequest(tr), nil
+}
+
+func readToVariablesMap(filename string, reader io.Reader, variables map[string]string) error {
+	// Load all variables from passed in variable files before
+	// converting to map[string]string
+	loadedVars := make(hcltmpl.Variables)
+	tfvars, err := tftmpl.LoadModuleVariables(filename, reader)
+	if err != nil {
+		return err
+	}
+
+	for k, v := range tfvars {
+		loadedVars[k] = v
+	}
+
+	// Value can be anything so marshal it to equivalent json
+	// and store json as the string value in the map
+	for k, v := range loadedVars {
+		b, err := ctyjson.Marshal(v, v.Type())
+		if err != nil {
+			return err
+		}
+		variables[k] = string(b)
+	}
+
+	return nil
 }
 
 // ToTaskConfig converts a TaskRequest object to a Config TaskConfig object.
--- a/api/representation.go
+++ b/api/representation.go
@@ -6,18 +6,43 @@ import (
 
 	"github.com/hashicorp/consul-terraform-sync/api/oapigen"
 	"github.com/hashicorp/consul-terraform-sync/config"
+	"github.com/hashicorp/consul-terraform-sync/templates/hcltmpl"
+	"github.com/hashicorp/consul-terraform-sync/templates/tftmpl"
+	ctyjson "github.com/zclconf/go-cty/cty/json"
 )
 
 // TaskRequest is a wrapper around the generated TaskRequest
 // this allows for the task request to be extended
 type TaskRequest oapigen.TaskRequest
 
-// TaskRequestFromTaskRequestConfig converts a taskRequest object to a Config TaskConfig object.
-func TaskRequestFromTaskRequestConfig(tc config.TaskConfig) TaskRequest {
-	// TODO: add variable parsing
-	tc.Variables = make(map[string]string)
+// TaskRequestFromTaskConfig converts a taskRequest object to a Config TaskConfig object.
+func TaskRequestFromTaskConfig(tc config.TaskConfig) (TaskRequest, error) {
+	if len(tc.VarFiles) > 0 {
+		// Load all variables from passed in variable files before
+		// converting to map[string]string
+		loadedVars := make(hcltmpl.Variables)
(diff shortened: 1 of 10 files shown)


lornasong (commented):
Changes makes sense to me! I was wondering if there are plans for any tests to exercise the new variable file code?

wilkermichael:
> Changes makes sense to me! I was wondering if there are plans for any tests to exercise the new variable file code?

Yes, I'm working on a e2e test to exercise it at the moment, I was going to do a separate PR, but actually I think it would make sense to just include it here. I'll tag you when it's up!

wilkermichael:
@lornasong I added a unit and e2e test

e2e/test_modules/with_tfvars_file/main.tf
```diff
@@ -0,0 +1,15 @@
...
+    # providers. Declare source for all providers required by the module.
+    local = {
+      source  = "hashicorp/local"
+      version = ">= 2.1.0"
+    }
+  }
+}
+
+resource "local_file" "obj" {
+  content  = join(",", var.sample.tags, [for key, value in var.sample.user_meta : "${key}=${value}"])
```
lornasong, line 13: Looks like this module uses the tfvars file object in the resource. it looks like the e2e test checks `variables.auto.tfvars` but not the resources. do you think it would be add value to add the text file contents?

api/representation.go
```diff
@@ -6,18 +6,43 @@ import (
...
+			for k, v := range tfvars {
+				loadedVars[k] = v
+			}
+		}
+
+		// Value can be anything so marshal it to equivalent json
+		// and store json as the string value in the map
+		tc.Variables = make(map[string]string)
+		for k, v := range loadedVars {
+			b, _ := ctyjson.Marshal(v, v.Type())
```
lornasong, line 39: Not familiar with ctyjson.Marshal. Looks like its marshaling user-provided content. Is there any need to check for error here? Or do we expect to never error or never need to care if there is an error?

lornasong (commented):
test coverage looks nice! left a few questions

api/representation_test.go
```diff
@@ -47,6 +47,289 @@ func TestTaskRequest_String(t *testing.T) {
 	require.Equal(t, expected, actual)
 }
 
+func TestRequest_TaskRequestFromTaskConfig(t *testing.T) {
```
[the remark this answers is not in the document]
wilkermichael: Yeah, I could rename this to `Test_oapigenTaskFromConfigTask` instead, and then write a simple test for `TestTaskResponse_taskResponseFromTaskConfig` which also calls `oapigenTaskFromConfigTask`.

As for testing the variables, I opted to test it in the e2e tests because of the file requirements. So unless we refactor to maybe use an io.Reader instead, it's not really unit testable. I can look into doing that?

api/representation_test.go
```diff
@@ -47,6 +47,289 @@ func TestTaskRequest_String(t *testing.T) {
...
+	cases := []struct {
+		name            string
+		taskConfig      config.TaskConfig
+		expectedRequest TaskRequest
+	}{
+		{
+			name: "minimum_required_only",
```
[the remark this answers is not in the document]
wilkermichael: I removed this and replaced with a test that checks that there are no panics when an empty config is passed in

api/representation_test.go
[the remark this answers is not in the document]
lornasong: Hmmm. There isn't much complexity to the Variables implementation but in general it could be nice to have coverage and exercise / check the specific functionality. I find e2e test great for verifying all the functionality works together to produce a reasonable outcome. I think it could be worth time-boxing a small amount to see if its reasonably easy to make it unit-testable. If not, I'm ok with not having a `TestTaskResponse_TaskResponse_FromTaskConfig`.

That's my thought process. Open to your thoughts

api/representation_test.go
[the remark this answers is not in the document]
lornasong: I like that idea!

api/representation_test.go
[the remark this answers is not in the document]
wilkermichael: Ok, sounds good, I'll give it a shot

api/representation.go
wilkermichael: yes, this should have been changed. When I was first trying out the function I left out the error. I've fixed this in all occurrences

e2e/test_modules/with_tfvars_file/main.tf
wilkermichael: I don't know if it would add much for this particular case which is just testing that things are hooked up correctly for the CLI.

This module was from my own testing, doing this check might be useful as we expand the e2e testing for other cases? I'll make a note of it in our task lifecycle e2e testing asana task

lornasong: I see, adding that check is more exercising the Terraform-side of CTS. For me it makes it a little more "to the end" in the e2e-ness of the test, if that makes sense. But, it seems subjective. I'm ok with whatever you decide. Keeping note of it works for me!

driver/task.go
```diff
@@ -91,7 +91,11 @@ func NewTask(conf TaskConfig) (*Task, error) {
 	// Load all variables from passed in variable files
 	loadedVars := make(hcltmpl.Variables)
 	for _, vf := range conf.VarFiles {
-		tfvars, err := tftmpl.LoadModuleVariables(vf)
```
wilkermichael, line 94: This is in the wrong commit, should be part of the refactor. I'll fix this before mering

api/representation_test.go
```diff
@@ -374,312 +656,20 @@ func TestTaskResponse_String(t *testing.T) {
 	require.Equal(t, expected, actual)
 }
 
+// Test only bare minimum, task conversion scenarios covered in
+// TestRequest_oapigenTaskFromConfigTask
 func TestTaskResponse_taskResponseFromTaskConfig(t *testing.T) {
```
lornasong, line 661: Nice catch deleting these shared test cases!

controller/server.go
```diff
@@ -165,16 +179,26 @@ func (rw *ReadWrite) Tasks(ctx context.Context) ([]config.TaskConfig, error) {
 	drivers := rw.drivers.Map()
 	confs := make([]config.TaskConfig, 0, len(drivers))
 	for _, d := range rw.drivers.Map() {
-		conf := configFromDriverTask(d.Task())
+		conf, err := configFromDriverTask(d.Task())
+		if err != nil {
+			return nil, nil
```
lornasong, line 184: Maybe this should be `return nil, err`?

api/representation_test.go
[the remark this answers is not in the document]
lornasong: I like the abstracting out io.Reader you went with!

lornasong (approved):
thanks for all those test changes! looks good to me

Issue #607, opened 2022-01-18
Expand usage of module_input / source_input block


lornasong:
#### Description

Currently the `source_input` block - recently renamed to `module_input` block https://github.com/hashicorp/consul-terraform-sync/issues/567 - has the following usage restrictions:
- `module_input` block can only be configured when the task is configured with `condition "schedule"`. It cannot be configured with any other type of task condition
- Only one `module_input` block can be configured per scheduled task

We want to expand `module_input` block usage such that:
- `module_input` block can be configured with any type of task, no longer restricting to scheduled tasks
- A task can be configured with multiple `module_input` blocks
- But a task cannot have multiple configurations for the same variable type. Examples:
    - A task cannot be configured with `condition "consul-kv"` and `module_input "consul-kv"` since they both configure the consul_kv variable
    - A task cannot be configured with a `services` list and `module_input "services"` since they both configure the services variable
    - A task cannot be configured with two or more `module_input "catalog-services"` since that will be multiple configurations for the catalog_services variable

#### Use Cases
The expanded usage of `module_input` allows users more flexibility to configure sources of module input for a given task.

This also supports an effort to deprecate task's `services` list so that it can be replaced with `module_input "services"`. For example:

```diff
task {
  name        = "task_a"
  description = "replace `services` with `module_input 'services'`"
  module      = "path/to/module"
- services    = ["api", "db"]
  condition "consul-kv" {
    cron = "* * * * Mon"
  }
+ module_input "services" {
+   names = ["api", "db"]
+ }
}
```

#### Sample Task Configs

Below are sample task configs that demonstrate the expanded usage of `module_input`

```
task {
  name        = "task_b"
  description = "multiple unique module inputs"
  module      = "path/to/module"

  condition "schedule" {
    cron = "* * * * Mon"
  }
  module_input "consul-kv" {
    path = "my/path"
  }
  module_input "services" {
    names = ["api", "db"]
  }
}
```

```
task {
  name        = "task_c"
  description = "non-scheduled tasks support module inputs"
  module      = "path/to/module"

  condition "catalog-services" {
    regexp = ".*"
  }
  module_input "consul-kv" {
    path = "my/path"
  }
  module_input "services" {
    names = ["api", "db"]
  }
}
```

PR #608, opened 2022-01-18
Add new configuration type ModuleInputConfigs

lornasong:
This is the first PR out of ~3 PRs for [Expand usage of module_input / source_input block](https://github.com/hashicorp/consul-terraform-sync/issues/607)

Change: add new configuration type `ModuleInputConfigs` in order to allow multiple configuration per task

Subsequent change: update TaskConfig to replace ModuleInputConfig with ModuleInputConfigs (note plural).

### DIFF
--- a/config/module_input.go
+++ b/config/module_input.go
@@ -18,6 +18,13 @@ type ModuleInputConfig interface {
 	MonitorConfig
 }
 
+// ModuleInputConfigs is a collection of ModuleInputConfig
+type ModuleInputConfigs []ModuleInputConfig
+
+func DefaultModuleInputConfigs() *ModuleInputConfigs {
+	return &ModuleInputConfigs{}
+}
+
 // EmptyModuleInputConfig sets un-configured module inputs with a non-null
 // value
 func EmptyModuleInputConfig() ModuleInputConfig {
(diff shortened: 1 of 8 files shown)


config/monitor_catalog_services.go
```diff
@@ -25,6 +25,10 @@ type CatalogServicesMonitorConfig struct {
 	DeprecatedSourceIncludesVar *bool `mapstructure:"source_includes_var"`
 }
 
+func (c *CatalogServicesMonitorConfig) VariableType() string {
+	return "catalog_services"
```
wilkermichael, line 29: I really like this approach to uniqueness! 

One thing, can we make this string a constant? Same for other variable types, then it would be easier to refer to them in other parts of code, test cases, etc. in the future

wilkermichael: could even use the existing catalogServicesType const, the other monitor's also include a const type variable

config/module_input.go
```diff
@@ -113,3 +120,133 @@ func decodeModuleInputToType(data interface{}, moduleInput ModuleInputConfig) (M
...
+		}
+	}
+
+	// Confirm module_input's type is different from condition
+	if condition == nil {
+		return nil
+	}
+	if ok := varTypes[condition.VariableType()]; ok {
+		err := fmt.Errorf("task's condition block and module_input block "+
+			" both monitor %q variable type. condition and module_input "+
```
wilkermichael, line 228: remove one of the spaces, two spaces between block and both

config/module_input.go
```diff
@@ -18,6 +18,13 @@ type ModuleInputConfig interface {
 	MonitorConfig
 }
 
+// ModuleInputConfigs is a collection of ModuleInputConfig
+type ModuleInputConfigs []ModuleInputConfig
+
+func DefaultModuleInputConfigs() *ModuleInputConfigs {
```
wilkermichael, line 24: Just wanted to point out that this currently isn't used, although I think you have plans for it in the future. Just wanted to make a note of it.

config/module_input_test.go
```diff
@@ -156,3 +157,331 @@ func TestModuleInput_DecodeConfig_Error(t *testing.T) {
...
+		{
+			"happy_path",
+			&ModuleInputConfigs{&ServicesConditionConfig{}},
+			&ModuleInputConfigs{&ConsulKVConditionConfig{}},
+			&ModuleInputConfigs{
+				&ServicesConditionConfig{},
+				&ConsulKVConditionConfig{},
+			},
+		},
+	}
```
wilkermichael, line 307: I guess the merge doesn't handle the case where two Module's of the same type are merged together...

I think this is okay though? Merge just merges the lists, it's up to the validation to catch such errors.

wilkermichael (approved):
I like your solution to identifying uniqueness! Great work Lorna, look forward to seeing the next couple updates as well.

config/monitor_catalog_services.go
lornasong: It looks like `catalogServicesType` const is actually the condition type vs. variable type e.g. `catalog-services` vs. `catalog_services`. It's a little bit different so worried that reusing may cause confusion. Otherwise, it would be great to reuse it.

Right now, there's no code referring to the variable type content within the configuration package. I personally prefer to abstract constants once they're needed. There are other packages that could potentially use the constants, but I think we were also trying to limit how much of the config package we import until we refactor - I might be remembering this wrong.

Let me know your thoughts!

config/module_input.go
lornasong: Thanks for catching

config/module_input.go
lornasong: As of this PR, DefaultModuleInputConfigs() is still currently being used outside of the config package e.g. controller, api, driver. I'll remove it in a subsequent PR when it's no longer being used.

config/module_input_test.go
lornasong: The Merge() itself can handle when two of the same module_inputs are merged e.g. the below passes

```
{
	"happy_path_same_type",
	&ModuleInputConfigs{&ServicesConditionConfig{}},
	&ModuleInputConfigs{&ServicesConditionConfig{}},
	&ModuleInputConfigs{
		&ServicesConditionConfig{},
		&ServicesConditionConfig{},
	},
},
```

You're right Merge() just handles the merge, and this test case would error for Validate(). I can add this test-case

lornasong: I noticed that the test cases are "Condition" instead of "ModuleInput". Will update this as well.

config/monitor_catalog_services.go
wilkermichael: ah ok, I didn't notice the difference

If you want to wait until the const is needed, I'm okay with leaving the consts out for now then.

lornasong:
rebased for latest on main

PR #609, opened 2022-01-18
Bump tests with timing for catalog-services or services+regexp delay

findkim:
https://github.com/hashicorp/consul-terraform-sync/pull/535 added 1 second delay to Catalog Services condition and services condition with regex

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -139,7 +139,7 @@ func TestServe(t *testing.T) {
 			})
 			require.NoError(t, err)
 			go api.Serve(ctx)
-			time.Sleep(time.Millisecond)
+			time.Sleep(500 * time.Millisecond)
 
 			u := fmt.Sprintf("http://localhost:%d/%s/%s",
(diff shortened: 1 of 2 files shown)


api/api_test.go
```diff
@@ -139,7 +139,7 @@ func TestServe(t *testing.T) {
 			})
 			require.NoError(t, err)
 			go api.Serve(ctx)
-			time.Sleep(time.Millisecond)
+			time.Sleep(500 * time.Millisecond)
```
findkim, line 142: ☹️ I touched this 3 times now.

lornasong (approved):
makes sense to me! jumping in to review since it'd be great to have this in before merging other changes in

wilkermichael (approved):
Looks good to me as well!

api/api_test.go
wilkermichael: 3rd times the charm!

PR #614, opened 2022-01-19
Feature 522 fix watcher after delete and inspect

wilkermichael:
Once a task is created via the CLI, the task doesn't update with latest var.services on Consul catalog changes.

I noticed that when a task is created with inspect, and when a task is deleted, service changes still trigger the watcher. I see the following log:

```
2022-01-18T15:14:04.169-0800 [DEBUG] ctrl: template was notified for update but the template ID does not match any task: template_id=389697c5d1a1b87d1b059e80c7c08103
```

I added Deregister functionality to hcat, and added a public method. Now inspect and delete calls will call the deregister function, which fixes the issue.

part of #522

### DIFF
--- a/driver/drivers.go
+++ b/driver/drivers.go
@@ -1,6 +1,7 @@
 package driver
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"sync"
@@ -143,6 +144,12 @@ func (d *Drivers) Delete(taskName string) error {
 
 	if taskName == "" {
 		return errors.New("task name cannot be empty")
+	}
+
+	driver, ok := d.drivers[taskName]
+
+	if ok {
+		driver.DeleteTask(context.Background())
 	}
 
 	delete(d.drivers, taskName)
--- a/driver/drivers.go
+++ b/driver/drivers.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+
+	"github.com/hashicorp/consul-terraform-sync/logging"
 )
 
 // Drivers wraps the map of task-name to associated driver so that the map
@@ -149,7 +151,16 @@ func (d *Drivers) Delete(taskName string) error {
 	driver, ok := d.drivers[taskName]
 
 	if ok {
-		driver.DeleteTask(context.Background())
+		driver.DestroyTask(context.Background())
+	} else {
+		logging.Global().Trace("attempted to destroy a non-existent task", taskNameLogKey, taskName)
+	}
+
+	// delete driver templates associated with task
+	for val, id := range d.driverTemplates {
+		if val == taskName {
+			delete(d.driverTemplates, id)
+		}
 	}
 
 	delete(d.drivers, taskName)
--- a/driver/terraform.go
+++ b/driver/terraform.go
@@ -164,6 +164,14 @@ func (tf *Terraform) InitTask(ctx context.Context) error {
 	return tf.initTask(ctx)
 }
 
+// DeleteTask deletes the task on the driver
+func (tf *Terraform) DeleteTask(ctx context.Context) {
+	tf.mu.Lock()
+	defer tf.mu.Unlock()
+
+	tf.deleteTask(ctx)
+}
+
 // SetBufferPeriod sets the buffer period for the task. Do not set this when
 // task needs to immediately render a template and run.
 func (tf *Terraform) SetBufferPeriod() {
@@ -419,6 +427,11 @@ func (tf *Terraform) initTask(ctx context.Context) error {
 	return nil
 }
 
+// deleteTask attempts to delete and deregister the hashicat template
+func (tf *Terraform) deleteTask(ctx context.Context) {
+	tf.watcher.Deregister(tf.template)
+}
+
 // renderTemplate attempts to render the hashicat template
 func (tf *Terraform) renderTemplate(ctx context.Context) (hcat.ResolveEvent, error) {
 	taskName := tf.task.Name()
@@ -481,6 +494,9 @@ func (tf *Terraform) inspectTask(ctx context.Context, returnPlan bool) (InspectP
 		return InspectPlan{}, errors.Wrap(err,
 			fmt.Sprintf("error tf-plan for '%s'", taskName))
 	}
+
+	// Delete the task once the inspection has completed
+	tf.deleteTask(ctx)
 
 	return InspectPlan{
 		ChangesPresent: c,
--- a/driver/terraform.go
+++ b/driver/terraform.go
@@ -164,8 +164,8 @@ func (tf *Terraform) InitTask(ctx context.Context) error {
 	return tf.initTask(ctx)
 }
 
-// DeleteTask deletes the task on the driver
-func (tf *Terraform) DeleteTask(ctx context.Context) {
+// DestroyTask destroys task dependencies so that it is safe for deletion
+func (tf *Terraform) DestroyTask(ctx context.Context) {
 	tf.mu.Lock()
 	defer tf.mu.Unlock()
 
--- a/driver/terraform.go
+++ b/driver/terraform.go
@@ -169,7 +169,7 @@ func (tf *Terraform) DeleteTask(ctx context.Context) {
 	tf.mu.Lock()
 	defer tf.mu.Unlock()
 
-	tf.deleteTask(ctx)
+	tf.deregisterTemplate(ctx)
 }
 
 // SetBufferPeriod sets the buffer period for the task. Do not set this when
@@ -243,6 +243,7 @@ func (tf *Terraform) InspectTask(ctx context.Context) (InspectPlan, error) {
 	}
 
 	plan, err := tf.inspectTask(ctx, true)
+	tf.deregisterTemplate(ctx)
 	return plan, err
 }
 
@@ -427,8 +428,8 @@ func (tf *Terraform) initTask(ctx context.Context) error {
 	return nil
 }
 
-// deleteTask attempts to delete and deregister the hashicat template
-func (tf *Terraform) deleteTask(ctx context.Context) {
+// deregisterTemplate attempts to deregister the hashicat template
+func (tf *Terraform) deregisterTemplate(ctx context.Context) {
 	tf.watcher.Deregister(tf.template)
 }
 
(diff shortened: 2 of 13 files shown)


go.mod
```diff
@@ -58,3 +58,5 @@ require (
 	golang.org/x/oauth2 v0.0.0-20210220000619-9bb904979d93 // indirect
 	google.golang.org/genproto v0.0.0-20210222212404-3e1e516060db // indirect
 )
+
+replace github.com/hashicorp/hcat => /Users/mwilkerson/dev/hcat
```
mkam, line 62: I believe an alternative way you can do this is to set it to the commit sha of your changes in hcat since they're pushed to the repo. You'd still need to update it to the proper version eventually, but then at least CircleCI can run with your changes. Disclaimer is I haven't tried this myself, but I think I've seen this done by others when reviewing PRs!

wilkermichael: oh yes, that makes sense, I'll give it a shot

e2e/command_test.go
```diff
@@ -599,6 +615,83 @@ func TestE2E_CreateTaskCommand_NoTaskFileProvided(t *testing.T) {
...
+	_, err = cts.Status().Task(taskName, nil)
+	assert.Error(t, err)
+
+	// Re-create task
+	_, err = runSubcommand(t, input, subcmdCreate...)
+	assert.NoError(t, err)
+
+	// Confirm task was re-created
+	_, err = cts.Status().Task(taskName, nil)
+	require.NoError(t, err)
```
wilkermichael, line 677: will change to assert

e2e/command_test.go
```diff
@@ -599,6 +615,83 @@ func TestE2E_CreateTaskCommand_NoTaskFileProvided(t *testing.T) {
...
+	_, err = cts.Status().Task(taskName, nil)
+	assert.NoError(t, err)
+
+	// Delete task
+	subcmdDelete := []string{"task", "delete",
+		fmt.Sprintf("-%s=%s", command.FlagHTTPAddr, cts.FullAddress()),
+	}
+	subcmdDelete = append(subcmdDelete, taskName)
+	_, err = runSubcommand(t, input, subcmdDelete...)
+	require.NoError(t, err)
```
wilkermichael, line 665: will change require's here to asserts

wilkermichael:
TestE2E_ReenableTaskTriggers(t *testing.T) is broken for some reason... as far as I can tell I haven't touched anything in it's pathway... so I'll keep digging.

driver/terraform.go
```diff
@@ -488,6 +501,9 @@ func (tf *Terraform) inspectTask(ctx context.Context, returnPlan bool) (InspectP
 			fmt.Sprintf("error tf-plan for '%s'", taskName))
 	}
 
+	// Delete the task once the inspection has completed
+	tf.deleteTask(ctx)
+
```
lornasong, line 506: I was looking to understand where the template is registered. It looks like it's in `initTaskTemplate()` which is called by `initTask()`. `UpdateTask()` calls `initTask` and `inspectTask`. What are your thoughts on moving this template deregistration call to `UpdateTask()`? I think it would add a little parity to the watcher registration and deregistration. But not sure if there's nuance I'm missing

e2e/command_test.go
```diff
@@ -599,6 +615,83 @@ func TestE2E_CreateTaskCommand_NoTaskFileProvided(t *testing.T) {
 	}
 }
 
+// TestE2E_CreateDeleteCreateTrigger tests that after creating a task, then deleting
+// the task, and then finally re-creating the task, the task will trigger
+func TestE2E_CreateDeleteCreateTrigger(t *testing.T) {
+	t.Parallel()
+
+	tempDir := fmt.Sprintf("%s%s", tempDirPrefix, "create_cmd")
```
lornasong, line 623: I think we try to keep these tempDir values unique. Looks like "create_cmd" is already used by `TestE2E_CreateTaskCommand`

e2e/command_test.go
```diff
@@ -599,6 +615,83 @@ func TestE2E_CreateTaskCommand_NoTaskFileProvided(t *testing.T) {
...
+}`, taskName)
+	taskConfig = taskConfig.appendString(inputTask)
+	taskFilePath := filepath.Join(tempDir, "task.hcl")
+	taskConfig.write(t, taskFilePath)
+
+	// Create task
+	subcmdCreate := []string{"task", "create",
+		fmt.Sprintf("-%s=%s", command.FlagHTTPAddr, cts.FullAddress()),
+	}
+	subcmdCreate = append(subcmdCreate, fmt.Sprintf("--task-file=%s", taskFilePath))
```
lornasong, line 649: nit: rather than appending `--task-file` after `subcmdCreate` is declared & initialized, could it be included when declared & initialized?

e2e/command_test.go
```diff
@@ -599,6 +615,83 @@ func TestE2E_CreateTaskCommand_NoTaskFileProvided(t *testing.T) {
...
+
+	// Confirm task was created
+	_, err = cts.Status().Task(taskName, nil)
+	assert.NoError(t, err)
+
+	// Delete task
+	subcmdDelete := []string{"task", "delete",
+		fmt.Sprintf("-%s=%s", command.FlagHTTPAddr, cts.FullAddress()),
+	}
+	subcmdDelete = append(subcmdDelete, taskName)
```
lornasong, line 663: nit - similarly here, can the `taskName` be included as part of the declaration/initialization like `:=[]string{..., taskName}`

lornasong (commented):
nice catch with the deregistration. left a couple of minor suggestions. since the tests aren't passing yet, i'll take a look again once they're green - let me know!

driver/terraform.go
wilkermichael: Just to clarify, you want me to look into including the call in UpdateTask() as well?

lornasong: I wanted to get your input on the possibility of moving `tf.deleteTask(ctx)` out of `inspectTask()` and into `UpdateTask()` so that `deleteTask()` is called below `inspectTask()`. See what your thoughts are and if it's even possible

wilkermichael: Still not sure I quite understand, but InspectTask() which calls inspectTask() is called by controller/readwrite.go createTask, which doesn't call UpdateTask(). So the tf.deleteTask(ctx) needs to be part of inspectTask() method.

wilkermichael: deleteTask needs to be called wherever an inspect is done

driver/terraform.go
```diff
@@ -170,6 +170,14 @@ func (tf *Terraform) InitTask(ctx context.Context) error {
 	return tf.initTask(ctx)
 }
 
+// DeleteTask deletes the task on the driver
+func (tf *Terraform) DeleteTask(ctx context.Context) {
+	tf.mu.Lock()
+	defer tf.mu.Unlock()
+
+	tf.deleteTask(ctx)
+}
```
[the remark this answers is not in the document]
wilkermichael: The public method, I think, makes sense as DeleteTask, it could do more in the future theoretically. It is called from the server TaskDelete only. Could also be called DeleteTask cleanup maybe? It's kinda like we're destructing the driver before deleting it from the map

I've renamed the private method to deregisterTemplate() however which I think is a bit clearer, does that make sense?

driver/drivers.go
```diff
@@ -145,6 +146,12 @@ func (d *Drivers) Delete(taskName string) error {
 		return errors.New("task name cannot be empty")
 	}
 
+	driver, ok := d.drivers[taskName]
+
+	if ok {
+		driver.DeleteTask(context.Background())
+	}
```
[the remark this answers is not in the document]
wilkermichael: If the map doesn't exist there's nothing to do, we could add a trace log, I don't know if it would be an error.

I think the scenario where the map is empty and Delete is called would be an implementation error, so maybe logging a trace would make sense.

e2e/command_test.go
wilkermichael: copypasta, will fix

driver/terraform.go
lornasong: For posterity, we chatted about this async (feel free to correct if I misrepresent anything):
- Talked about renaming the private method to `deregisterTemplate()`
- There are instances where inspectTask() may not want to deregister the template, so deregisterTemplate can be moved outside of inspectTask

driver/drivers.go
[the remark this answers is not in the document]
lornasong: Gotcha. That makes sense. I like your idea of adding a log - I think seeing that could be helpful for users.

driver/terraform.go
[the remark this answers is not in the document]
lornasong: For posterity, we chatted about this async (feel free to correct if I misrepresent anything):
- Agreed that that renaming the private method to `deregisterTemplate()` works
- For the public method, it may do more than deregister the template in the future so the naming shouldn't have parity with the private method
- `DeleteTask` won't quite literally delete the task from the driver - it's more of a destructor for the driver
- Brainstormed names: `Destruct` and `TearDown`

I also just thought of `Cleanup` and `Destroy` - even `DestroyTask` might work 🤔 it kind of pairs well with existing `InitTask`. If this method will support future options to delete the task's directories, files, and resources, then destroy might be fitting. No strong preferences from me at the moment. Let me know what your thoughts are

driver/drivers.go
```diff
@@ -145,6 +146,12 @@ func (d *Drivers) Delete(taskName string) error {
 		return errors.New("task name cannot be empty")
 	}
 
+	driver, ok := d.drivers[taskName]
+
+	if ok {
+		driver.DeleteTask(context.Background())
```
mkam, line 152: Another step we could add here is to remove all the template IDs associated with this task from the driverTemplates map. This would be more of a memory cleanup.

https://github.com/hashicorp/consul-terraform-sync/blob/main/driver/drivers.go#L51-L53

wilkermichael: I like this idea!

driver/terraform.go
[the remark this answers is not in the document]
wilkermichael: Went with DestroyTask, I agree it pairs well with InitTask

driver/terraform.go
```diff
@@ -170,6 +170,14 @@ func (tf *Terraform) InitTask(ctx context.Context) error {
 	return tf.initTask(ctx)
 }
 
+// DestroyTask deletes the task on the driver
+func (tf *Terraform) DestroyTask(ctx context.Context) {
```
lornasong, line 174: nit: suggestion to update this comment to move away from "delete" language

lornasong (approved):
makes sense to me. thanks for taking the time to chat a bunch about naming!

driver/terraform.go
wilkermichael: > makes sense to me. thanks for taking the time to chat a bunch about naming!

No problem! glad we got it figured out :D

wilkermichael: thanks for catching

PR #617, opened 2022-01-20
Deprecate `port` CLI option

jwair:
**Description**
The option `-http-addr` was added to the CLI as part of implementing TLS on the CTS API (https://github.com/hashicorp/consul-terraform-sync/pull/490). This allows users to specify the whole address of the CTS API, including the port, so `-port` should be deprecated in favor of `-http-addr`.

---

Added 'deprecated' to the helper text of 'port' flag.
Switched usage of 'port' to 'http-addr' in e2e tests.

Output after the change:
```
$ $GOPATH/bin/consul-terraform-sync task disable -port=8558 example-task
==> Waiting to disable 'example-task'...

Warning: 'port' option is deprecated and will be removed in a later version. It is preferred to use the 'http-addr' option instead.

==> 'example-task' disable complete!
```

### DIFF
--- a/command/meta.go
+++ b/command/meta.go
@@ -54,7 +54,9 @@ func (m *meta) defaultFlagSet(name string) *flag.FlagSet {
 
 	// Values provide both default values, and documentation for the default value when -help is used
 	m.port = m.flags.Int(FlagPort, config.DefaultPort,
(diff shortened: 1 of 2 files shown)


mkam (approved):
Looks good to me! Thanks for updating the tests as well!

findkim (approved):
🎉

Issue #569, opened 2021-12-21
Deprecate Task Status API payload's non-status related fields


lornasong:
**REMOVAL DATE: v0.8.0 or a future CTS major release after v0.8.0**
Depends on the specific field. See below for specifics

#### Description

Related deprecations:
 - [Rename task configuration field from "source" to "module" #566](https://github.com/hashicorp/consul-terraform-sync/issues/566)
 - [Deprecate task configuration field "services" #669](https://github.com/hashicorp/consul-terraform-sync/issues/669)

Currently the Task Status API payload contains non-status related fields. These types of fields are all configuration information about the task. These response fields are:
 - TaskStatus.Services
 - TaskStatus.Providers
 - TaskStatus.Events.Config.Services
 - TaskStatus.Events.Config.Providers
 - TaskStatus.Events.Config.Source

In the CTS task configuration, we are deprecating and replacing the corresponding `services` and `source`. Rather than replacing the API response fields with the new corresponding fields, we will replace these fields with a new Get Task API. The Get Task API is more suited to return the task's configuration information.

**User impact**

For the Task Status API response fields listed above, users should switch to using the Get Task API to retrieve that information

**Removal Dates**
 - TaskStatus.Services:  a future major version after v0.8. Will be removed with when corresponding CTS config `task.services` is also removed
 - TaskStatus.Providers: v0.8
 - TaskStatus.Events.Config.Services: v0.8
 - TaskStatus.Events.Config.Providers: v0.8
 - TaskStatus.Events.Config.Source: v0.8

**Example**

Example Task Status API request:
```
$ curl localhost:8558/v1/status/tasks/task_a?include=events
```

Example response with deprecated fields highlighted:
```diff
{
  "task_a": {
    "task_name": "task_a",
    "status": "successful",
    "enabled": true,
-   "providers": ["local"],
-   "services": ["api"],
    "events_url": "/v1/status/tasks/task_a?include=events",
    "events": [
      {
        "id": "97c6e58e-53bd-6a73-53ce-eb0e78b65c2a",
        "success": true,
        "start_time": "2022-02-09T13:16:15.316907-05:00",
        "end_time": "2022-02-09T13:16:46.313164-05:00",
        "task_name": "task_a",
        "error": null,
-       "config": {
-         "providers": ["local"],
-         "services": ["api"],
-         "source": "path/to/module",
-       }
      }
    ]
  }
}
```

PR #618, opened 2022-01-20
Deprecate events payload 'source' field for new 'module' field

lornasong:
Closes: https://github.com/hashicorp/consul-terraform-sync/issues/569

This is part of the work to rename 'source' to 'module'. This PR adds the new 'module' field but leaves the 'source' field as is for now. Later at an announced time, 'source' field will be removed.

Future work: renaming driver for various 'source' language.

Note: I kept the 'source' variable name the same i.e. `Config.Source` rather than following other deprecation patterns to rename the field to `Config.DeprecatedSource`. My thought was that it might be potentially a breaking change since users might use the `Config{}` to unmarshall JSON? Feedback appreciated. Thank you


lornasong:
rebased

PR #619, opened 2022-01-20
More detailed logging about dependencies received by notifiers

lornasong:
Sample:
```
[DEBUG] notifier.suppress: received dependency: type=services names=["api", "web"]
```


wilkermichael (approved):
Looks great, I really like this!

Great addition to logging, this should help us catch more issues in the future.

PR #622, opened 2022-01-20
Move service propagation wait time into tmplfunc Fetch

lornasong:
Context: `TestCondition_Schedule_Basic/module_input_services` (services regex
module_input) was flaking because when two monitored services are registered
simultaneously, the servicesRegexQuery.Fetch() would only return one of the
monitored services because the second services hadn't propagated through Consul
yet.

Previously tasks were all triggered when any dependency was updated. This likely
created a natural delay for service propagation. Recent refactors to trigger
individual tasks only when their dependency updates, removed that delay and
exposed this race condition and increased test flakiness.

Worked on together with @findkim

### DIFF
--- a/templates/tftmpl/tmplfunc/services_regex.go
+++ b/templates/tftmpl/tmplfunc/services_regex.go
@@ -5,6 +5,7 @@ import (
 	"regexp"
 	"sort"
 	"strings"
+	"time"
 
 	consulapi "github.com/hashicorp/consul/api"
 	"github.com/hashicorp/go-bexpr"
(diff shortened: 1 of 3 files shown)


lornasong:
@findkim wanted to also get your review here to make sure I captured what we talked about accurately. Thanks!

templates/tftmpl/tmplfunc/services_regex.go
```diff
@@ -185,6 +186,19 @@ func (d *servicesRegexQuery) Fetch(clients dep.Clients) (interface{}, *dep.Respo
...
+	// additional delay to process new updates to this query result. 1 second is
+	// used to account for Consul cluster propagation of the change at scale.
+	// https://www.hashicorp.com/blog/hashicorp-consul-global-scale-benchmark
+	//
+	// This affects template functions that have propagation depencies on
+	// services. KV query is not affected by this because it is a different
+	// entity within Consul.
+	//
+	// Without this delay, CatalogServices may have services that are not yet
+	// propagated to HealthServices
```
findkim, line 199: Or the service has not returned healthy yet!

The [initial health status for checks is immediately set to "critical"](https://www.consul.io/docs/discovery/checks#initial-health-check-status) by default and has a delay to change healthy based on the checks.

findkim (approved):
Thanks @lorna for the better idea to bubble up the underlying fix to `Fetch` methods instead of randomly in the controller.

https://github.com/hashicorp/consul-terraform-sync/pull/535 had accounted for this behavior for services regex and catalog services as conditions, but not when used as module input for schedule condition. Glad we had that test to flag our attention!

Adding on a little more context on checking for flakey-ness
* [First commit](https://github.com/hashicorp/consul-terraform-sync/commit/24b5727cad191b41b6c324733e8c8bff4864ea59) runs the flakey test `TestCondition_Schedule_Basic/with_module_input_services_regex` 10 times
    * 4/10 success
* [Second commit](https://github.com/hashicorp/consul-terraform-sync/commit/9d7bc58edb0059094b88b6da2b62e00e032205a9) adds the second delay seen in this PR and runs the test 10 times
    * 10/10 success

templates/tftmpl/tmplfunc/services_regex.go
lornasong: Oh yes! Thank you. I updated the comment to include the health status and the link you shared.

lornasong:
@findkim thank you for writing up and sharing all that useful information 💞

lornasong:
@wilkermichael I'm going to go ahead to merge since this PR attempts to address the `TestCondition_Schedule_Basic/module_input_services` test and it will make hopefully make things easier for circleci runs for other branches and merges. If you have any feedback, I'll open another PR to address. Thank you

PR #624, opened 2022-01-21
Embed error in all API response types

mkam:
Rather than using the separate error response type, changed to having all response types optionally return an error message. The `error` object will only be returned in responses if there is an error.

This will allow a user of the API client to decode responses with the same type regardless of if there was an error or not. The responses can still be decoded with `oapigen.ErrorResponse`, though, since it's a subset of the standard responses.

Also made a few unrelated improvements, namely:
- Fixed the expected response code for deleting a task
- Added a handler method for writing the response which sets the headers and encodes the JSON response body
- Updated the old client (still used for status and enable/disable) to check for 4XX or 5XX response codes rather than not 200


wilkermichael:
@mkam I'll take a look at this first thing tomorrow

api/handler.go
```diff
@@ -24,19 +24,22 @@ type Handlers struct {
 //go:generate oapi-codegen  -package oapigen -generate types -o oapigen/types.go openapi.yaml
 //go:generate oapi-codegen  -package oapigen -generate chi-server,spec -o oapigen/server.go openapi.yaml
 
+// writeResponse sets headers and JSON encodes the response body in the response writter
```
wilkermichael, line 27: sp: `response writter` should be `response writer`

api/handler.go
```diff
@@ -24,19 +24,22 @@ type Handlers struct {
...
+func writeResponse(w http.ResponseWriter, r *http.Request, code int, resp interface{}) {
+	logger := logging.FromContext(r.Context()).Named(handlerSubsystemName)
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(code)
+	if err := json.NewEncoder(w).Encode(resp); err != nil {
+		logger.Error("error encoding json", "error", err, "response", resp)
+	}
+}
```
wilkermichael, line 35: Nice refactor!

api/handler.go
```diff
@@ -43,3 +43,21 @@ func sendError(w http.ResponseWriter, r *http.Request, code int, err error) {
 		RequestId: requestIDFromContext(r.Context()),
 	})
 }
+
+func sendTaskError(w http.ResponseWriter, r *http.Request, code int, err error) {
```
wilkermichael, line 47: I think you can simplify `sendTaskError` and `sendTaskDeleteError` to `sendError` by just using the oapigen.ErrorResponse object. Anything consuming the API should still be able to unmarshal as expected, since the ErrorResponse object will match the embedded error response.

api/task_create.go
```diff
@@ -24,7 +24,7 @@ func (h *TaskLifeCycleHandler) CreateTask(w http.ResponseWriter, r *http.Request
 	requestID := requestIDFromContext(ctx)
 	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
 		logger.Error("bad request", "error", err, "create_task_request", r.Body)
-		sendError(w, r, http.StatusBadRequest, fmt.Errorf("error decoding the request: %v", err))
+		sendTaskError(w, r, http.StatusBadRequest, fmt.Errorf("error decoding the request: %v", err))
```
wilkermichael, line 27: See above, but I think this could remain sendError without issue

wilkermichael (approved):
Nice work Melissa!

Just one suggested refactor, I think we can still keep the existing "sendError" functionality. Anything expecting a TaskCreateResponse or TaskDeleteResponse should still be able to unmarshal an ErrorResponse.

api/handler.go
mkam: Oh, that makes sense! I didn't really love these methods anyway because I figured the amount would just grow as we added more response types. Will make the necessary changes!

mkam:
Removed the unnecessary error methods and fixed the typo. Thanks for the suggestion, it's a much simpler change now!

PR #625, opened 2022-01-24
Initialize disabled tasks

mkam:
Removed the restriction where only enabled tasks were initialized. This solves a bug that occurs only in the case where a task is initially disabled and then enabled for the first time, which was introduced during a previous refactor. It also has the advantage of catching initialization errors earlier at start up or creation instead of when it is first enabled.

**Bug Details:**
After enabling a task that was initially disabled (setting `enabled=false` in either the config file or in a create task API request), the task was not executing when there were changes in Consul and logging `template was notified for update but the template ID does not match any task`.

The cause of this is that hcat is returning template IDs to the controller and the controller then uses a mapping of the template IDs to task names to determine which task to run (see https://github.com/hashicorp/consul-terraform-sync/pull/535). However, when a task is disabled, the initialization is skipped so there is no template ID, thus no entry in the map.

**Change Details:**
- Updated e2e test for enabling a task so that the test now catches the bug (see failure [here](https://app.circleci.com/pipelines/github/hashicorp/consul-terraform-sync/2979/workflows/ef249b62-58f4-41dd-84d6-f88d14a9b2da/jobs/7373))
- Removed check for enabled tasks in the Terraform driver init task method

An alternative fix was to add the template ID to the map in `rw.TaskUpdate()`. We'll potentially have to do something like this in the future when we implement updating tasks fully where the template and template ID may change. However, removing the init restriction seemed like a better general fix, as it made a task disabled initially less of a special case, and also had the added benefit of quicker feedback if the init failed.

### DIFF
--- a/driver/terraform_test.go
+++ b/driver/terraform_test.go
@@ -656,16 +656,30 @@ func TestGetTerraformHandlers(t *testing.T) {
 
 func TestDisabledTask(t *testing.T) {
 	t.Run("disabled-tasks", func(t *testing.T) {
-		// tests that disabled tasks don't require mocking any calls and does
-		// not throw any errors
-
-		tf := &Terraform{
-			task:    &Task{name: "disabled_task", enabled: false},
-			watcher: new(mocksTmpl.Watcher),
-			logger:  logging.NewNullLogger(),
-		}
+		// tests that disabled tasks don't require mocking any calls other
+		// than initialization and validation and that the calls do not
+		// throw any errors
 
 		ctx := context.Background()
+		c := new(mocks.Client)
+		c.On("Init", ctx).Return(nil).Once()
+		c.On("Validate", ctx).Return(nil).Once()
+
+		w := new(mocksTmpl.Watcher)
+		w.On("Register", mock.Anything).Return(nil).Once()
+
+		dirName := "disabled-task-test"
+		deleteTemp := testutils.MakeTempDir(t, dirName)
+		defer deleteTemp()
+
+		tf := &Terraform{
+			task: &Task{name: "disabled_task", enabled: false,
+				workingDir: dirName, logger: logging.NewNullLogger()},
+			client:     c,
+			fileReader: func(string) ([]byte, error) { return []byte{}, nil },
+			watcher:    w,
+			logger:     logging.NewNullLogger(),
+		}
 
 		err := tf.InitTask(ctx)
 		assert.NoError(t, err)
--- a/e2e/api_test.go
+++ b/e2e/api_test.go
@@ -257,8 +257,8 @@ func TestE2E_StatusEndpoints(t *testing.T) {
 func TestE2E_TaskEndpoints_UpdateEnableDisable(t *testing.T) {
 	t.Parallel()
 	// Test enabling and disabling a task
-	// 1. Start with disabled task. Confirm task not initialized, resources not
-	//    created
+	// 1. Start with disabled task. Confirm task is initialized, but
+	//    not run (resources not created)
 	// 2. API to inspect enabling task. Confirm plan looks good, resources not
 	//    created, and task not actually enabled.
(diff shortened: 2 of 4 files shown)


driver/terraform_test.go
```diff
@@ -656,17 +656,29 @@ func TestGetTerraformHandlers(t *testing.T) {
 
 func TestDisabledTask(t *testing.T) {
 	t.Run("disabled-tasks", func(t *testing.T) {
-		// tests that disabled tasks don't require mocking any calls and does
-		// not throw any errors
+		// tests that disabled tasks don't require mocking any calls other
+		// than initialization and does not throw any errors
```
wilkermichael, line 660: I guess initialization and validation?

e2e/api_test.go
```diff
@@ -272,12 +272,14 @@ func TestE2E_TaskEndpoints_UpdateEnableDisable(t *testing.T) {
...
 	cts := ctsSetup(t, srv, tempDir, disabledTaskConfig(tempDir))
 
-	// Confirm that terraform files were not generated for a disabled task
-	files := testutils.CheckDir(t, true, fmt.Sprintf("%s/%s", tempDir, "disabled_task"))
-	require.Equal(t, len(files), 0)
+	// Confirm that terraform files were generated for a disabled task
+	taskPath := filepath.Join(tempDir, disabledTaskName)
+	files := testutils.CheckDir(t, true, taskPath)
+	require.Greater(t, len(files), 0)
+	testutils.CheckFile(t, true, taskPath, "terraform.tfvars.tmpl")
```
wilkermichael, line 279: nice! this was the missing ingredient. In the future this would be a good check to see that a task actually ran. Maybe in the future we could make this into a utils call? Something like "testutils.VerifyTaskRan" which could do the event checking and and check for the terraform.tfvars.tmpl file. Something like that.

e2e/command_test.go
```diff
@@ -159,6 +159,26 @@ func TestE2E_EnableTaskCommand(t *testing.T) {
...
+				"event count did not increment once. task was not triggered as expected")
+
+			// make a change in Consul and confirm a new event is triggered
+			now := time.Now()
+			service := testutil.TestService{ID: "api-1", Name: "api"}
+			testutils.RegisterConsulService(t, srv, service, defaultWaitForRegistration)
+			api.WaitForEvent(t, cts, disabledTaskName, now, defaultWaitForEvent)
+			eventCountNow = eventCount(t, disabledTaskName, cts.Port())
+			require.Equal(t, eventCountBase+1, eventCountNow,
+				"event count did not increment once. task was not triggered as expected")
```
wilkermichael, line 181: Don't you need to also check for generated files here as well? To verify the task actually runs?

wilkermichael (approved):
Great job, nice quick change too!

Did you test the scenario where you create a task via the api disabled, and then enable it via the api? No reason why it shouldn't work, just curious

e2e/command_test.go
mkam: We could! The reason I didn't  is because I found out that even before the fix, there wasn't a situation where the event was created but the task didn't run. I know you had mentioned that when we last discussed this bug, but I believe the confusion is that you were looking at the reenable test, which starts with an enabled task.

But no real reason to leave it out, there's always a chance it could be a bug in the future.

e2e/api_test.go
mkam: Yeah, this in particular could be more of a check task is initialized rather than it ran. But it would be a good validation method since we'll have more situations where a task is initialized but not run!

e2e/command_test.go
mkam: Added the check for services!

e2e/api_test.go
mkam: Will punt this to a later date!

PR #626, opened 2022-01-24
Task field in create request

jwair:
Context:
API payload was different from CLI task-file
 - API: json {“name”: “”}
 - CLI: json {“task”: {“name”: “”}}

This PR unifies the two styles by introducing the `task` field to the API request payload


mkam (approved):
Looks good to me! Good call on making task a required field.

jwair:
> Looks good to me! Good call on making task a required field.

That was @findkim 's great suggestion!

PR #634, opened 2022-01-26
add netcgo tag to darwin build

modrake:
This corrects a longstanding issue with DNS on mac builds that we were unable to address until migrating our build platform.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -347,6 +347,8 @@ jobs:
           go-version: ${{ matrix.go }}
 
       - name: Build
+        env:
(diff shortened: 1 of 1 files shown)


lornasong (approved):
thanks for the changes!

lornasong:
@modrake we're trying to figure out if we need to backport this change to our release branches. Is there a way for us to know if our older branches migrated to using the new build platform that resolves this DNS issue? Please let me know if you have any recommendations! Thanks!

lornasong:
Update: confirmed with @ modrake that we are good to backport to other branches

PR #636, opened 2022-01-26
Sequential E2E Tests for Local Runs

jwair:
E2E tests are very resource demanding when run in parallel. Most tests spin up instances of Consul, CTS, and Terraform, and that results in many tests failing with a timeout because CTS does not become ready within the reasonable timeout period we set waiting for its API to be up.

In this change, we introduce a new mode for running the E2E tests sequentially. This is gated by test flag `-local`. There is a new `make` target that sets the `-local` flag, expands the timeout duration, and runs the tests in verbose mode.

The other `make` targets are not impacted. The CI pipelines continue to run without any modifications (E2E tests run in parallel).

Tested locally, all tests ran successfully:
image
image


api/representation_test.go
```diff
@@ -485,7 +485,7 @@ func TestTaskRequest_ToRequestTaskConfig(t *testing.T) {
 				Name:   config.String("task"),
 				Module: config.String("path"),
 				Condition: &config.CatalogServicesConditionConfig{
-					config.CatalogServicesMonitorConfig{
+					CatalogServicesMonitorConfig: config.CatalogServicesMonitorConfig{
```
jwair, line 488: unifying style, fixing `go vet` issue

controller/server_test.go
```diff
@@ -23,7 +23,7 @@ var validTaskConf = config.TaskConfig{
 	Name:    config.String("task"),
 	Module:  config.String("module"),
 	Condition: &config.CatalogServicesConditionConfig{
-		config.CatalogServicesMonitorConfig{Regexp: config.String("regex")},
+		CatalogServicesMonitorConfig: config.CatalogServicesMonitorConfig{Regexp: config.String("regex")},
```
jwair, line 26: unifying style, fixing `go vet` issue

e2e/utils.go
```diff
@@ -38,7 +40,7 @@ const (
 
 	// liberal default times to wait
 	defaultWaitForRegistration = 8 * time.Second
-	defaultWaitForEvent        = 8 * time.Second
+	defaultWaitForEvent        = 15 * time.Second
```
jwair, line 43: This timeout needed to be relaxed because it was too tight even when tests run sequentially

e2e/utils.go
```diff
@@ -310,3 +314,23 @@ func validateVariable(t *testing.T, contains bool, workingDir, name, value strin
 	}
 	assert.Fail(t, fmt.Sprintf("variable '%s' not found in terraform.tfvars", name))
 }
+
+func setParallelism(t *testing.T) {
```
findkim, line 318: Nice! Appreciate the extra work to catch all of the parallel calls through out the test package.

You had mentioned the approach to use a test flag, and was wondering if you could share why it didn't work with the structure you have here?

```
var localFlag = flag.Bool("local", false, "")

func setParallelism(t *testing.T) {
    if !*local {
        t.Parallel()
    }
}
```

```
go test ./e2e -local
```

findkim (approved):

e2e/utils.go
jwair: I tried the `-short` flag but `testing.Short()` was always false and I didn't know why. I didn't dig too deep because I thought we might be overloading the usage of `-short` and adding some confusion anyway.
I wasn't aware that I can set custom flags, it is a cleaner approach for sure. I will test this and switch to it instead of using an env var if it works as expected.

wilkermichael (approved):

e2e/utils.go
findkim: I see, it was around the `-short` option. I've typically paired `-short` with `t.Skip()`, but also haven't dug into the nuances with that case vs our parallel option. I like your callout that overloading that flag could add confusion, because that wouldn't make the testing run shorter, longer actually 😅 

Cool, interested to see how that plays out!

jwair: Addressed in https://github.com/hashicorp/consul-terraform-sync/pull/636/commits/90228f263a6ea9ed79cf6ef524766175b82330c3

PR #643, opened 2022-01-28
Final updates for "source", "src", and "include"

lornasong:
Previously, deprecated configurations regarding "source", "src", "included" were
updated in the templates, driver, and e2e directory.

This is the final search across the whole CTS repo for this language

Commits:
 - Rename "source", "src", and "include"
 - Rename NoMonitorConfig => NoConditionConfig to adjust for change to support multiple `module_input` blocks


findkim (approved):
🎉 🎉

PR #646, opened 2022-01-31
Fixed indentation of openapi.yaml (example fields)

jwair:
Some array examples were nested one level too deep, they are examples of the array type, not the item type within it.


lornasong (approved):
makes sense to me!

PR #653, opened 2022-02-01
Cleanup and Test Improvements

jwair:
Improved tests - added missing asserts, removed overriding builtin 'delete', unified naming
Removed redundant type declarations and casting
Removed conflicts with package names
Improved naming and comments
Unified usage of 'require' test package instead of a mix of require and assert

### DIFF
--- a/e2e/command_test.go
+++ b/e2e/command_test.go
@@ -245,8 +245,7 @@ func TestE2E_ReenableTaskTriggers(t *testing.T) {
 
 	cts, stop := api.StartCTS(t, configPath)
 	t.Cleanup(func() {
-		err := cleanup()
-		require.NoError(t, err)
+		_ = cleanup()
 		stop(t)
 	})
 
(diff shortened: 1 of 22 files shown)


jwair:
Working on the e2e test failures ...

e2e/command_test.go
```diff
@@ -244,7 +245,8 @@ func TestE2E_ReenableTaskTriggers(t *testing.T) {
 
 	cts, stop := api.StartCTS(t, configPath)
 	t.Cleanup(func() {
-		cleanup()
+		err := cleanup()
+		require.NoError(t, err)
```
mkam, line 249: I'm not so sure we should be failing on the cleanup like this. If there is an issue with deleting a temporary directory, then we'd have to rerun the tests even though the behavior of CTS was passing.

jwair: Addressed in https://github.com/hashicorp/consul-terraform-sync/pull/653/commits/40cb2fc4f6637d896e608977841e3c4e022645ac

mkam (approved):
Changes look good to me! Thanks for all the improvements!

PR #654, opened 2022-02-01
Update changelog with buffer period min behavior change

wilkermichael:
This changelog update reflects the change to how the minimum value of the buffer period is used when triggering a task.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ FEATURES:
 * Support for Terraform v1.1 [[GH-588](https://github.com/hashicorp/consul-terraform-sync/pull/588)]
 
 IMPROVEMENTS:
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ FEATURES:
 * Support for Terraform v1.1 [[GH-588](https://github.com/hashicorp/consul-terraform-sync/pull/588)]
 
 IMPROVEMENTS:
+* Library enhancement improves the handling of the min buffer period configuration to more closely match documentation and intended behavior. Tasks will now trigger only after the min time of the buffer period has transpired, previously tasks would trigger immediately, once, before honoring the min buffer period time
```
mkam, line 14:
```suggestion
* Tasks will now always trigger only after the minimum time of the buffer period has transpired. Previously tasks would trigger immediately, once, before honoring the min buffer period configuration.
```
I was thinking that we should lead with the change so it's an easier at-a-glance read. And then I would personally drop the library enhancement and add these details to the PR and link the PR.

mkam: Looks good to me!

mkam (approved):

lornasong (approved):
looks good to me, details on the linked GH pr are helpful!

PR #656, opened 2022-02-01
[RDX-272] Enable security-scan and update build.yml for release/0.4.x branch

claire-labry:
What this PR does is it updates the `build.yml` to the `release/0.4.x` branch to make the workflow a bit more drier and matches the current convention that's in [crt-core-helloworld.](https://github.com/hashicorp/crt-core-helloworld/blob/main/.github/workflows/build.yml)

This updates the build to follow a matrix approach, cleans up the linux-packaging, and making the workflow more abstract and easier to read. 

I also noticed that this branch did not have `security-scanner`, that was added as well :) 

If there are any questions or if anything is not clear, please feel free to ask! ❤️  

Test are passing [here.](https://github.com/hashicorp/crt-workflows-common/actions/runs/1780547448)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,10 +5,10 @@ on:
     # Sequence of patterns matched against refs/heads
     branches:    
       # Push events on main branch
-      # - main
+      - main
       # Push events to branches matching refs/heads/release/**
-      # - 'release/**'
-      - 'clabry-update-crt'
+      - 'release/**'
+
 
 env:
   PKG_NAME: "consul-terraform-sync"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,9 +5,10 @@ on:
     # Sequence of patterns matched against refs/heads
     branches:    
       # Push events on main branch
-      - main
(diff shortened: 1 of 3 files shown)


lornasong (approved):
thank you! makes sense to me!

one question I had - in the 0.3 equivalent PR, it looks like at the top of the build.yml there is a `workflow_dispatch` ([ex](https://github.com/hashicorp/consul-terraform-sync/blob/86b1504a251f1e7c69190f80674532992c73c750/.github/workflows/build.yml#L4)). I was curious if 0.4 would also need that?

sarahethompson (approved):

claire-labry:
Great eye @lornasong! This actually snuck into the `release/0.3.x` branch and can definitely be removed. It is a way for folks to manually trigger the workflow in GitHub Actions without pushing a new commit every time. It's likely not really useful for product teams and more for testing purposes for RDX team. My sincere apologies for that, I will go and raise a PR in that branch to get it removed. Thank you 💕

lornasong:
thank you for explaining! no worries at all - I've definitely been there 💞

PR #662, opened 2022-02-03
Validate template dependency queries at task initialization

mkam:
Added a validation step after the template has been registered that executes the fetch request on all the Consul dependencies. This is intended to catch issues like invalid query parameters (setting a datacenter that does not exist, using the Consul enterprise namespace feature against an OSS Consul instance, etc.). By failing faster on invalid query parameters, it also fixes a bug in the create API workflow, as described below.

**Example Output**
```
consul-terraform-sync task create -task-file=task.hcl
==> Inspecting changes to resource if creating task 'mkam_test_task'...

    Generating plan that Consul Terraform Sync will use Terraform to execute

==> Error: unable to generate plan for 'mkam_test_task'
    request returned 400 status code with error: unable to retrieve data from
Consul: catalog.services.registration(dc=foo&regexp=api|web): Unexpected
response code: 500 (No path to datacenter), Request ID:
7bdd9a6f-3530-4314-5051-a62f349ffce8
```

**Bug Description**
If a task created via the API had an invalid query parameter, the request would hang indefinitely. This is happens because of two reasons:

 1.)  The RenderTemplate method does not error and also will never return Complete when a fetch errors. This is because it deals only with rendering the data, not how/when the data was retrieved.

 2.) The fetch error is returned to the [main Run loop](https://github.com/hashicorp/consul-terraform-sync/blob/70acf6445a49456aa922e1299833eeb1eccd6380/controller/readwrite.go#L89-L97) that is watching the dependencies, and there's not a clean way to pass this error to the API method that's processing the create.
 
 **Solution Details**
- Add a timeout for the render loop in the API task creation to prevent any future infinite looping. Chose 5 minutes arbitrarily.
- Validate a template by calling the fetch method on a template's dependencies as part of task initialization. Use latest hcat version in order to get access to Watcher's clients.

**Additional Changes**
Pardon the bit of scope creep! I wanted to fix some stuff up with the error messages and the rendering loop that I observed as I was working on this feature.
- Remove some of the error wrapping so that the end user of the API/CLI is given a more straightforward error
- Improve an error message for a failed catalog services fetch
- Add a small delay in the render template loop of the API task creation to keep down the number of render calls and log lines. This doesn't have to be a permanent solution, but it seemed like the best option for now.

### DIFF
--- a/controller/readwrite.go
+++ b/controller/readwrite.go
@@ -383,13 +383,14 @@ func (rw *ReadWrite) createTask(ctx context.Context, taskConfig config.TaskConfi
 		return nil, err
 	}
 
-	timeout := time.After(5 * time.Minute)
+	timeout := time.After(1 * time.Minute)
 	for {
 		select {
 		case <-ctx.Done():
 			return nil, ctx.Err()
 		case <-timeout:
-			return nil, fmt.Errorf("timed out rendering template for task '%s'", taskName)
+			logger.Error("timed out rendering template")
+			return nil, fmt.Errorf("error initializing task")
 		default:
 		}
 		ok, err := d.RenderTemplate(ctx)
@@ -401,7 +402,7 @@ func (rw *ReadWrite) createTask(ctx context.Context, taskConfig config.TaskConfi
 			// Once template rendering is finished, return
 			return d, nil
 		}
-		time.Sleep(250 * time.Millisecond) // waiting because cannot block on a dependency change
+		time.Sleep(50 * time.Millisecond) // waiting because cannot block on a dependency change
 	}
 }
 
--- a/controller/readwrite.go
+++ b/controller/readwrite.go
@@ -373,13 +373,14 @@ func (rw *ReadWrite) createTask(ctx context.Context, taskConfig config.TaskConfi
 	d, err := rw.createNewTaskDriver(taskConfig)
 	if err != nil {
 		logger.Error("error creating new task driver", "error", err)
-		return nil, fmt.Errorf("error creating new task driver: %v", err)
+		return nil, err
 	}
 
 	// Initialize the new task
 	err = d.InitTask(ctx)
 	if err != nil {
-		return nil, fmt.Errorf("error initializing new task, %s", err)
+		logger.Error("error initializing new task", "error", err)
+		return nil, err
 	}
 
 	timeout := time.After(5 * time.Minute)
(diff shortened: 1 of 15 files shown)


e2e/module_input_test.go
```diff
@@ -404,3 +404,44 @@ func TestModuleInput_ConsulKVCondition(t *testing.T) {
 		})
 	}
 }
+
+// TODO: TestModuleInput_Services_InvalidQueries when CLI options are implemented
```
mkam, line 408: Oh, oops, I've merged the PR to implement this now so I'll add the test!

controller/readwrite.go
```diff
@@ -373,30 +373,35 @@ func (rw *ReadWrite) createTask(ctx context.Context, taskConfig config.TaskConfi
...
 		if err != nil {
 			logger.Error("error rendering task template")
-			return nil, fmt.Errorf("error rendering template for task '%s': %s", taskName, err)
+			return nil, err
 		}
 		if ok {
 			// Once template rendering is finished, return
 			return d, nil
 		}
+		time.Sleep(250 * time.Millisecond) // waiting because cannot block on a dependency change
```
mkam, line 404: I know a sleep is not a great solution, but the constant logging was really ruining our logs in TRACE mode and I wanted to put in at least a stop gap for now. Happy to discuss other options or to adjust the sleep time to lower!

wilkermichael: yeah, I'm okay with this for now, Melissa and I discussed async that we should revisit logs in general and perhaps do an audit. As it stands right now, trace is the necessary log level for development work, but if most of the development level logs were debug logs instead, then high output trace logs like this would be more acceptable.

wilkermichael (approved):
Good work Melissa, I can tell a lot of effort went into finding the best solution here!

controller/readwrite.go
mkam: Decreased down to 50 ms.

controller/readwrite.go
```diff
@@ -382,10 +382,13 @@ func (rw *ReadWrite) createTask(ctx context.Context, taskConfig config.TaskConfi
 		return nil, fmt.Errorf("error initializing new task, %s", err)
 	}
 
+	timeout := time.After(5 * time.Minute)
```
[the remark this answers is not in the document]
mkam: Decreased to 1 minute! We can revisit this later if we want to reduce it further.

PR #663, opened 2022-02-03
API Client Improvements and Tests

jwair:
Improved tests/cleanup for API client
Added unit tests for API client
Improved environment lookup for API client creation
Unified using of 'require' test package instead of a mix of 'require' and 'assert'

### DIFF
--- a/api/client_test.go
+++ b/api/client_test.go
@@ -1,15 +1,24 @@
 package api
 
 import (
+	"math/rand"
 	"os"
 	"strconv"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-func TestDefaultClientConfig_WithOSEnvSet(t *testing.T) {
+func Test_DefaultClientConfig_WithEnvVars(t *testing.T) {
+	t.Cleanup(func() {
+		_ = os.Unsetenv(EnvAddress)
+		_ = os.Unsetenv(EnvTLSCACert)
+		_ = os.Unsetenv(EnvTLSCAPath)
+		_ = os.Unsetenv(EnvTLSClientCert)
+		_ = os.Unsetenv(EnvTLSClientKey)
+		_ = os.Unsetenv(EnvTLSSSLVerify)
+	})
+
 	url := "https://1.2.3.4:5678"
 	caCert := "test/path/ca.pem"
 	caPath := "test/path"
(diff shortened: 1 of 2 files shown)


jwair:
I added @lornasong and @wilkermichael since we were just discussing these changes

api/client_test.go
```diff
@@ -1,71 +1,92 @@
...
-	assert.Equal(t, url, config.Addr)
-	assert.Equal(t, caCert, config.TLSConfig.CACert)
-	assert.Equal(t, caPath, config.TLSConfig.CAPath)
-	assert.Equal(t, clientCert, config.TLSConfig.ClientCert)
-	assert.Equal(t, clientKey, config.TLSConfig.ClientKey)
+	require.Equal(t, url, clientConfig.Addr)
+	require.Equal(t, caCert, clientConfig.TLSConfig.CACert)
+	require.Equal(t, caPath, clientConfig.TLSConfig.CAPath)
+	require.Equal(t, clientCert, clientConfig.TLSConfig.ClientCert)
+	require.Equal(t, clientKey, clientConfig.TLSConfig.ClientKey)
```
lornasong, line 42: Curious to learn about the reason to unify on 'require'. I know that some of the mix of 'require' and 'assert' is done deliberately. Here, I think  'assert' was used because if one Equal check fails, it's still worth continuing to check the others. The earlier 'require's are likely because they check test setup - if the test setup fails, then its not worth continuing the test.

api/client_test.go
```diff
@@ -1,71 +1,92 @@
...
+	require.Equal(t, DefaultAddress, clientConfig.Addr)
+	require.Equal(t, caCert, clientConfig.TLSConfig.CACert)
+	require.Equal(t, caPath, clientConfig.TLSConfig.CAPath)
+	require.Equal(t, clientCert, clientConfig.TLSConfig.ClientCert)
+	require.Equal(t, clientKey, clientConfig.TLSConfig.ClientKey)
+	require.Equal(t, DefaultSSLVerify, clientConfig.TLSConfig.SSLVerify)
+}
+
+func Test_ClientPort(t *testing.T) {
+	expectedPort := rand.Intn(10000)
```
lornasong, line 66: Looks like this test is using random data, which I haven't seen in other tests in this file. Do you mind sharing context behind using a random vs. hard-coded value?

lornasong (approved):
makes sense to me! left a couple of non-blocking question. the cleanup and consolidation in the tests improve the readability a lot

api/client_test.go
jwair: There isn't a good reason here for using a random value really. I will switch it to a constant to simplify it.

api/client_test.go
jwair: Ah thanks for explaining that, I wasn't aware that using `require` implies 'fail fast'. I thought the behavior was the same and we were getting it from two different packages so I wanted to unify that.
I will switch back to using asserts in places we don't need to fail fast.

api/client_test.go
jwair: Addressed in https://github.com/hashicorp/consul-terraform-sync/pull/663/commits/63d9a6d23abf5cec7b5928f73d6f56da5836fbcb

api/client_test.go
jwair: Addressed in https://github.com/hashicorp/consul-terraform-sync/pull/663/commits/63d9a6d23abf5cec7b5928f73d6f56da5836fbcb

lornasong: Gotcha, thanks for clarifying! Sounds good.

PR #665, opened 2022-02-04
Client Refactoring and Testing

jwair:
Added and improved unit tests for client config and different clients
Refactored client config and client to use URL instead of a mix of schema/addr/port
Moved logic for handling port flag outside of client
Added validation for address

### DIFF
--- a/api/client_test.go
+++ b/api/client_test.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"strconv"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 )
@@ -375,4 +376,46 @@ func Test_StatusClient_Overall(t *testing.T) {
 	o, err := c.Status().Overall()
 	assert.Nil(t, err)
 	assert.Equal(t, expectedOverallStatus, o)
+}
+
+func Test_WaitForAPI(t *testing.T) {
+	expectedOverallStatus := OverallStatus{TaskSummary: TaskSummary{
+		Status:  StatusSummary{Successful: 1},
+		Enabled: EnabledSummary{True: 2},
+	}}
+
+	bytes, err := json.Marshal(&expectedOverallStatus)
+	assert.NotEmpty(t, bytes)
+	assert.NoError(t, err)
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		_, err = fmt.Fprintf(w, string(bytes))
+		assert.NoError(t, err)
+	}))
+
+	defer server.Close()
+
+	clientConfig := BaseClientConfig()
+	clientConfig.URL = server.URL
+	c, err := NewClient(clientConfig, nil)
+	assert.NoError(t, err)
+
+	err = c.WaitForAPI(100 * time.Millisecond)
+	assert.NoError(t, err)
+}
+
+func Test_WaitForAPI_Timeout(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// do nothing
+	}))
+
+	defer server.Close()
+
+	clientConfig := BaseClientConfig()
+	clientConfig.URL = server.URL
+	c, err := NewClient(clientConfig, nil)
+	assert.NoError(t, err)
+
+	err = c.WaitForAPI(100 * time.Millisecond)
+	assert.Error(t, err)
 }
--- a/api/client_test.go
+++ b/api/client_test.go
@@ -109,14 +109,27 @@ func Test_Client_Task(t *testing.T) {
 }
 
 func Test_NewClient(t *testing.T) {
-	clientConfig := BaseClientConfig()
-	c, err := NewClient(clientConfig, nil)
+	t.Run("no TLS", func(t *testing.T) {
+		clientConfig := BaseClientConfig()
+		c, err := NewClient(clientConfig, nil)
 
-	assert.NotNil(t, c)
-	assert.NoError(t, err)
+		assert.NotNil(t, c)
+		assert.NoError(t, err)
+	})
+
+	t.Run("with TLS", func(t *testing.T) {
+		clientConfig := BaseClientConfig()
+		clientConfig.TLSConfig.ClientCert = "../testutils/certs/localhost_cert.pem"
+		clientConfig.TLSConfig.ClientKey = "../testutils/certs/localhost_key.pem"
+
+		c, err := NewClient(clientConfig, nil)
+
+		assert.NotNil(t, c)
+		assert.NoError(t, err)
+	})
 }
 
-func Test_NewClient_Error_URL(t *testing.T) {
+func Test_NewClient_Error(t *testing.T) {
 	t.Run("invalid scheme", func(t *testing.T) {
 		clientConfig := &ClientConfig{URL: "foo://bar"}
 		c, err := NewClient(clientConfig, nil)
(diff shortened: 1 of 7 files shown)


api/client_test.go
```diff
@@ -47,46 +75,347 @@ func Test_DefaultClientConfig_WithEnvVars(t *testing.T) {
...
+				IncludeEvents: tt.fields.IncludeEvents,
+				Status:        tt.fields.Status,
+				Run:           tt.fields.Run,
+			}
+			assert.Equalf(t, tt.want, q.Encode(), "Encode()")
+		})
+	}
+}
+
+func Test_StatusClient_Overall_BadResponse(t *testing.T) {
```
lornasong, line 271: Looks like there's a series of `Test_StatusClient_Overall_*` test cases with different test-servers. Suggestion - it seems like it would be possible to reorganize these into a single test func with a test-table that has the server as a field. I think it would reduce code and unify these test cases in a single function.

e.g.
```
cases := []struct {
	name   string
	server *httptest.Server
} {
	{
		"error with errorr message",
		httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			e := &ErrorResponse{Error: &ErrorObject{Message: "foo"}}
			bytes, err := json.Marshal(e)
			assert.NotEmpty(t, bytes)
			assert.NoError(t, err)

			w.WriteHeader(http.StatusInternalServerError)
			_, err = w.Write(bytes)
			assert.NoError(t, err)
		})),
	},
...
```

api/client_test.go
```diff
@@ -47,46 +75,347 @@ func Test_DefaultClientConfig_WithEnvVars(t *testing.T) {
...
+	clientConfig.URL = server.URL
+	c, err := NewClient(clientConfig, nil)
+	assert.NoError(t, err)
+
+	o, err := c.Status().Overall()
+	assert.Error(t, err)
+	assert.EqualValues(t, OverallStatus{}, o)
+}
+
+func Test_StatusClient_Overall(t *testing.T) {
```
lornasong, line 354: All these test functions for Overall Status client are really thorough! Since there's other client methods, I'm curious to know your thoughts on those. Note - not trying to imply that they should be tested here.

api/client_test.go
```diff
@@ -47,46 +75,347 @@ func Test_DefaultClientConfig_WithEnvVars(t *testing.T) {
...
+				Run:           tt.fields.Run,
+			}
+			assert.Equalf(t, tt.want, q.Encode(), "Encode()")
+		})
+	}
+}
+
+func Test_StatusClient_Overall_BadResponse(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		_, err := fmt.Fprintf(w, "foo")
```
lornasong, line 273: My IDE marks a linting error here:

```
printf-style function with dynamic format string and no further arguments should use print-style function instead 
```

Updating it to `fmt.Fprint(w, "foo")` (removing the 'f' after "print") resolves it

api/client_test.go
```diff
@@ -47,46 +75,347 @@ func Test_DefaultClientConfig_WithEnvVars(t *testing.T) {
...
+		Status:  StatusSummary{Successful: 1},
+		Enabled: EnabledSummary{True: 2},
+	}}
+
+	bytes, err := json.Marshal(&expectedOverallStatus)
+	assert.NotEmpty(t, bytes)
+	assert.NoError(t, err)
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		_, err = fmt.Fprintf(w, string(bytes))
```
lornasong, line 365: Similar linting error here with `fmt.Fprintf`

lornasong (commented):
I like the consolidation of port, address, and scheme to the single URL field. Really cleaned up things! Left a couple suggestions and questions around the tests to check my understanding

api/client_test.go
```diff
@@ -47,46 +75,347 @@ func Test_DefaultClientConfig_WithEnvVars(t *testing.T) {
...
+
+	t.Run("with TLS", func(t *testing.T) {
+		clientConfig := BaseClientConfig()
+		clientConfig.TLSConfig.ClientCert = "../testutils/certs/localhost_cert.pem"
+		clientConfig.TLSConfig.ClientKey = "../testutils/certs/localhost_key.pem"
+
+		c, err := NewClient(clientConfig, nil)
+
+		assert.NotNil(t, c)
+		assert.NoError(t, err)
```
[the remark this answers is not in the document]
jwair: Yes, there is nothing from outside we can check for TLS, so just confirm that creating the client with TLS configs does not cause issues. We know under the hood it take a slightly different code path, so this test will impact test-coverage as well.

api/client_test.go
jwair: I will try that. Initially I didn't do it because I thought the individual tests are a little too big, but putting the server in the table is a good idea.

api/client_test.go
jwair: If you are talking about Task() and Update(), then yes, I do plan on testing them as well, but I wanted to check with you because I think there's an opportunity to refactor them and change their signature a bit for consistency. I thought putting all those changes in a different PR would be best.
If I misunderstood your question, please elaborate :p

api/client_test.go
```diff
@@ -47,46 +75,347 @@ func Test_DefaultClientConfig_WithEnvVars(t *testing.T) {
...
+
+	t.Run("invalid host", func(t *testing.T) {
+		clientConfig := &ClientConfig{URL: "http://"}
+		c, err := NewClient(clientConfig, nil)
+
+		assert.Nil(t, c)
+		assert.Error(t, err)
+	})
+
+	t.Run("missing cert", func(t *testing.T) {
```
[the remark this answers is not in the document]
jwair: Addressed in https://github.com/hashicorp/consul-terraform-sync/pull/665/commits/808db068de5e422e1b0ce82a0f8a0c8374509a20

api/client_test.go
```diff
@@ -47,46 +75,347 @@ func Test_DefaultClientConfig_WithEnvVars(t *testing.T) {
...
+		clientConfig := BaseClientConfig()
+		clientConfig.TLSConfig.ClientCert = "../testutils/certs/localhost_cert.pem"
+
+		c, err := NewClient(clientConfig, nil)
+
+		assert.Nil(t, c)
+		assert.Error(t, err)
+	})
+
+	t.Run("missing key", func(t *testing.T) {
```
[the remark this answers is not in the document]
jwair: Addressed in https://github.com/hashicorp/consul-terraform-sync/pull/665/commits/808db068de5e422e1b0ce82a0f8a0c8374509a20

api/client_test.go
```diff
@@ -47,46 +75,347 @@ func Test_DefaultClientConfig_WithEnvVars(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			q := &QueryParam{
+				IncludeEvents: tt.fields.IncludeEvents,
+				Status:        tt.fields.Status,
+				Run:           tt.fields.Run,
+			}
```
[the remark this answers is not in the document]
jwair: Addressed in https://github.com/hashicorp/consul-terraform-sync/pull/665/commits/808db068de5e422e1b0ce82a0f8a0c8374509a20

api/client_test.go
jwair: Addressed in https://github.com/hashicorp/consul-terraform-sync/pull/665/commits/f5d431cceb101755a27a5dffa9b828e918129a3f

api/client_test.go
jwair: Addressed in https://github.com/hashicorp/consul-terraform-sync/commit/808db068de5e422e1b0ce82a0f8a0c8374509a20

api/client_test.go
jwair: Addressed in https://github.com/hashicorp/consul-terraform-sync/commit/808db068de5e422e1b0ce82a0f8a0c8374509a20

api/client_test.go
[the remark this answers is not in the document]
lornasong: Thanks for clarifying!

api/client_test.go
lornasong: Awesome, looks good. Thanks for trying it out.

api/client_test.go
lornasong: Gotcha. Yes, I was wondering if you had plans to test methods like Task() and Update(). I agree, a separate PR sounds like a good idea. Thanks for explaining.

lornasong (approved):
looks good to me! test coverage improvements looks great!

PR #666, opened 2022-02-04
Add basic compile test for weekly tests to CI

lornasong:
Some of our weekly tests have separate build tags.

This makes it easy to accidentally make a change so that they cannot compile
without noticing. Because these tests also run weekly, the issue may not be
discovered as late as a week later.

This change adds a compile check to be run in circleci when a branch is pushed
or on a merge to main.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -169,6 +169,33 @@ jobs:
           path: /tmp/benchmark_results
           destination: benchmark_results
 
+  compile_weekly_tests:
+    executor:
+      name: go
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - ct-modcache-v2-{{ checksum "go.mod" }}
+      - run:
+          name: Check that weekly tests can compile
+          command:
+            make compile-weekly-tests
+      - when:
+          condition:
+            or:
+              - equal: [ main, <<pipeline.git.branch>> ]
+              - matches: { pattern: "release/.+", value: <<pipeline.git.branch>> }
+          steps:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -202,6 +229,7 @@ workflows:
     jobs:
       - go-checks
       - terraform-checks
+      - compile_weekly
```
findkim, line 232: The job name `compile_weekly` and Make command `test-compile-weekly` confused me quite a bit to think this job is compiled weekly, instead of this compiles weekly tests. A small nit of a name suggestion: `weekly-job-checks`

findkim (approved):
Thanks for adding this! I remember how this was an issue we saw a few months ago too

.circleci/config.yml
lornasong: Thanks for the suggestion! That's a really good point - the name can be easily read in another way that is unintended.

🤔 the "weekly" word is kind of tricky. What are your thoughts on `compile-weekly-tests`? I'm afraid `weekly-job-checks` could be interpreted as weekly job-checks i.e. a job check that occurs weekly.

Other brainstormed ideas: `check-weekly-jobs`, `check-weekly-tests`

findkim: ha yeah I can see how the `weekly-job-checks` suggestion runs into the same issue now. `compile-weekly-tests` is more clear than our first two ideas!

lornasong: Awesome, I'll go with that then. Thanks for chatting about this!

PR #671, opened 2022-02-07
Asynchronous Delete Task API

mkam:
Changes task deletion via the API to be asynchronous. The delete API now marks a task for deletion, and the task is deleted in a separate goroutine. This goroutine also waits for the task to complete if it is active before deleting.

Previously, if a task was running and a user tried to delete it, the delete call would return a 409 Conflict and prompt the user to try to delete again later. This was not the ideal user experience because a user would have to wait for a task to finish running with no easy way to tell when that was. Additionally, if a task was flapping and running multiple times in a row, it would be difficult for a user to time the delete without getting a conflict.

The asynchronous deletion is accomplished with the following changes:

- Add a map to Drivers that keeps track of what tasks are marked for deletion
- Check in all three run methods (run dynamic, run scheduled, run via API) if a task is marked for deletion before setting the task to active and running
- Add a delete channel to coordinate from the API to the controller's Run loop that a task needs to be deleted
- Move the delete logic from the API to a method that is called by the controller after receiving the message from the delete channel. Wait (indefinitely) until the task is no longer active.
- Update the API to instead mark the task for deletion and alert the Run loop via the delete channel

With this new workflow, tasks will always be marked for deletion. If the task is not running, then it should delete immediately (though still asynchronously).

Also updated the CLI output and the HTTP response code (from `200 OK` to `202 Accepted`) to reflect the asynchronous behavior.

```
$ consul-terraform-sync task delete test-task
    Only 'yes' will be accepted to approve, enter 'no' or leave blank to reject

Enter a value: yes

==> Marking task 'test-task' for deletion...

==> Task 'test-task' is marked for deletion and will be deleted when not running.
```

Example logs from a task that is deleted after it is running:
```
2022-02-07T14:30:59.652-0600 [DEBUG] driver.terraform: change detected for task: task_name=delayed-task
2022-02-07T14:30:59.673-0600 [INFO]  ctrl: executing task: task_name=delayed-task
2022-02-07T14:31:00.608-0600 [INFO]  api: received request: request_id=1c5de089-c58b-f94b-1518-cde08f2a54f8 time=2022-02-07T14:31:00.608-0600 remote_ip=127.0.0.1:57588 uri=[/v1/tasks/delayed-task]() method=DELETE host=localhost:8558
2022-02-07T14:31:00.608-0600 [DEBUG] ctrl: task marked for deletion: task_name=delayed-task
2022-02-07T14:31:00.609-0600 [INFO]  api: request complete: request_id=1c5de089-c58b-f94b-1518-cde08f2a54f8 duration=554us status_code=202
2022-02-07T14:31:05.160-0600 [INFO]  ctrl: task completed: task_name=delayed-task
2022-02-07T14:31:05.165-0600 [DEBUG] ctrl: task deleted: task_name=delayed-task
```


driver/drivers.go
```diff
@@ -164,5 +168,23 @@ func (d *Drivers) Delete(taskName string) error {
...
+
+func (d *Drivers) MarkForDeletion(name string) {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	d.deletion[name] = true
+}
+
+func (d *Drivers) IsMarkedForDeletion(name string) bool {
+	d.mu.Lock()
+	defer d.mu.Unlock()
```
findkim, line 183: Since this method only deals with reading from the map, the mutex could be better leveraged using the read lock so multiple reads can occur at the same time. This way it will only block if there's an active write lock.

```suggestion
	d.mu.RLock()
	defer d.mu.RUnlock()
```

findkim (approved):
Great summary, the approach and execution look good

api/task_delete_test.go
```diff
@@ -37,13 +37,13 @@ func TestTaskDelete_DeleteTaskByName(t *testing.T) {
 			http.StatusNotFound,
 		},
 		{
-			"task_is_running",
+			"task_errored",
```
jwair, line 40: nit, I see that we're doing it here for consistency, but in general the name can include spaces. `gotest` handles them nicely when generating the test names the test cases within a unit test

mkam: I started preferring `_` over spaces since I found it was easier to search for the test code when I had a failure since go test prints it as `_`. But I also know we don't do this everywhere, so maybe something to settle on as a team!

controller/readwrite.go
```diff
@@ -449,6 +472,33 @@ func (rw *ReadWrite) runTask(ctx context.Context, d driver.Driver) error {
...
+func (rw *ReadWrite) deleteTask(ctx context.Context, name string) error {
+WAIT:
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			// wait for task to not be running
+			if !rw.drivers.IsActive(name) {
+				break WAIT
```
jwair, line 487: Is the use of label and break label absolutely necessary?
I haven't seen it yet in the code base (I haven't looked thoroughly) but it is generally strongly discouraged. Can the same behavior be achieved in some other way?

driver/drivers.go
```diff
@@ -164,5 +168,23 @@ func (d *Drivers) Delete(taskName string) error {
 	}
 
 	delete(d.drivers, taskName)
+	delete(d.deletion, taskName)
 	return nil
 }
+
+func (d *Drivers) MarkForDeletion(name string) {
```
jwair, line 175: Can you please add a unit test for this?

driver/drivers.go
```diff
@@ -164,5 +168,23 @@ func (d *Drivers) Delete(taskName string) error {
...
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	d.deletion[name] = true
+}
+
+func (d *Drivers) IsMarkedForDeletion(name string) bool {
```
jwair, line 181: Can you please add a unit test for this?

controller/readwrite.go
mkam: Wasn't aware that labels were not encouraged, thanks for the insight! I can figure something else out, maybe move this whole wait loop to its own function!

controller/readwrite_test.go
```diff
@@ -543,6 +543,102 @@ func TestReadWrite_OnceAndRun(t *testing.T) {
...
+		// Attempt to delete the active task
+		ch := make(chan error)
+		go func() {
+			err := ctrl.deleteTask(ctx, taskName)
+			ch <- err
+		}()
+
+		// Check that the task is not deleted while active
+		time.Sleep(500 * time.Millisecond)
+		_, exists := drivers.Get(taskName)
```
jwair, line 618: Nit, this type of blind-wait can be a source of flakeyness, can this check be turned into a small loop with a timeout?

mkam: A loop didn't seem as useful here because we're waiting for the go routine to have started and for that go routine to be waiting in its infinite loop. There's not really a way to check that we're in that loop, so I went with a generic short wait.

This shouldn't cause a false negative because if a task is deleted after 500 ms but before `SetInactive()` is called, then that's a true failure.

mkam:
@jwair We currently only have the [auto-generated client method](https://github.com/hashicorp/consul-terraform-sync/blob/main/api/oapigen/client.go#L126-L136), which is what the CLI uses. I don't think there's any particular reason why we didn't make an equivalent method in the [task lifecycle client](https://github.com/hashicorp/consul-terraform-sync/blob/main/api/task_lifecycle_client.go), so that's something I can add in a follow up PR.

PR #673, opened 2022-02-08
Deprecate 'services' field and 'service' block in config snippets

lornasong:
Upgrades our config snippets containing the deprecated `services` field and `service` block with the replacement or removes it completely.

Adds a placeholder e2e test to specifically test all config streamlining related deprecations

### DIFF
--- a/e2e/api_test.go
+++ b/e2e/api_test.go
@@ -89,7 +89,7 @@ func TestE2E_StatusEndpoints(t *testing.T) {
 					Status:    api.StatusSuccessful,
 					Enabled:   true,
 					Providers: []string{"fake-sync"},
-					Services:  []string{"api"},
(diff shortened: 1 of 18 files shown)


e2e/api_test.go
```diff
@@ -89,23 +89,23 @@ func TestE2E_StatusEndpoints(t *testing.T) {
 					Status:    api.StatusSuccessful,
 					Enabled:   true,
 					Providers: []string{"fake-sync"},
-					Services:  []string{"api"},
+					Services:  []string{},
```
lornasong, line 92: This is an empty array now since this task is no longer configured with the `services` field. This field will be deprecated as part of separate work.

lornasong:
rebased to latest on main

PR #674, opened 2022-02-08
Feature 522 remove task services field

wilkermichael:
# Summary
This PR 
1) removes the deprecated `task.services` from the create task API/CLI. 
2) handles all currently deprecated task fields in the CLI

As the API/CLI are a new feature being added, the `task.services` does not need to be supported going forward, and was only included since config streamlining was still undergoing development.

The API will not return an error as we currently don't error on unexpected fields, we just don't consume them. If erroring on unexpected fields is desired, then this will need to be an additional enhancement. The CLI however will handle the deprecated cases, and return an error along with a suggested update. It was important to do this for the CLI since the `task-file` closely matches the CTS `config.hcl`, so there are cases where a user might copy an paste a deprecated a config.

# Example CLI Outputs

## Task.Services field included without a condition

Here we provide a block that can be directly copy and pasted into an existing task

```
==> Error: unable to create request
    the 'services' field in the task block is not supported 

'services' in a task configuration are to be replaced with the following:
 * condition "services": if there is _no_ preexisting condition block configured in your task
 * module_input "services": if there is a preexisting condition block configured in your task

Replace the 'services' field with the following 'condition "services"' block

task {
  ...
  condition "services" {
    names=["api","web"]
  }
  ...
}

For more details and additional examples, please see:
https://consul.io/docs/nia/release-notes/0-5-0#deprecate-services-field
```

## Task.Services field included with a condition

Here we provide a block that can be directly copy and pasted into an existing task

```
consul-terraform-sync task create --task-file=deprecated/task_services_kv.hcl    
==> Error: unable to create request
    the 'services' field in the task block is not supported 

'services' in a task configuration are to be replaced with the following:
 * condition "services": if there is _no_ preexisting condition block configured in your task
 * module_input "services": if there is a preexisting condition block configured in your task

Replace the 'services' field with the following 'module_input' block

task {
  ...
  module_input "services" {
    names=["api","web"]
  }
  ...
}

For more details and additional examples, please see:
https://consul.io/docs/nia/release-notes/0-5-0#deprecate-services-field
```

## Task.Services field included with a services condition

Here we suggest the changes that need to be made to the existing condition

```
consul-terraform-sync task create --task-file=deprecated/task_services_con.hcl
==> Error: unable to create request
    the 'services' field in the task block is not supported 

'services' in a task configuration are to be replaced with the following:
 * condition "services": if there is _no_ preexisting condition block configured in your task
 * module_input "services": if there is a preexisting condition block configured in your task

Please add the suggested names field to your current 'condition "services"' block and assure that 'use_as_module_input' is true or not present

Suggested replacement:
|    task {
|  -   services =  ["api","web"]
|      condition "services"{
|  +     names = ["api","web"]
|  +     use_as_module_input = true
|        ...
|      }
|      ...
|    }


For more details and additional examples, please see:
https://consul.io/docs/nia/release-notes/0-5-0#deprecate-services-field
```

## Task.Source instead of Task.Module
```
consul-terraform-sync task create --task-file=deprecated/task_source.hcl
==> Error: unable to create request
    the 'source' field in the task block is not supported 

Please replace 'source' with 'module' in your task configuration.

Suggested replacement:
|    task {
|  -   source =  "./example-module"
|  +   module =  "./example-module"
|      ...
|    }

For more details and examples, please see:
https://consul.io/docs/nia/release-notes/0-5-0#deprecate-source-field
```

## Multiple invalid fields at once

We do not return on a single error when handling the deprecatations, but instead accumulate all errors before returning

```
==> Error: unable to create request
    the 'source' field in the task block is not supported 

Please replace 'source' with 'module' in your task configuration.

Suggested replacement:
|    task {
|  -   source =  "./example-module"
|  +   module =  "./example-module"
|      ...
|    }

For more details and examples, please see:
https://consul.io/docs/nia/release-notes/0-5-0#deprecate-source-field

==> Error: unable to create request
    the 'source_input' block in the task is not supported 

Please replace 'source_input' with 'module_input' in your task configuration.

Suggested replacement:
|    task {
|  -   source_input "consul-kv" {
|  +   module_input "consul-kv" {
|        ...
|      }
|      ...
|    }

For more details and examples, please see:
https://consul.io/docs/nia/release-notes/0-5-0#deprecate-source_input-block
```

part of #522


wilkermichael:
@jherschman I'm also tagging you to take a look at the example errors here that have been added.

api/oapigen/types.go
```diff
@@ -130,7 +130,6 @@ type Task struct {
 	ModuleInput  *ModuleInput  `json:"module_input,omitempty"`
 	Name         string        `json:"name"`
 	Providers    *[]string     `json:"providers,omitempty"`
-	Services     *[]string     `json:"services,omitempty"`
```
findkim, line 133: Since `task.services` is not supported in the API, could `task.condition` now be a required field?

wilkermichael: that's a really good point! I think it should be

wilkermichael:
@findkim ready for you to take a second look, I just added the required condition

api/representation.go
```diff
@@ -113,50 +113,48 @@ func (tr TaskRequest) ToTaskConfig() (config.TaskConfig, error) {
 	}
 
 	// Convert condition
-	if tr.Task.Condition != nil {
```
wilkermichael, line 120: This is a weird diff, but I just removed this if nil check

api/representation.go
```diff
@@ -280,47 +278,44 @@ func oapigenTaskFromConfigTask(tc config.TaskConfig) oapigen.Task {
 		}
 	}
 
-	if tc.Condition != nil {
```
wilkermichael, line 291: same here

command/task_create.go
```diff
@@ -197,3 +205,217 @@ func (c *taskCreateCommand) Run(args []string) int {
...
+    names=[%s]
+  }
+  ...
+}`
+
+func generateServiceFieldMsg(action string) string {
+	return fmt.Sprintf(servicesFieldMsg, action)
+}
+
+const servicesFieldMsg = `the 'services' field in the task block is not supported ` +
```
jherschman, line 346: @wilkermichael for all mentions of this phrase, I would suggest on changing the language to provide context to users that may have been used to using 'services'. The only change here i would suggest it to replace "not" with "no longer"  so the error message would read like the below:

`"the 'services' field in the task block is no longer supported" `

command/task_create.go
```diff
@@ -197,3 +205,217 @@ func (c *taskCreateCommand) Run(args []string) int {
...
+
+For more details and additional examples, please see:
+https://consul.io/docs/nia/release-notes/0-5-0#deprecate-services-field
+`
+
+func generateSourceFieldMsg(module string) string {
+	return fmt.Sprintf(sourceFieldMsg, module, module)
+}
+
+const sourceFieldMsg = `the 'source' field in the task block is not supported ` +
```
jherschman, line 363: @wilkermichael same comment as the above... for all mentions of this phrase, I would suggest on changing the language to provide context to users that may have been used to using 'source'. The only change here I would suggest is to replace "not" with "no longer" so the error message would read like the below:

`the 'source' field in the task block is no longer supported`

command/task_create.go
```diff
@@ -197,3 +205,217 @@ func (c *taskCreateCommand) Run(args []string) int {
...
+  }
+  ...
+}`
+
+func generateServiceModuleInputBlockAction(services []string) string {
+	list := `"` + strings.Join(services, `","`) + `"`
+	return fmt.Sprintf(servicesModuleInputBlockAction, list)
+}
+
+const servicesModuleInputBlockAction = `Replace the 'services' field with the following 'module_input' block
```
jherschman, line 332: For consistency, I would suggest updating this statement to add "Please" to match the other error messages mentioned. The suggested update would read as follows:

`Please replace the 'services' field with the following 'module_input' block`

jherschman (approved):
These error messages look great! Just made a few small suggestions, let me know if you have any questions!

findkim (approved):

PR #676, opened 2022-02-08
Refactored client, simplified client creation logic

jwair:


lornasong (approved):
nice consolidation of setting up the TLS config + http client

PR #677, opened 2022-02-08
Improve validate template signature and test names

mkam:
Some follow up improvements to the validate template changes (https://github.com/hashicorp/consul-terraform-sync/pull/662) that I realized after this merged.

- Changed to using our interface for hcat template instead of hcat.Template directly
- Renamed `testInvalidQueries` to `testInvalidTaskConfig` because this test can be used for any invalid config that fails to create, not just for invalid queries

### DIFF
--- a/driver/terraform.go
+++ b/driver/terraform.go
@@ -589,7 +589,7 @@ func (tf *Terraform) initTaskTemplate() error {
 
 // validateTemplate verifies that executing the fetch requests of
 // a template's dependencies does not error.
-func validateTemplate(t *hcat.Template, clients hcat.Looker) error {
+func validateTemplate(t templates.Template, clients hcat.Looker) error {
 	var outer_err error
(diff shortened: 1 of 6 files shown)


findkim (approved):

driver/terraform.go
```diff
@@ -589,7 +589,7 @@ func (tf *Terraform) initTaskTemplate() error {
 
 // validateTemplate verifies that executing the fetch requests of
 // a template's dependencies does not error.
-func validateTemplate(t *hcat.Template, clients hcat.Looker) error {
+func validateTemplate(t templates.Template, clients hcat.Looker) error {
```
jwair, line 592: Any specific reason why we switched from pass-by-reference to pass-by-value here?

findkim: I believe `template.Templates` is an interface, and so the underlying object would be passed by reference

PR #680, opened 2022-02-09
Update Task Status API response deprecations

lornasong:
In a previous PR I deprecated `Events.Config.Source` and introduced `Events.Config.Module` as a replacement in the Task Status API response: https://github.com/hashicorp/consul-terraform-sync/pull/618

Instead, we want to deprecate `Events.Config` and additional task-config related fields from the Task Status API response. This is because they are not status information.

This information is better suited for a Get Task API, which we will introduce to replace the response deprecations.

Changes (details in commit)
 - Manually revert https://github.com/hashicorp/consul-terraform-sync/pull/618
 - Add deprecation comments

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -77,7 +77,7 @@ func TestServe(t *testing.T) {
 					On("Events", mock.Anything, taskName).Return(map[string][]event.Event{}, nil)
 			},
 			http.StatusOK,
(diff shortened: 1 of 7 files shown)


lornasong:
Note: forgot to mention - I decided not to deprecate `TaskStatus.Enabled` since it seems like status-relevant information. Feedback welcome!

findkim (approved):

PR #682, opened 2022-02-10
Create task module no exist error

wilkermichael:
This PR fixes an issue where the attempted creation of a bad task, for example, one with a non existent module, would lead to the inability to then re-create or delete the task.

The cause of this was that because the validity of the module path was not checked until after the driver was created, the driver was not appropriately cleaned up before returning the error leading to a dangling watcher. This task adds cleanup steps if a task driver is successfully created, but then it errors later in the creation process

part of #522

### DIFF
--- a/controller/server.go
+++ b/controller/server.go
@@ -44,12 +44,19 @@ func (rw *ReadWrite) TaskCreate(ctx context.Context, taskConfig config.TaskConfi
 	d.SetBufferPeriod()
 
 	// Add the task driver to the driver list only after successful create
-	err = rw.drivers.Add(*taskConfig.Name, d)
+	name := *taskConfig.Name
+	err = rw.drivers.Add(name, d)
 	if err != nil {
 		return config.TaskConfig{}, err
 	}
 	conf, err := configFromDriverTask(d.Task())
 	if err != nil {
+		// Cleanup driver
+		err := rw.TaskDelete(ctx, name)
+		if err != nil {
+			rw.logger.Error("unable to cleanup task after error", "task_name", name)
+		}
+		// return top level error only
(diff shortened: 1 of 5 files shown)


controller/server.go
```diff
@@ -44,12 +44,19 @@ func (rw *ReadWrite) TaskCreate(ctx context.Context, taskConfig config.TaskConfi
 	d.SetBufferPeriod()
 
 	// Add the task driver to the driver list only after successful create
-	err = rw.drivers.Add(*taskConfig.Name, d)
+	name := *taskConfig.Name
+	err = rw.drivers.Add(name, d)
 	if err != nil {
 		return config.TaskConfig{}, err
```
mkam, line 50: I know adding a driver is not likely to error, but should we technically also be cleaning up here too?

e2e/command_test.go
```diff
@@ -722,6 +722,92 @@ task {
...
+			inputTask: fmt.Sprintf(`
+task {
+  name           = "%s"
+  description    = "Creates a new task"
+  module         = "./test_modules/local_instances_file"
+  providers      = ["local"]
+  services       = ["web"]
+  enabled = true
+}`, taskName),
+			isExpectExist: false,
```
mkam, line 763: Shouldn't this be true? This would be the task that is created successfully after the first bad one failed, right?

wilkermichael: yep, thanks for catching

controller/server.go
wilkermichael: yeah that's a good point, I'll add cleanup here as well

mkam (approved):
Looks good to me!

PR #685, opened 2022-02-11
Handle deprecated 'services' in API representation

lornasong:
Context:
 - Create Task API does not support task creation with `services` field
 - As a result, the `services` field is not supported in openapi
 - CTS config file still supports task creation with `services` field

The Get Task API needs to support returning `services` field information when
the task has a `services` field. Because the `services` field is not available
to the API, convert the `services` information to a condition / module_input.

### DIFF
--- a/api/representation_test.go
+++ b/api/representation_test.go
@@ -142,7 +142,7 @@ func TestRequest_oapigenTaskFromConfigTask(t *testing.T) {
 				Description:  config.String("test-description"),
 				Name:         config.String("test-name"),
 				Providers:    []string{"test-provider-1", "test-provider-2"},
-				Services:     []string{"api", "web"},
+				Services:     []string{},
 				Module:       config.String("path"),
 				Version:      config.String("test-version"),
 				BufferPeriod: config.DefaultBufferPeriodConfig(),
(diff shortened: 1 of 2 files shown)


api/representation_test.go
```diff
@@ -375,6 +375,64 @@ func TestRequest_oapigenTaskFromConfigTask(t *testing.T) {
...
+				ModuleInputs: &config.ModuleInputConfigs{
+					&config.ConsulKVModuleInputConfig{
+						ConsulKVMonitorConfig: config.ConsulKVMonitorConfig{
+							Path: config.String("fake-path"),
+						},
+					},
+				},
+				Condition: &config.ScheduleConditionConfig{Cron: config.String("*/10 * * * * * *")},
+			},
+			expectedRequest: oapigen.Task{
```
wilkermichael, line 422: `expectedRequest` probably doesn't really make sense anymore, maybe just `expectedTask` , `expectedOapigenTask`, `expectedConversion` etc.

wilkermichael (approved):
Looks good to me!

api/representation_test.go
lornasong: Yea, that's true. Renamed to `expected` to allow flexibility for future changes. Test tests a function, so actual/expected language seems straightforward

PR #686, opened 2022-02-12
refactor task lifecycle API/CLI documentation and instructions

wilkermichael:
As I work on the consul.io docs for create task CLI and API, I have done a brief refactor to better match what is in consul.io docs.

## Summary of Changes

- Add descriptions to JSON fields which are to be described in the [task consul.io api documentation](https://consul-e663wnf9f-hashicorp.vercel.app/docs/nia/api/tasks#task-object) being worked on by Lorna
- Add option description for `task-file` which indicates it is required. Previously there was no description being provided
- After a discussion with Lorna, a brief update to some of the deprecated field handling in the create task CLI

## Example change to `create task -help`

```shell-session
$ consul-terraform-sync task create -help
Usage: consul-terraform-sync task create [options] --task-file=<task config>

  Task Create is used to create a new task. It is not to be used for updating a task, it will not create a task if the
  task name already exists.

Options:
   //<removed some options for brevity>

  port 8558
    [Deprecated] The port to use for the Consul-Terraform-Sync API server, it is preferred to use the http-addr field instead.

  ssl-verify true
    Boolean to verify SSL or not. Set to true to verify SSL. This can also be specified using the CTS_SSL_VERIFY environment variable.

  task-file 
    [Required] A file containing the hcl or json definition of a task

Example:

// <removed for brevity>
```


wilkermichael:
@lornasong as per our discussion, I'm also tagging you to look at the updates to the CLI deprecation handling

command/task_create.go
```diff
@@ -79,15 +82,12 @@ func (c *taskCreateCommand) Synopsis() string {
 
 // Run runs the command
 func (c *taskCreateCommand) Run(args []string) int {
-	c.meta.setFlagsUsage(c.flags, args, c.Help())
```
lornasong, line 82: Just wanted to double-check removing this `setFlagsUsage()`. It looks like it causes the CLI to print the help text when someone tries to pass in an invalid flag

e.g.

```
build/consul-terraform-sync task create -task=task_file.hcl
==> Error: unsupported arguments in flags '-task=task_file.hcl'
    Please see --help information below for supported options:

Usage: consul-terraform-sync task create [options] -task-file=<task config>
...
```

command/task_create.go
```diff
@@ -265,7 +264,8 @@ func handleDeprecations(ui mcli.Ui, tc *config.TaskConfig) error {
 		if tc.Condition != nil {
 			switch tc.Condition.(type) {
 			case *config.ServicesConditionConfig:
-				action := generateServicesConditionBlockAction(tc.Services)
+				action := `"list of 'services' and 'condition "services"' block cannot both be configured. ` +
```
lornasong, line 267: It seems like the double-quotes at the start of the string might not be necessary

command/task_create.go
```diff
@@ -274,9 +274,18 @@ func handleDeprecations(ui mcli.Ui, tc *config.TaskConfig) error {
 				action := generateServiceModuleInputBlockAction(tc.Services)
 				ui.Output(generateServiceFieldMsg(action))
 			case *config.ScheduleConditionConfig:
-				action := generateServicesAction(tc.Services)
+				action := generateServiceModuleInputBlockAction(tc.Services)
 				ui.Output(generateServiceFieldMsg(action))
```
lornasong, line 278: It looks like all the other conditions output the same message. Would it be possible to consolidate this by replacing with a `default` block? Maybe there's something I'm not thinking of?

command/task_create.go
```diff
@@ -274,9 +274,18 @@ func handleDeprecations(ui mcli.Ui, tc *config.TaskConfig) error {
...
 			}
+		} else if tc.ModuleInputs != nil {
```
lornasong, line 280: Would we want to make this a separate if-statement rather than an else-if statement? That would allow returning messages for module input when a user configures a task with `condition "services"` and `module_input "services"`. (It would impact the final else-statement)

command/task_create.go
```diff
@@ -274,9 +274,18 @@ func handleDeprecations(ui mcli.Ui, tc *config.TaskConfig) error {
...
+			for _, si := range *tc.ModuleInputs {
+				switch si.(type) {
+				case *config.ServicesModuleInputConfig:
+					action := `"list of 'services' and 'module_input "services"' block cannot both be configured. ` +
```
lornasong, line 284: Similar here, looks like there's an extra double-quote at the beginning of the string

lornasong (commented):
thanks for adding the descriptions to the openapi yaml and all the deprecation messages for the create task cli! the messages with the different combination of configs makes sense to me. left a couple questions

command/task_create.go
wilkermichael: ah, yes, I'll add it back

command/task_create.go
wilkermichael: yes, you're right, they're all the same now. Good point.

command/task_create.go
```diff
@@ -265,19 +265,27 @@ func handleDeprecations(ui mcli.Ui, tc *config.TaskConfig) error {
 		if tc.Condition != nil {
 			switch tc.Condition.(type) {
 			case *config.ServicesConditionConfig:
-				action := generateServicesConditionBlockAction(tc.Services)
-				ui.Output(generateServiceFieldMsg(action))
-			case *config.CatalogServicesConditionConfig:
-				action := generateServiceModuleInputBlockAction(tc.Services)
+				action := `list of 'services' and 'condition "services"' block cannot both be configured. ` +
+					`Consider using the 'names' field under 'condition "services'`
```
lornasong, line 269: nit: looks like the last double-quote in 'condition "services"' got removed

lornasong (approved):
looks good to me!

PR #687, opened 2022-02-14
Revert back to validating template with hcat templates

mkam:
After I had changed to using our template interface instead of directly using a hcat template for template validation (https://github.com/hashicorp/consul-terraform-sync/pull/677), I realized that our custom notifiers that implement the template interface won't mark the template as dirty with `Notify(nil)`. I considered modifying the custom notifiers, but decided to revert this back to expecting `*hcat.Template` for more predictable behavior.

Made some additional improvements that I thought of as I was debugging an issue related to this change:
- Added some debug logs around the template validation
- Moved validation before registration so that we don't have to deregister on error
- Changed the variable names of the errors in ValidateTemplate for improved clarity


findkim (approved):
Ah good catch, missed this as well. It may be worthwhile to add a code comment around the `template.Template` interface for the Notify method that it is different than the `hcat.Template.Notify` method. It doesn't mark the template dirty and the underlying hcat.Template is still needed.

PR #696, opened 2022-02-16
Backport of Added `additionalProperties: false` to all objects in openapi.yaml into release/0.5.x

hc-github-team-consul-ecosystem:
## Backport

This PR is auto-generated from #695 to be assessed for backporting due to the inclusion of the label backport/0.5.

The below text is copied from the body of the original PR.

---

Update openapi spec so that when validating against a schema, any additional fields which are not part of the schema will fail validation.

This is done by adding the field `addtionalProperties: false` to all schema objects


Cnp11784 (commented):
Next step

PR #697, opened 2022-02-16
Clarify that infrastructure is not destroyed on task deletion

mkam:
Made it the second bullet point since I didn't want it overlooked!
```
 consul-terraform-sync task delete test-task
     - Deleting a task will not destroy the infrastructure managed by the task.
     - If the task is not running, it will be deleted immediately.
     - If the task is running, it will be deleted once it has completed.
    Only 'yes' will be accepted to approve, enter 'no' or leave blank to reject.

Enter a value: yes

==> Marking task 'test-task' for deletion...

==> Task 'test-task' has been marked for deletion and will be deleted when not running.
```


wilkermichael (approved):
Looks good @mkam

PR #703, opened 2022-02-17
Use custom dependency for validating template

mkam:
Using nil was having unintended consequences in the cases using
catalog-service condition when there are no services to start with.


mkam:
Well, this did not end up fixing it like I thought! Closing!

PR #718, opened 2022-02-23
Check that created task triggers as expected

mkam:


e2e/condition_catalog_service_test.go
```diff
@@ -529,6 +529,20 @@ task {
...
+	service = testutil.TestService{ID: "web-test-new", Name: "web-test-new", Tags: []string{"tag_b"}}
+	testutils.RegisterConsulService(t, srv, service, defaultWaitForRegistration)
+
+	// Check that created task does trigger
+	api.WaitForEvent(t, cts, taskName, registrationTime, defaultWaitForEvent)
+	count = eventCount(t, taskName, cts.Port())
+	expectedCount++
+	assert.Equal(t, expectedCount, count)
+	resourcesPath := filepath.Join(tempDir, taskName, resourcesDir)
+	validateServices(t, true, []string{"services_" + sharedService}, resourcesPath)
```
lornasong, line 543: I noticed there's a `"services_"` here unlike previous `validateServices()`. I wasn't able to figure out the context for it. Would you mind staring?

lornasong (approved):
looks good! left a non-blocking question

e2e/condition_catalog_service_test.go
mkam: That's due to the module itself adds this in front of the file name for the service! I don't think we check this very often so I can leave a comment here and for the tags file!

lornasong: Ohhh gotcha. Thank you for explaining that to me. Makes sense.

Issue #732, opened 2022-02-24
Service state change during task execution can cause locked state


wilkermichael:
### Describe the bug
If a task is executed, and a service which triggers that task changes state while the task is executing, the TFC driver enters a locked state. Terraform CLI does not lock up, however, changes which happens while the task is running are ignored. 

**Note**
- This only affects Tasks when a change to a resource, monitored by these Tasks, is made while these Tasks are running. Tasks monitoring different resources, and newly created tasks, regardless of which resource they're monitoring, will work as intended.

### Versions
**Consul Terraform Sync**
```
consul-terraform-sync-enterprise >= v0.5.0
```

**Consul Version**
```
Tested with Consul v1.11.3
```

**Terraform Version**
```
Tested with Terraform v1.1.6
```

### Expected Behavior
If a service changes state during task execution, where the task is triggered by the service which changes state, then after the task completes a new task execution should take place.

### Actual Behavior
CTS logs an event, and then will no longer trigger on service changes for that particular task:

```
2022-02-24T09:49:01.011-0800 [INFO]  ctrl: task completed: task_name=michael-github-task
2022-02-24T09:49:01.011-0800 [TRACE] ctrl: adding event: event="&Event{ID:633d74eb-94e1-deff-333b-a57e7ead7a1c, TaskName:michael-github-task, Success:true, StartTime:2022-02-24 09:48:34.148593 -0800 PST m=+22.755140190, EndTime:2022-02-24 09:49:01.011806 -0800 PST m=+49.618325080, EventError:%!s(*event.Error=<nil>), Config:&Config{Providers:[local], Services:[web api], Source:./git-module}}"
```

### Steps to Reproduce
1. Start CTS with a task that triggers on a service change
2. Change the state of the service (Deregister or Register, whatever is opposite of the current state)
3. While CTS is executing the task, change the state again
4. Once the event is logged, observe that no changes to task trigger the task

### Additional Context

- attempted to create a task with and without enabled buffer_period. Issue exists in both cases, buffer_period settings do not seem to have an effect.
- have not tested with tasks triggering on non-service conditions (catalog-services, consul-kv, etc.). So may effect tasks triggering on other conditions as well.

wilkermichael:
##OSS Research
[...]

PR #741, opened 2022-03-01
Wait for task to be inactive before running

mkam:
Fixes issue where a task triggering change that occurs while the task is currently running is ignored (https://github.com/hashicorp/consul-terraform-sync/issues/732). 

Prior to reverting a hcat change (https://github.com/hashicorp/consul-terraform-sync/pull/740), when using the TFC driver, this ignored run led to a bad state where the task will never be triggered by dependency changes.

**Background**
Previously (as in 0.4.3 and earlier), when any change was detected, we would try to [run all dynamic tasks](https://github.com/hashicorp/consul-terraform-sync/blob/v0.4.3/controller/readwrite.go#L79-L93) and [wait for them to complete](https://github.com/hashicorp/consul-terraform-sync/blob/v0.4.3/controller/readwrite.go#L134). If a change happened while a task was running, it would be executed after the current run had completed.

This also meant that if task_a was running, and a Consul change for task_b was made, task_b would not run until task_a was completed.

Now in 0.5.0, the watcher returns specifically what template needs updating, so we [only run the task that is monitoring the dependency change](https://github.com/hashicorp/consul-terraform-sync/blob/1ae6360e754c8dd3bb420bc9b7848e060ae236f8/controller/readwrite.go#L108-L115). We also run it as a goroutine, so there's no more waiting for the task to complete.

This has the advantage where task_b can still be triggered immediately even if task_a is currently running.

**Issue**
Because we're no longer waiting, we start a process to run the task even though it's currently running. The `runDynamicTask()` method checks that the task is active and returns without running, which is why the change is ignored completely. That `IsActive()` check was supposed to ignore runs triggered by the enable API when the task was already in progress.

**Solution**
If a task is active and currently running, wait for the task to complete and start a new run when the task becomes inactive. 

One caveat is that if a task has multiple changes while it's running, we would potentially run multiple times, even though it's not technically necessary. We still wouldn't run Terraform apply multiple times, though, because the template rendering should detect that there are no changes. This behavior is the same as we made the watcher changes, so it didn't seem necessary to optimize this now.

Fixes https://github.com/hashicorp/consul-terraform-sync/issues/732

### DIFF
--- a/controller/readwrite_test.go
+++ b/controller/readwrite_test.go
@@ -358,6 +358,44 @@ func TestReadWrite_runDynamicTask(t *testing.T) {
 		err := controller.runDynamicTask(context.Background(), d)
 		assert.NoError(t, err)
 	})
+
+	t.Run("active-task", func(t *testing.T) {
+		controller := newTestController()
+		controller.EnableTestMode()
+
+		ctx := context.Background()
+		d := new(mocksD.Driver)
+		taskName := "task"
+		mockDriver(ctx, d, enabledTestTask(t, taskName))
+		drivers := controller.drivers
+		drivers.Add(taskName, d)
+		drivers.SetActive(taskName)
+
+		// Attempt to run the active task
+		ch := make(chan error)
+		go func() {
+			err := controller.runDynamicTask(ctx, d)
+			ch <- err
+		}()
+
+		// Check that the task did not run while active
+		select {
+		case <-controller.taskNotify:
+			t.Fatal("task ran even though active")
+		case <-time.After(250 * time.Millisecond):
+			break
+		}
+
+		// Set task to inactive, wait for run to happen
+		drivers.SetInactive(taskName)
+		select {
+		case <-time.After(250 * time.Millisecond):
+			t.Fatal("task did not run after it became inactive")
+		case <-controller.taskNotify:
+			break
+		}
+	})
+
 }
 
 func TestReadWrite_runScheduledTask(t *testing.T) {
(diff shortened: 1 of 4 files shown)


e2e/e2e_test.go
```diff
@@ -502,6 +502,87 @@ func TestE2E_ConfigStreamlining_Deprecations(t *testing.T) {
...
+
+	// Trigger the other task while the other task is running
+	inactiveRegister := time.Now()
+	testutils.RegisterConsulService(t, srv,
+		testutil.TestService{Name: "web", ID: "web-1"}, defaultWaitForRegistration)
+
+	// Other task should complete without waiting for first task to complete
+	api.WaitForEvent(t, cts, inactiveTaskName, inactiveRegister, defaultWaitForEvent)
+	count := eventCount(t, inactiveTaskName, cts.Port())
+	assert.Equal(t, 2, count, "unexpected number of events")
```
mkam, line 565: Oh, I just realized I didn't add any checks to make sure the module output reflected the expected services! Events are probably sufficient, but I'll add it just in case!

controller/readwrite_test.go
```diff
@@ -638,6 +756,56 @@ func TestReadWrite_deleteTask(t *testing.T) {
...
+		mockD := new(mocksD.Driver)
+		mockD.On("TemplateIDs").Return(nil)
+		ctrl.drivers.Add(taskName, mockD)
+		ctrl.drivers.SetInactive(taskName)
+
+		// Wait for task to be inactive, should return immediately
+		err := ctrl.waitForTaskInactive(ctx, taskName)
+		require.NoError(t, err)
+	})
+}
```
wilkermichael, line 808: Really appreciate all the unit testing!

e2e/e2e_test.go
```diff
@@ -502,6 +502,87 @@ func TestE2E_ConfigStreamlining_Deprecations(t *testing.T) {
...
 }
 
+// TestE2E_TriggerTaskWhenActive tests making changes to monitored dependencies
+// while a task is running. It verifies:
+//
+// 1. a change to the same task will trigger the task after the run completes
+// 2. a change to a different task runs immediately
+// 3. a change after the task completes will still trigger the task to run
+//
+// https://github.com/hashicorp/consul-terraform-sync/issues/732
```
wilkermichael, line 512: One other test case I tried manually, is making sure that the task that applies is the latest change i.e.

- register a service (Change 1)
- CTS starts up task
- deregister a service (Change 2)
- register a service (Change 3)
- CTS completes first task
- CTS executes a new task for Change 3 not 2

Do you think this is worth automating? I verified that it behaves as expected manually

wilkermichael (approved):

wilkermichael:
@mkam great work!

e2e/e2e_test.go
wilkermichael: I also verified the buffering still works manually. One day we'll create those buffering tests ...

mkam: Yeah, I think it would be good to test that the latest change is used! Thinking about whether it should be a separate test or added to this one, though! 🤔

mkam: Added and went with a separate test!

PR #754, opened 2022-03-07
Improve driver test coverage

wilkermichael:
I went through the driver package and improved the test coverage by creating tests for public methods. I tried to stick to tests that could be written without a large refactor, we can revisit refactoring code later to be more testable (for example terraform_download.go)


driver/drivers_test.go
```diff
@@ -46,6 +54,10 @@ func TestDrivers_Add(t *testing.T) {
 				actual, ok := drivers.drivers[tc.taskName]
 				require.True(t, ok)
 				assert.Equal(t, tc.driver, actual)
+
+				for _, val := range md.templateIDs {
```
jwair, line 58: You might be able to use https://pkg.go.dev/github.com/stretchr/testify/assert#ElementsMatch

driver/terraform_test.go
```diff
@@ -107,6 +110,53 @@ func TestRenderTemplate(t *testing.T) {
 	}
 }
 
+func TestTerraform_Version(t *testing.T) {
+	TerraformVersion = &goVersion.Version{}
```
jwair, line 114: What is this line for?

driver/terraform_test.go
```diff
@@ -107,6 +110,53 @@ func TestRenderTemplate(t *testing.T) {
...
+		w.On("Deregister", mock.Anything).Return()
+		c.On("Plan", ctx).Return(true, nil).Once()
+		c.On("SetStdout", mock.Anything).Twice()
+
+		ctx = context.Background()
+		plan, err := tf.InspectTask(ctx)
+		assert.NoError(t, err)
+		require.Equal(t, "", plan.Plan)
+	})
+
```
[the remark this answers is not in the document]
jwair: Redundant line

templates/tftmpl/notifier/notifier.go
```diff
@@ -7,6 +7,8 @@ import (
 	"github.com/hashicorp/hcat/dep"
 )
 
+//go:generate mockery --name=Overrider --filename=overrider.go --output=../../../mocks/notifier
```
jwair, line 10: leftovers from development?

wilkermichael: Nope, this comment allows this generate to be picked up when you run `go generate ./...` from the root of the project. This way that's all you have to do to do all code generations for CTS!

https://go.dev/blog/generate

driver/drivers_test.go
wilkermichael: Oh that's cool! There are definitely a number of tests I could use that for. Thanks for finding that.

driver/terraform_test.go
wilkermichael: I'm initializing the global variable

wilkermichael: I'll add a comment, I'll also improve this:

```Go
func TestTerraform_Version(t *testing.T) {
	var err error
	TerraformVersion, err = goVersion.NewVersion("1.2")
	require.NoError(t, err)
	var tf Terraform
	s := tf.Version()
	assert.Equal(t, "1.2.0", s)
}
```

driver/drivers_test.go
wilkermichael: Although, I'm not sure if it makes sense in this case, I just want to see that map contains a particular value, I can add a comment

driver/terraform_test.go
```diff
@@ -768,10 +832,67 @@ func TestInitTask(t *testing.T) {
...
 
+func TestTerraform_DestroyTask(t *testing.T) {
+	var w mocksTmpl.Watcher
+	tf := Terraform{
+		watcher: &w,
+	}
+	ctx := context.Background()
+
+	w.On("Deregister", mock.Anything).Return().Once()
+	tf.DestroyTask(ctx)
```
mkam, line 850: Could we add an AssertExpectations here to check that deregister was called once?

driver/drivers_test.go
```diff
@@ -96,6 +108,81 @@ func TestDrivers_Get(t *testing.T) {
...
+			ok:     true,
+		},
+		{
+			name:   "driver does not exist",
+			tmplID: "no-exist",
+			driver: nil,
+			ok:     false,
+		},
+		{
+			name:   "driver does not exist",
```
mkam, line 144: Can you update the name of this case to distinguish it from the previous one?

driver/drivers.go
```diff
@@ -161,9 +161,9 @@ func (d *Drivers) Delete(taskName string) error {
 	}
 
 	// delete driver templates associated with task
-	for val, id := range d.driverTemplates {
-		if val == taskName {
-			delete(d.driverTemplates, id)
+	for k, v := range d.driverTemplates {
+		if v == taskName {
+			delete(d.driverTemplates, k)
```
mkam, line 166: Is there any potential issue that will occur from modifying the map you're iterating over? Not sure how that works in go.

driver/task_test.go
```diff
@@ -57,6 +59,186 @@ func TestNewClient(t *testing.T) {
...
+			}
+		})
+	}
+}
+
+func TestTask_Condition(t *testing.T) {
+	t.Parallel()
+
+	var task Task
+	task.condition = &config.ConsulKVConditionConfig{}
```
mkam, line 104: Is there any significance of using a Consul KV condition in particular here?

driver/task_test.go
```diff
@@ -57,6 +59,186 @@ func TestNewClient(t *testing.T) {
...
+
+func TestTask_IsEnabled(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name      string
+		isEnabled bool
+	}{
+		{
+			name:      "schedule condition",
```
mkam, line 172: These names need updating.

driver/task_test.go
```diff
@@ -57,6 +59,186 @@ func TestNewClient(t *testing.T) {
...
+func TestTask_Env(t *testing.T) {
+	var task Task
+	task.env = map[string]string{
+		"k1": "v1",
+		"k2": "v2",
+	}
+	env := task.Env()
+	assert.Equal(t, task.Env(), env)
+
+	// Make sure we made a deep copy
```
mkam, line 214: Really like that you added this check for deep copy!

mkam (approved):
Looks good to me, thanks for taking on this improvement! Left some minor questions and comments.

driver/terraform_test.go
wilkermichael: Ah, I thought `once` was sufficient, and that it meant it could only be called `once`. I misunderstood, good to know.

wilkermichael: I've added this in other places as well

driver/drivers.go
wilkermichael: yep, this is okay, and idiomatic as well: 

https://go.dev/doc/effective_go#for (see If you only need the first item in the range (the key or index), drop the second)

driver/task_test.go
wilkermichael: no, could be any condition

driver/terraform_test.go
mkam: Yeah, it can only be called once, but it can also be called zero times! We don't seem to do AssertExpectations very often, but I had seen it here or there.

PR #770, opened 2022-03-18
Prepare for autocomplete

wilkermichael:


wilkermichael:
@mkam ignore the tests failing, I wanted to split the PRs to make it a bit more digestible, the second PR will be here: https://github.com/hashicorp/consul-terraform-sync/pull/771

api/task_get.go
```diff
@@ -6,6 +6,29 @@ import (
 	"github.com/hashicorp/consul-terraform-sync/logging"
 )
 
+func (h *TaskLifeCycleHandler) GetAllTasks(w http.ResponseWriter, r *http.Request) {
```
wilkermichael, line 9: Need to add a function comment

api/task_lifecycle_client.go
```diff
@@ -56,9 +59,10 @@ func (c *TaskLifecycleClient) Scheme() string {
 	return c.url.Scheme
 }
 
-// CreateTask takes a task request and run option and sends this information to the client. It then returns
+// CreateTaskSimple takes a task request and run option and sends this information to the client. It then returns
```
wilkermichael, line 62: I made this change so as to not override the interface, this is useful for mocking down the line.

api/task_get.go
```diff
@@ -6,6 +6,29 @@ import (
...
+	ctx := r.Context()
+	requestID := requestIDFromContext(ctx)
+	logger := logging.FromContext(r.Context()).Named(getTaskSubsystemName)
+	logger.Trace("get task request")
+
+	// Retrieve task if it exists
+	taskConfigs, err := h.ctrl.Tasks(ctx)
+	if err != nil {
+		logger.Trace("tasks not found", "error", err)
+		sendError(w, r, http.StatusNotFound, err)
```
mkam, line 22: If Tasks() returns an error, it looks like it would be because of a parsing error, so a 500 response feels more appropriate.

https://github.com/hashicorp/consul-terraform-sync/blob/main/controller/server.go#L189

api/task_get_test.go
```diff
@@ -69,3 +69,81 @@ func TestTaskLifeCycleHandler_GetTaskByName(t *testing.T) {
 		})
 	}
 }
+
+func TestTaskLifeCycleHandler_GetTasks(t *testing.T) {
```
mkam, line 73: nit: This corresponds to GetAllTasks rather than GetTasks, but also maybe we could just call that method GetTasks?

api/task_lifecycle_client.go
mkam: Could we brainstorm a different name? "Simple" to me doesn't really reflect the behavior and implies that this method is a subset of the full create functionality. 

Alternatively, do we have a different solution for mocking than implementing the ClientInterface? It feels like we're not using it properly if we have to rename this method to make it work.

command/commands_test.go
```diff
@@ -0,0 +1,35 @@
+package command
+
+import (
+	"testing"
+
+	"github.com/mitchellh/cli"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_Commands(t *testing.T) {
```
mkam, line 10: Nice addition of a unit test!

mkam (commented):
Really exciting stuff with the CLI changes! My main point of discussion for this is my comment on the `CreateTaskSimple`, want to evaluate either the method name or the use of the interface.

I agree with your assessment on leaving out pagination and query parameter filtering for now, but let's double check with product and make sure everyone's on the same page!

I've also been thinking about if we'd ever want to return a subset of the task configuration rather than all the values, but I think that could also be a future feature that's controlled with a query parameter. An example could be if we need a list of just task names to output in the CLI and there is a large amount of tasks, we might want to optimize by only returning names.

wilkermichael:
> 

Yes, I totally agree with eventually have the ability to query and support field returns

The simpler way:
example:
I want all task names where tasks are enabled:

GET /v1/tasks/names?enabled=true

The more complex way would be to have some sort of query field in the body of the request, something like what elasticsearch has:

```JSON 
{
    "fields": ["user", "message", ...],
    "query": ...,
    "size": ...
}
```

api/task_get.go
wilkermichael: yes, that makes sense, it's acceptable to return an empty list

api/task_get_test.go
wilkermichael: yeah, I renamed the method to `GetAllTasks` from `GetTasks` and it didn't catch everything, I'll fix this and others if I see them

api/task_lifecycle_client.go
wilkermichael: hmm I may have another idea I can run by you here

mkam (approved):
Looks good to me! Glad you found a good solution to the interfacing!

mkam:
@wilkermichael I know we're also discussing alternative options for `run`, I'm good for you to merge this as is into the temporary branch! If we want to change command, that can be a follow up that we do before merging to `main`.

command/task_create.go
```diff
@@ -170,6 +170,7 @@ func (c *taskCreateCommand) Run(args []string) int {
 		return ExitCodeError
 	}
 
+	taskResp := resp.JSON201
```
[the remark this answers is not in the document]
wilkermichael: In the yaml definition of the endpoints we have response objects paired with response codes. This returns back the response for a 201, which is the task response.

Issue #772, opened 2022-03-22
Support for Terraform Cloud Agents


mkam:
#### Description
Add support for creating Terraform Cloud workspaces configured with [Terraform Cloud Agents](https://www.terraform.io/cloud-docs/agents).

CTS currently only supports "agent" execution mode for existing workspaces, which requires users to manually set the execution mode and agent pool via the Terraform Cloud UI. With this change, users should be able to configure a task with the execution mode and agent pool name or ID. CTS will then either create a new workspace or update an existing workspace with this configuration.

PR #777, opened 2022-03-28
Add Terraform Cloud workspace configuration to tasks

mkam:
Add an enterprise-only configuration `task.terraform_cloud_workspace` that handles TFC workspace configurations that are specific to the task (rather than to all tasks, which is handled by the [`driver.workspaces` configuration](https://www.consul.io/docs/nia/configuration#workspaces)). The new configuration is needed to support configuring Terraform Cloud agents with the Terraform Cloud driver. The requirements for the configuration are as followed:

1. Only support "remote" and "agent".
2. Allow agent pool to be identified by either the ID or the name. If both are provided, ID takes precedence.
3. At least one agent pool config is required to be configured if the execution mode is "agent".
4. Agent pool configs should not be set if the execution mode is not "agent".

Example:
```
task {
  name = "mkam_agent_test"
  module = "mkam/hello/cts"
  condition "services" {
    names = ["web", "api"]
  }
  terraform_cloud_workspace {
    execution_mode = "agent"
    agent_pool_name = "test_agent_pool"
  }
}
```

The existing enterprise-only config `task.terraform_version`, which controls the Terraform version for the workspace, will be added to `task.terraform_cloud_workspace` in a follow up PR and deprecated. 

Part of https://github.com/hashicorp/consul-terraform-sync/issues/772

### DIFF
--- a/config/task.go
+++ b/config/task.go
@@ -67,6 +67,11 @@ type TaskConfig struct {
 	// when using CTS OSS or the Terraform driver.
 	TFVersion *string `mapstructure:"terraform_version"`
 
+	// The workspace configurations to use for the task when configured with CTS
+	// enterprise and the Terraform Cloud driver. This option is not supported
+	// when using CTS OSS or the Terraform driver.
+	TFCWorkspace *TerraformCloudWorkspaceConfig `mapstructure:"terraform_cloud_workspace"`
+
 	// BufferPeriod configures per-task buffer timers.
 	BufferPeriod *BufferPeriodConfig `mapstructure:"buffer_period"`
 
@@ -120,6 +125,10 @@ func (c *TaskConfig) Copy() *TaskConfig {
 	o.Version = StringCopy(c.Version)
 
 	o.TFVersion = StringCopy(c.TFVersion)
+
+	if c.TFCWorkspace != nil {
+		o.TFCWorkspace = c.TFCWorkspace.Copy()
+	}
 
 	o.BufferPeriod = c.BufferPeriod.Copy()
 
@@ -192,6 +201,10 @@ func (c *TaskConfig) Merge(o *TaskConfig) *TaskConfig {
 
 	if o.TFVersion != nil {
 		r.TFVersion = StringCopy(o.TFVersion)
+	}
+
+	if o.TFCWorkspace != nil {
+		r.TFCWorkspace = r.TFCWorkspace.Merge(o.TFCWorkspace)
 	}
 
 	if o.BufferPeriod != nil {
@@ -267,6 +280,11 @@ func (c *TaskConfig) Finalize(globalBp *BufferPeriodConfig, wd string) {
 	if c.Version == nil {
 		c.Version = String("")
 	}
+
+	if c.TFCWorkspace == nil {
+		c.TFCWorkspace = &TerraformCloudWorkspaceConfig{}
+	}
+	c.TFCWorkspace.Finalize()
 
 	if c.TFVersion == nil {
 		c.TFVersion = String("")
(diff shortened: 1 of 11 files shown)


config/terraform_cloud_workspace.go
```diff
@@ -0,0 +1,139 @@
...
+
+// TerraformCloudWorkspaceConfig is an enterprise-only configuration that controls
+// workspace attributes that are specific to a task.
+type TerraformCloudWorkspaceConfig struct {
+	ExecutionMode *string `mapstructure:"execution_mode"`
+	AgentPoolID   *string `mapstructure:"agent_pool_id"`
+	AgentPoolName *string `mapstructure:"agent_pool_name"`
+}
+
+func DefaultTerraformCloudWorkspaceConfig() *TerraformCloudWorkspaceConfig {
```
mkam, line 19: Wasn't sure if this should be called `DefaultTerraformCloudWorkspaceConfig()` or `EmptyTerraformCloudWorkspaceConfig()`. Open to suggestions!

wilkermichael: I think Default is okay, makes it easier to add default values in the future should we extend the config

config/terraform_cloud_workspace.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// Finalize ensures that the receiver contains no nil pointers.
+func (c *TerraformCloudWorkspaceConfig) Finalize() {
+	if c == nil { // config is not required, return early
+		return
+	}
+
+	if c.ExecutionMode == nil {
+		c.ExecutionMode = String("remote")
```
wilkermichael: can we add this string as a constant? "DefaultExecutionMode" or something to that effect?

config/terraform_cloud_workspace_test.go
```diff
@@ -0,0 +1,316 @@
...
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			c := tc.c.Copy()
+			assert.Equal(t, tc.c, c)
+		})
+	}
+}
```
wilkermichael, line 41: add a new line between test functions

config/terraform_cloud_workspace_test.go
```diff
@@ -0,0 +1,316 @@
...
+			&TerraformCloudWorkspaceConfig{
+				ExecutionMode: String("remote"),
+				AgentPoolID:   String(""),
+				AgentPoolName: String(""),
+			},
+		},
+		{
+			"fully_configured",
+			&TerraformCloudWorkspaceConfig{
+				ExecutionMode: String("remote"),
```
wilkermichael, line 162: since "remote" is the default value that's being set by the function, maybe a better test would be to use a different string (i.e. agent) here. Just for future proofing

config/terraform_cloud_workspace.go
wilkermichael: you got rid of this in the next commit, so all good

config/task.go
```diff
@@ -353,6 +371,12 @@ func (c *TaskConfig) Validate() error {
 			"version within the Terraform driver block", *c.Name)
 	}
 
+	if c.TFCWorkspace != nil && !c.TFCWorkspace.isEmpty() {
```
wilkermichael, line 374: So is the idea that this would be removed for CTS Enterprise?

Maybe we handle switching this on/off similar to how we handle the dev license, where a global flag is set only if building with the enterprise build tag? Not sure what the best way to handle it is, but common code would lead to the least merge conflicts... especially since this file will be changed particularly often

wilkermichael (approved):
Great work Melissa!

config/task.go
wilkermichael: Btw, not a directive or anything, just curious what you were thinking :)

config/terraform_cloud_workspace_test.go
mkam: I changed from not setting remote as default, but that's still a valid point in case we change this in the future!

config/task.go
mkam: Yeah, I'm following what we've done for `terraform_version` which is to fail here always in OSS but check it in enterprise. Definitely could lead to merge conflicts in the future, but only if we're modifying code around this if statement.

I think licensing is a different case to here because it's adding a feature that doesn't exist in OSS rather than changing the behavior. There's still probably a good way to utilize build tags, though, I just don't have enough experience to know how without doing some research first. We can discuss these details offline!

For context, the reason I included this configuration in OSS was to prevent merge conflicts in our API representation methods. Our OSS openapi spec needs to include the enterprise-only attributes so that enterprise users can use it as well, and we had the precedence of `terraform_version` in the OSS config, so this seemed like less of a pain than having to resolve the API files. Again, something we can rethink if needed!

wilkermichael: ok, that sounds good to me

Can see how it goes, and always refactor later if we need to

PR #781, opened 2022-03-30
Introduce "state store" and refactor events

lornasong:
This PR introduces the "state store" and starts converting CTS to using state store by first moving events into the state store. Subsequent PRs will move configuration and anything else into the state store.

Overview of changes:
 - Introduce state.Store and state.InMemoryStore. Consumed by baseController (in anticipation of configuration being moved into state store)
 - Refactor event.Event into sub-directory of new state directory
    - event/event.go => state/event/event.go
    - event/event_test.go => state/event/event_test.go
 - Refactor event.Store into new state directory and make it private
    - event/store.go => state/event_store.go
    - event/store_test.go => state/event_store_test.go

See individual commits for more details and explanations

Particularly open to feedback to where event and eventStore should live


controller/server_test.go
```diff
@@ -413,13 +412,13 @@ func TestServer_TaskUpdate(t *testing.T) {
...
 
-		events := ctrl.store.Read(taskName)
+		events := ctrl.state.GetTaskEvents(taskName)
 		assert.Len(t, events, 1)
 	})
 }
 
 // mockDriver sets up a mock driver with the happy path for all methods
-func mockDriver(ctx context.Context, d *mocks.Driver, task *driver.Task) {
+func mockDriver(ctx context.Context, d *mocksD.Driver, task *driver.Task) {
```
lornasong, line 421: Context fyi - mocks.Driver was imported twice in this file which was giving me a linting error. I removed `mocks` in the imports and kept `mocksD`:

```diff
-	mocks "github.com/hashicorp/consul-terraform-sync/mocks/driver"
 	mocksD "github.com/hashicorp/consul-terraform-sync/mocks/driver"
```

state/event_store.go
```diff
@@ -1,31 +1,33 @@
...
+
+	"github.com/hashicorp/consul-terraform-sync/state/event"
 )
 
 const defaultEventCountLimit = 5
 
-// Store stores events
-type Store struct {
+// eventStore stores events
+type eventStore struct {
```
mkam, line 13: It makes sense to me to make this private! Just wondering if the naming will be confusing because we've got the store interface and the in memory store implementation of that interface, but event store is kind of its own thing. 🤔  It probably won't be a big deal since it's private anyway, so not strongly advocating one way or another!

mkam (approved):
Changes look good to me! I like where you've put event and event store, your reasoning makes a lot of sense to me. Did leave a comment on event store around the naming but it's not blocking!

state/event_store.go
lornasong: That's a good point! Yea `eventStore` looks like it is an implementation of the `Store` interface.

Noting for posterity: we chatted about this offline and decided to go with `eventStorage` for now. This new name is still a bit close to `Store` but it will make it clear that `eventStorage` doesn't implement `Store` interface. Down the line, we can also readjust the `Store` name as well as we see how it gets consumed by other code. Thanks!

PR #783, opened 2022-04-01
Create Consul Client

### LINKED ISSUE #784: Add Consul Client
#### Description
Add a Consul Client to CTS that can be reused for interactions with Consul

#### Proposed Tasks:
- Set up a Consul interface / refactor existing Consul-related code
- Wrap all Consul requests in retries with logs
- Audit for existing Consul requests to use this
- Define criteria by which new methods may be added to the Consul Client Interface
  - each request to Consul is:
     - Retried
     - Logged at DEBUG-level
     - Easily mocked

wilkermichael:
## Summary
This feature adds a Consul Client and Consul Client Interface to CTS. It also unifies behaviour between the new client and the client used by hcat and enhances the exponential backoff functionality of the retry package.

All Consul calls now
1. are capable of retries
2. are capable of infinite retries (set maxRetries = -1)
3. support a maximum wait time between retries 

## Scope
- This feature adds support to extend the Consul Client interface for future use cases, currently it supports the use case of:
   - Identifying the SKU of Consul being connected to (Enterprise or OSS)
   - Retrieving a signed license from Consul
-  This feature enhances the retry package to support indefinite retries and a maximum wait time between retries

## Not In Scope
- Currently the maxRetries is set to be finite, it makes no attempt to configure the maxRetries to be infinite, but adds support for possible configuration at a later date.

## Future Considerations:
Currently the Consul client to be used by CTS, and the Consul client used by HCAT are different clients. In the future, a refactor could be done to make these clients the same.

Resolves #784


retry/retry_test.go
```diff
@@ -46,33 +46,39 @@ func TestWithRetry(t *testing.T) {
...
-			"happy path: retry once, success on retry (2)",
+			"happy path: retry once, success on retry (1)",
+			2, // max retries is 2, but will succeed on first retry
 			1,
-			2,
+			nil,
+		},
+		{
+			"happy path: infinite retries, success on retry (3)",
+			-1, // max retries is 2, but will succeed on first retry
```
wilkermichael, line 61: need to fix this commnet

retry/retry_test.go
```diff
@@ -82,14 +88,57 @@ func TestWithRetry(t *testing.T) {
...
+			if tc.expected == nil {
+				assert.NoError(t, err)
+			} else {
+				assert.Equal(t, tc.expected.Error(), err.Error())
+			}
+		})
+	}
+}
+
+func TestWithRetry_WithDelay(t *testing.T) {
```
wilkermichael, line 110: I'm going to remove this test case, I don't think it add a lot of value

controller/watcher.go
```diff
@@ -54,29 +54,39 @@ func newWatcher(conf *config.Config) (*hcat.Watcher, error) {
...
 	logger := logging.Global().Named(hcatLogSystemName)
-	if retryCount > max {
+	if max >= 0 && retryCount > max {
 		logger.Error("error connecting with Consul even after retries", "retries", retryCount)
 		return false, 0 * time.Second
 	}
 
 	random := rand.New(rand.NewSource(time.Now().UnixNano()))
-	wait := retry.WaitTime(uint(retryCount), random)
+	wait := wr.waitFunc(uint(retryCount), random) // max wait time of 15 minutes
```
wilkermichael, line 89: going to remove the literal max wait time here, and instead reference the constant for future proofing

client/consul.go
```diff
@@ -0,0 +1,159 @@
...
+
+	sku, ok := parseSKU(info)
+	if !ok {
+		return "", errors.New("unable to parse sku")
+	}
+	return sku, nil
+}
+
+func parseSKU(info Self) (string, bool) {
+	v, ok := info["Config"]["Version"].(string)
```
lornasong, line 146: I'm not sure if there's any sensitive data in SKU? I think it could be useful to log or return an error containing the actual vs. expected SKU format for debugging purposes

retry/retry.go
```diff
@@ -11,27 +11,31 @@ import (
...
 type Retry struct {
-	maxRetry uint // doesn't count initial try
+	maxRetry int // doesn't count initial try, set to -1 for infinite retries
 	random   *rand.Rand
 	testMode bool
 	logger   logging.Logger
 }
 
-// NewRetry initializes a retry handler
+// NewRetry initializes a maxRetry handler
```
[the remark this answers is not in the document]
wilkermichael: I think this was a case of bad auto-refactoring from goland/me

I'll fix and correct these, you're right in that it should be `retry`

retry/retry_test.go
```diff
@@ -46,33 +46,39 @@ func TestWithRetry(t *testing.T) {
...
 	cases := []struct {
 		name      string
-		retry     uint
+		maxRetry  int
 		successOn int
 		expected  error
 	}{
 		{
-			"happy path: retry once, success on retry (2)",
+			"happy path: retry once, success on retry (1)",
```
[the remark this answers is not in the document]
wilkermichael: yeah, I was trying to clarify the language here a bit, so let me know if it makes sense. I don't count the first try as a "retry" so from what I can tell it:

Fails on the first try
Succeeds on the first retry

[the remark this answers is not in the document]
wilkermichael: I'm kinda basing this on the comment I found about maxRetries which distinguishes tries from retries:

// maxRetry is *retries*, so maxRetry of 2 means 3 total tries.

retry/retry_test.go
```diff
@@ -162,6 +211,18 @@ func TestWaitTime(t *testing.T) {
 			4,
 			6,
 		},
+		{
+			"max attempt minimum",
+			10,
+			900,
+			900,
```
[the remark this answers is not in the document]
wilkermichael: yep, good catch

retry/retry_test.go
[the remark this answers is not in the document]
wilkermichael: alternatively, I could write this as:

"happy path: try twice, success on retry (1)"

client/consul.go
```diff
@@ -0,0 +1,159 @@
+package client
```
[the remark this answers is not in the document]
wilkermichael: yeah, that makes sense, we can talk more about this after this PR closes

client/consul.go
```diff
@@ -0,0 +1,159 @@
...
+type ConsulClient struct {
+	*consulapi.Client
+	retry  retry.Retry
+	logger logging.Logger
+}
+
+// Self represents the response body from Consul /v1/agent/self API endpoint.
+// Care must always be taken to do type checks when casting, as structure could
+// potentially change over time.
+type Self = map[string]map[string]interface{}
```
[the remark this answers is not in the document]
wilkermichael: sure, sounds good

client/consul.go
```diff
@@ -0,0 +1,159 @@
...
+
+func parseSKU(info Self) (string, bool) {
+	v, ok := info["Config"]["Version"].(string)
+	if !ok {
+		return "", ok
+	}
+
+	ver, vErr := version.NewVersion(v)
+	if vErr != nil {
+		return "", false
```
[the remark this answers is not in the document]
wilkermichael: I pulled this from nomad: https://github.com/hashicorp/nomad/blob/d2b69b61da452ef71a14b245134186a4c3b48a75/command/agent/consul/self.go#L14

So I didn't really look at this code too deeply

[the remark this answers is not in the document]
wilkermichael: I can look into this

controller/watcher_test.go
```diff
@@ -0,0 +1,54 @@
...
+			count := 0
+			for true {
+				isSuccessRetry, _ := wr.retryConsul(count)
+				if !isSuccessRetry || count > tc.breakLimit {
+					break
+				}
+				count++
+			}
+
+			assert.Equal(t, tc.expectedRetries, count-1)
```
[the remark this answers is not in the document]
wilkermichael: yep, that's right. I'll add this comment

client/consul.go
wilkermichael: I'll look into this

client/consul.go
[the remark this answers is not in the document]
wilkermichael: So yeah, this will return an error with context around why it errored, I can restructure this to be a string, err return

retry/retry_test.go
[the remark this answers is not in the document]
lornasong: I see what you mean, the language is a bit tricky here.

> happy path: try twice, success on retry (1)

^ I like your suggestion. I find it easy to understand and I like the alignment between "twice" and "2" (set the line below)

client/consul.go
```diff
@@ -0,0 +1,159 @@
...
+			license = ""
+			return err
+		}
+		return nil
+	}, desc)
+
+	return license, err
+}
+
+// GetSKU queries Consul for information about itself, it then
```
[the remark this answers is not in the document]
lornasong: Interesting, thanks for sharing. I'm not familiar with "SKU" in an engineering context - I might be the only one though? What are your thoughts on a name that doesn't require prior knowledge? Maybe something like AgentInfo? In particular I think it would be helpful to have something more descriptive for users in the error message and logging

client/consul.go
wilkermichael: So it looks like the response doesn't include sensitive information (returns hidden, for example DebugConfig.ACLInitialManagementToken)


Here's and example:
```JSON
{
    "Config": {
        "Datacenter": "dc1",
        "PrimaryDatacenter": "dc1",
        "NodeName": "mwilkerson-C02FX04LMD6R",
        "NodeID": "ab7ec898-1bce-a45c-94b3-7e1ab5d12bf3",
        "Revision": "e319d7ed",
        "Server": true,
        "Version": "1.11.3"
    },
    "DebugConfig": {
        "ACLEnableKeyListPolicy": false,
        "ACLInitialManagementToken": "hidden",
        "ACLResolverSettings": {
            "ACLDefaultPolicy": "allow",
            "ACLDownPolicy": "extend-cache",
            "ACLPolicyTTL": "30s",
            "ACLRoleTTL": "0s",
            "ACLTokenTTL": "30s",
            "ACLsEnabled": false,
            "Datacenter": "dc1",
            "EnterpriseMeta": {},
            "NodeName": "mwilkerson-C02FX04LMD6R"
        },
        "ACLTokenReplication": false,
        "ACLTokens": {
            "ACLAgentRecoveryToken": "hidden",
            "ACLAgentToken": "hidden",
            "ACLDefaultToken": "hidden",
            "ACLReplicationToken": "hidden",
            "DataDir": "",
            "EnablePersistence": false,
            "EnterpriseConfig": {}
        },
        "ACLsEnabled": false,
        "AEInterval": "1m0s",
        "AdvertiseAddrLAN": "127.0.0.1",
        "AdvertiseAddrWAN": "127.0.0.1",
        "AdvertiseReconnectTimeout": "0s",
        "AllowWriteHTTPFrom": [],
        "AutoConfig": {
            "Authorizer": {
                "AllowReuse": false,
                "AuthMethod": {
                    "ACLAuthMethodEnterpriseFields": {},
                    "Config": {
                        "BoundAudiences": null,
                        "BoundIssuer": "",
                        "ClaimMappings": null,
                        "ClockSkewLeeway": 0,
                        "ExpirationLeeway": 0,
                        "JWKSCACert": "",
                        "JWKSURL": "",
                        "JWTSupportedAlgs": null,
                        "JWTValidationPubKeys": null,
                        "ListClaimMappings": null,
                        "NotBeforeLeeway": 0,
                        "OIDCDiscoveryCACert": "",
                        "OIDCDiscoveryURL": ""
                    },
                    "Description": "",
                    "DisplayName": "",
                    "EnterpriseMeta": {},
                    "MaxTokenTTL": "0s",
                    "Name": "Auto Config Authorizer",
                    "RaftIndex": {
                        "CreateIndex": 0,
                        "ModifyIndex": 0
                    },
                    "TokenLocality": "",
                    "Type": "jwt"
                },
                "ClaimAssertions": [],
                "Enabled": false
            },
            "DNSSANs": [],
            "Enabled": false,
            "IPSANs": [],
            "IntroToken": "hidden",
            "IntroTokenFile": "",
            "ServerAddresses": []
        },
        "AutoEncryptAllowTLS": false,
        "AutoEncryptDNSSAN": [],
        "AutoEncryptIPSAN": [],
        "AutoEncryptTLS": false,
        "AutopilotCleanupDeadServers": true,
        "AutopilotDisableUpgradeMigration": false,
        "AutopilotLastContactThreshold": "200ms",
        "AutopilotMaxTrailingLogs": 250,
        "AutopilotMinQuorum": 0,
        "AutopilotRedundancyZoneTag": "",
        "AutopilotServerStabilizationTime": "10s",
        "AutopilotUpgradeVersionTag": "",
        "BindAddr": "127.0.0.1",
        "Bootstrap": false,
        "BootstrapExpect": 0,
        "CAFile": "",
        "CAPath": "",
        "Cache": {
            "EntryFetchMaxBurst": 2,
            "EntryFetchRate": 1.7976931348623157e+308,
            "Logger": null
        },
        "CertFile": "",
        "CheckDeregisterIntervalMin": "1m0s",
        "CheckOutputMaxSize": 4096,
        "CheckReapInterval": "30s",
        "CheckUpdateInterval": "5m0s",
        "Checks": [],
        "ClientAddrs": [
            "127.0.0.1"
        ],
        "ConfigEntryBootstrap": [],
        "ConnectCAConfig": {},
        "ConnectCAProvider": "",
        "ConnectEnabled": true,
        "ConnectMeshGatewayWANFederationEnabled": false,
        "ConnectSidecarMaxPort": 21255,
        "ConnectSidecarMinPort": 21000,
        "ConnectTestCALeafRootChangeSpread": "0s",
        "ConsulCoordinateUpdateBatchSize": 128,
        "ConsulCoordinateUpdateMaxBatches": 5,
        "ConsulCoordinateUpdatePeriod": "100ms",
        "ConsulRaftElectionTimeout": "52ms",
        "ConsulRaftHeartbeatTimeout": "35ms",
        "ConsulRaftLeaderLeaseTimeout": "20ms",
        "ConsulServerHealthInterval": "10ms",
        "DNSARecordLimit": 0,
        "DNSAddrs": [
            "tcp://127.0.0.1:8600",
            "udp://127.0.0.1:8600"
        ],
        "DNSAllowStale": true,
        "DNSAltDomain": "",
        "DNSCacheMaxAge": "0s",
        "DNSDisableCompression": false,
        "DNSDomain": "consul.",
        "DNSEnableTruncate": false,
        "DNSMaxStale": "87600h0m0s",
        "DNSNodeMetaTXT": true,
        "DNSNodeTTL": "0s",
        "DNSOnlyPassing": false,
        "DNSPort": 8600,
        "DNSRecursorStrategy": "sequential",
        "DNSRecursorTimeout": "2s",
        "DNSRecursors": [],
        "DNSSOA": {
            "Expire": 86400,
            "Minttl": 0,
            "Refresh": 3600,
            "Retry": 600
        },
        "DNSServiceTTL": {},
        "DNSUDPAnswerLimit": 3,
        "DNSUseCache": false,
        "DataDir": "",
        "Datacenter": "dc1",
        "DefaultQueryTime": "5m0s",
        "DevMode": true,
        "DisableAnonymousSignature": true,
        "DisableCoordinates": false,
        "DisableHTTPUnprintableCharFilter": false,
        "DisableHostNodeID": true,
        "DisableKeyringFile": true,
        "DisableRemoteExec": true,
        "DisableUpdateCheck": false,
        "DiscardCheckOutput": false,
        "DiscoveryMaxStale": "0s",
        "EnableAgentTLSForChecks": false,
        "EnableCentralServiceConfig": true,
        "EnableDebug": true,
        "EnableLocalScriptChecks": false,
        "EnableRemoteScriptChecks": false,
        "EncryptKey": "hidden",
        "EncryptVerifyIncoming": true,
        "EncryptVerifyOutgoing": true,
        "EnterpriseRuntimeConfig": {},
        "ExposeMaxPort": 21755,
        "ExposeMinPort": 21500,
        "GRPCAddrs": [
            "tcp://127.0.0.1:8502"
        ],
        "GRPCPort": 8502,
        "GossipLANGossipInterval": "100ms",
        "GossipLANGossipNodes": 3,
        "GossipLANProbeInterval": "100ms",
        "GossipLANProbeTimeout": "100ms",
        "GossipLANRetransmitMult": 4,
        "GossipLANSuspicionMult": 3,
        "GossipWANGossipInterval": "100ms",
        "GossipWANGossipNodes": 3,
        "GossipWANProbeInterval": "100ms",
        "GossipWANProbeTimeout": "100ms",
        "GossipWANRetransmitMult": 4,
        "GossipWANSuspicionMult": 3,
        "HTTPAddrs": [
            "tcp://127.0.0.1:8500"
        ],
        "HTTPBlockEndpoints": [],
        "HTTPMaxConnsPerClient": 200,
        "HTTPMaxHeaderBytes": 0,
        "HTTPPort": 8500,
        "HTTPResponseHeaders": {},
        "HTTPSAddrs": [],
        "HTTPSHandshakeTimeout": "5s",
        "HTTPSPort": -1,
        "HTTPUseCache": true,
        "KVMaxValueSize": 524288,
        "KeyFile": "hidden",
        "LeaveDrainTime": "5s",
        "LeaveOnTerm": false,
        "Logging": {
            "EnableSyslog": false,
            "LogFilePath": "",
            "LogJSON": false,
            "LogLevel": "DEBUG",
            "LogRotateBytes": 0,
            "LogRotateDuration": "0s",
            "LogRotateMaxFiles": 0,
            "Name": "",
            "SyslogFacility": "LOCAL0"
        },
        "MaxQueryTime": "10m0s",
        "NodeID": "ab7ec898-1bce-a45c-94b3-7e1ab5d12bf3",
        "NodeMeta": {},
        "NodeName": "mwilkerson-C02FX04LMD6R",
        "PidFile": "",
        "PrimaryDatacenter": "dc1",
        "PrimaryGateways": [],
        "PrimaryGatewaysInterval": "30s",
        "RPCAdvertiseAddr": "tcp://127.0.0.1:8300",
        "RPCBindAddr": "tcp://127.0.0.1:8300",
        "RPCConfig": {
            "EnableStreaming": true
        },
        "RPCHandshakeTimeout": "5s",
        "RPCHoldTimeout": "7s",
        "RPCMaxBurst": 1000,
        "RPCMaxConnsPerClient": 100,
        "RPCProtocol": 2,
        "RPCRateLimit": -1,
        "RaftBoltDBConfig": {
            "NoFreelistSync": false
        },
        "RaftProtocol": 3,
        "RaftSnapshotInterval": "30s",
        "RaftSnapshotThreshold": 16384,
        "RaftTrailingLogs": 10240,
        "ReadReplica": false,
        "ReconnectTimeoutLAN": "0s",
        "ReconnectTimeoutWAN": "0s",
        "RejoinAfterLeave": false,
        "RetryJoinIntervalLAN": "30s",
        "RetryJoinIntervalWAN": "30s",
        "RetryJoinLAN": [],
        "RetryJoinMaxAttemptsLAN": 0,
        "RetryJoinMaxAttemptsWAN": 0,
        "RetryJoinWAN": [],
        "Revision": "e319d7ed",
        "SegmentLimit": 64,
        "SegmentName": "",
        "SegmentNameLimit": 64,
        "Segments": [],
        "SerfAdvertiseAddrLAN": "tcp://127.0.0.1:8301",
        "SerfAdvertiseAddrWAN": "tcp://127.0.0.1:8302",
        "SerfAllowedCIDRsLAN": [],
        "SerfAllowedCIDRsWAN": [],
        "SerfBindAddrLAN": "tcp://127.0.0.1:8301",
        "SerfBindAddrWAN": "tcp://127.0.0.1:8302",
        "SerfPortLAN": 8301,
        "SerfPortWAN": 8302,
        "ServerMode": true,
        "ServerName": "",
        "ServerPort": 8300,
        "Services": [
            {
                "Address": "",
                "Check": {
                    "AliasNode": "",
                    "AliasService": "",
                    "Body": "",
                    "CheckID": "",
                    "DeregisterCriticalServiceAfter": "0s",
                    "DockerContainerID": "",
                    "FailuresBeforeCritical": 0,
                    "FailuresBeforeWarning": 0,
                    "GRPC": "",
                    "GRPCUseTLS": false,
                    "H2PING": "",
                    "H2PingUseTLS": false,
                    "HTTP": "",
                    "Header": {},
                    "Interval": "0s",
                    "Method": "",
                    "Name": "",
                    "Notes": "",
                    "OutputMaxSize": 0,
                    "ProxyGRPC": "",
                    "ProxyHTTP": "",
                    "ScriptArgs": [],
                    "Shell": "",
                    "Status": "",
                    "SuccessBeforePassing": 0,
                    "TCP": "",
                    "TLSServerName": "",
                    "TLSSkipVerify": false,
                    "TTL": "0s",
                    "Timeout": "0s"
                },
                "Checks": [],
                "Connect": null,
                "EnableTagOverride": false,
                "EnterpriseMeta": {},
                "ID": "",
                "Kind": "",
                "Meta": {},
                "Name": "web",
                "Port": 80,
                "Proxy": null,
                "SocketPath": "",
                "TaggedAddresses": {},
                "Tags": [
                    "example"
                ],
                "Token": "hidden",
                "Weights": {
                    "Passing": 1,
                    "Warning": 1
                }
            },
            {
                "Address": "",
                "Check": {
                    "AliasNode": "",
                    "AliasService": "",
                    "Body": "",
                    "CheckID": "",
                    "DeregisterCriticalServiceAfter": "0s",
                    "DockerContainerID": "",
                    "FailuresBeforeCritical": 0,
                    "FailuresBeforeWarning": 0,
                    "GRPC": "",
                    "GRPCUseTLS": false,
                    "H2PING": "",
                    "H2PingUseTLS": false,
                    "HTTP": "",
                    "Header": {},
                    "Interval": "0s",
                    "Method": "",
                    "Name": "",
                    "Notes": "",
                    "OutputMaxSize": 0,
                    "ProxyGRPC": "",
                    "ProxyHTTP": "",
                    "ScriptArgs": [],
                    "Shell": "",
                    "Status": "",
                    "SuccessBeforePassing": 0,
                    "TCP": "",
                    "TLSServerName": "",
                    "TLSSkipVerify": false,
                    "TTL": "0s",
                    "Timeout": "0s"
                },
                "Checks": [],
                "Connect": null,
                "EnableTagOverride": false,
                "EnterpriseMeta": {},
                "ID": "",
                "Kind": "",
                "Meta": {},
                "Name": "api",
                "Port": 9000,
                "Proxy": null,
                "SocketPath": "",
                "TaggedAddresses": {},
                "Tags": [
                    "test"
                ],
                "Token": "hidden",
                "Weights": {
                    "Passing": 1,
                    "Warning": 1
                }
            }
        ],
        "SessionTTLMin": "0s",
        "SkipLeaveOnInt": true,
        "StartJoinAddrsLAN": [],
        "StartJoinAddrsWAN": [],
        "SyncCoordinateIntervalMin": "15s",
        "SyncCoordinateRateTarget": 64,
        "TLSCipherSuites": [],
        "TLSMinVersion": "tls12",
        "TLSPreferServerCipherSuites": false,
        "TaggedAddresses": {
            "lan": "127.0.0.1",
            "lan_ipv4": "127.0.0.1",
            "wan": "127.0.0.1",
            "wan_ipv4": "127.0.0.1"
        },
        "Telemetry": {
            "AllowedPrefixes": [],
            "BlockedPrefixes": [],
            "CirconusAPIApp": "",
            "CirconusAPIToken": "hidden",
            "CirconusAPIURL": "",
            "CirconusBrokerID": "",
            "CirconusBrokerSelectTag": "",
            "CirconusCheckDisplayName": "",
            "CirconusCheckForceMetricActivation": "",
            "CirconusCheckID": "",
            "CirconusCheckInstanceID": "",
            "CirconusCheckSearchTag": "",
            "CirconusCheckTags": "",
            "CirconusSubmissionInterval": "",
            "CirconusSubmissionURL": "",
            "Disable": false,
            "DisableCompatOneNine": false,
            "DisableHostname": false,
            "DogstatsdAddr": "",
            "DogstatsdTags": [],
            "FilterDefault": true,
            "MetricsPrefix": "consul",
            "PrometheusOpts": {
                "CounterDefinitions": [],
                "Expiration": "0s",
                "GaugeDefinitions": [],
                "Name": "consul",
                "Registerer": null,
                "SummaryDefinitions": []
            },
            "StatsdAddr": "",
            "StatsiteAddr": ""
        },
        "TranslateWANAddrs": false,
        "TxnMaxReqLen": 524288,
        "UIConfig": {
            "ContentPath": "/ui/",
            "DashboardURLTemplates": {},
            "Dir": "",
            "Enabled": true,
            "MetricsProvider": "",
            "MetricsProviderFiles": [],
            "MetricsProviderOptionsJSON": "",
            "MetricsProxy": {
                "AddHeaders": [],
                "BaseURL": "",
                "PathAllowlist": []
            }
        },
        "UnixSocketGroup": "",
        "UnixSocketMode": "",
        "UnixSocketUser": "",
        "UseStreamingBackend": true,
        "VerifyIncoming": false,
        "VerifyIncomingHTTPS": false,
        "VerifyIncomingRPC": false,
        "VerifyOutgoing": false,
        "VerifyServerHostname": false,
        "Version": "1.11.3",
        "VersionPrerelease": "",
        "Watches": []
    },
    "Coord": {
        "Vec": [
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0
        ],
        "Error": 1.5,
        "Adjustment": 0,
        "Height": 0.00001
    },
    "Member": {
        "Name": "mwilkerson-C02FX04LMD6R",
        "Addr": "127.0.0.1",
        "Port": 8301,
        "Tags": {
            "acls": "0",
            "build": "1.11.3:e319d7ed",
            "dc": "dc1",
            "ft_fs": "1",
            "ft_si": "1",
            "id": "ab7ec898-1bce-a45c-94b3-7e1ab5d12bf3",
            "port": "8300",
            "raft_vsn": "3",
            "role": "consul",
            "segment": "",
            "vsn": "2",
            "vsn_max": "3",
            "vsn_min": "2",
            "wan_join_port": "8302"
        },
        "Status": 1,
        "ProtocolMin": 1,
        "ProtocolMax": 5,
        "ProtocolCur": 2,
        "DelegateMin": 2,
        "DelegateMax": 5,
        "DelegateCur": 4
    },
    "Stats": {
        "agent": {
            "check_monitors": "0",
            "check_ttls": "0",
            "checks": "0",
            "services": "2"
        },
        "build": {
            "prerelease": "",
            "revision": "e319d7ed",
            "version": "1.11.3"
        },
        "consul": {
            "acl": "disabled",
            "bootstrap": "false",
            "known_datacenters": "1",
            "leader": "true",
            "leader_addr": "127.0.0.1:8300",
            "server": "true"
        },
        "raft": {
            "applied_index": "15670",
            "commit_index": "15670",
            "fsm_pending": "0",
            "last_contact": "0",
            "last_log_index": "15670",
            "last_log_term": "2",
            "last_snapshot_index": "0",
            "last_snapshot_term": "0",
            "latest_configuration": "[{Suffrage:Voter ID:ab7ec898-1bce-a45c-94b3-7e1ab5d12bf3 Address:127.0.0.1:8300}]",
            "latest_configuration_index": "0",
            "num_peers": "0",
            "protocol_version": "3",
            "protocol_version_max": "3",
            "protocol_version_min": "0",
            "snapshot_version_max": "1",
            "snapshot_version_min": "0",
            "state": "Leader",
            "term": "2"
        },
        "runtime": {
            "arch": "amd64",
            "cpu_count": "16",
            "goroutines": "108",
            "max_procs": "16",
            "os": "darwin",
            "version": "go1.17.5"
        },
        "serf_lan": {
            "coordinate_resets": "0",
            "encrypted": "false",
            "event_queue": "1",
            "event_time": "2",
            "failed": "0",
            "health_score": "0",
            "intent_queue": "0",
            "left": "0",
            "member_time": "1",
            "members": "1",
            "query_queue": "0",
            "query_time": "1"
        },
        "serf_wan": {
            "coordinate_resets": "0",
            "encrypted": "false",
            "event_queue": "0",
            "event_time": "1",
            "failed": "0",
            "health_score": "0",
            "intent_queue": "0",
            "left": "0",
            "member_time": "1",
            "members": "1",
            "query_queue": "0",
            "query_time": "1"
        }
    },
    "Meta": {
        "consul-network-segment": ""
    },
    "xDS": {
        "SupportedProxies": {
            "envoy": [
                "1.20.1",
                "1.19.1",
                "1.18.4",
                "1.17.4"
            ]
        },
        "Port": 8502
    }
}
```

wilkermichael: I think it would be okay to include the response if the keys can't be found?

retry/retry_test.go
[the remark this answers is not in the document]
wilkermichael: yes, although that (2) is the number of maxRetries, so doesn't actually pair with the number of tries as retry (2) would be try (3)

client/consul.go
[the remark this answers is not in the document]
wilkermichael: agentInfo isn't right either though, since we aren't getting all the info, we just want to know if it's Enterprise or OSS.

Version doesn't really work either, since we don't care about the number... I know we use "repo" in our release guide, but I don't think that really makes sense here. I can't think of anything other than SKU here.

client/consul.go
```diff
@@ -0,0 +1,159 @@
...
+
+// GetLicense queries Consul for a signed license, and returns it if available
+func (c *ConsulClient) GetLicense(ctx context.Context, q *consulapi.QueryOptions) (string, error) {
+	c.logger.Debug("getting license")
+
+	desc := "consul client get license"
+	var err error
+	var license string
+
+	err = c.retry.Do(ctx, func(context.Context) error {
```
jwair, line 104: nit, this is hard to read IMO

client/consul.go
```diff
@@ -0,0 +1,159 @@
...
+	v, ok := info["Config"]["Version"].(string)
+	if !ok {
+		return "", ok
+	}
+
+	ver, vErr := version.NewVersion(v)
+	if vErr != nil {
+		return "", false
+	}
+	if strings.Contains(ver.Metadata(), ConsulEnterpriseSKU) {
```
jwair, line 155: Nit, space out code blocks for better readability. The second `if` can be confused with an `else` based on how the code is spaced.

client/consul.go
wilkermichael: sure, I can make this it's own separate function

```Go
	f := func(context.Context) error {
		license, err = c.Operator().LicenseGetSigned(q)
		if err != nil {
			license = ""
			return err
		}
		return nil
	}

	err = c.retry.Do(ctx, f, desc)
```

client/consul_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestSelf_SKU(t *testing.T) {
+	t.Parallel()
+
```
jwair, line 11: Would this be better as a table?

mocks/readme.md
```diff
@@ -0,0 +1,17 @@
+# Mock Code Generation
```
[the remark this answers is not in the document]
jwair: +1

retry/retry.go
```diff
@@ -11,27 +11,31 @@ import (
 	"github.com/pkg/errors"
 )
 
-const taskSystemName = "task"
+const (
+	taskSystemName = "task"
+	maxWaitTime    = 15 * float64(time.Minute) // 15 minutes
```
jwair, line 16: Please put the unit in the variable name, when `maxWaitTime` is later used, it is not clear if it is in mins or seconds.
Also, is it possible to use a type that is better suited for time such as duration?

client/consul_test.go
wilkermichael: yeah, I just copied the test from [Nomad ](https://github.com/hashicorp/nomad/blob/d2b69b61da452ef71a14b245134186a4c3b48a75/command/agent/consul/self_test.go#L17) so didn't bother touching it. I can refactor it into a table.

retry/retry.go
```diff
@@ -11,27 +11,31 @@ import (
...
+)
 
 // Retry handles executing and retrying a function
 type Retry struct {
-	maxRetry uint // doesn't count initial try
+	maxRetry int // doesn't count initial try, set to -1 for infinite retries
```
jwair, line 21: might be clearer to achieve this with a `*int` type, so it will just be set to `nil` to indicate no max retries

retry/retry.go
wilkermichael: I guess it's technically nanoseconds. I didn't want to change to use duration since the WaitTime(*) algorithm uses floats

retry/retry.go
```diff
@@ -48,7 +52,7 @@ func (r Retry) Do(ctx context.Context, f func(context.Context) error, desc strin
 	for {
 		select {
 		case <-ctx.Done():
-			r.logger.Info("stopping retry", "description", desc)
+			r.logger.Info("stopping maxRetry", "description", desc)
```
jwair, line 55: This is an info level log message, so it won't only be read be developers. `maxRetry` is a variable name that doesn't mean much for an 'operator'

jwair: Was this incorrectly changed as part of a search and replace?

retry/retry.go
jwair: why is it singular `retry` as opposed to `retries`?

retry/retry.go
wilkermichael: it should be retry, this was a mistake Lorna identified as a bad automatic refactor by myself and goland.

This is fixed

retry/retry.go
```diff
@@ -84,22 +89,26 @@ func (r Retry) waitTime(attempt uint) int {
 	return WaitTime(attempt, r.random)
 }
 
-// WaitTime calculates the wait time based off the attempt number based off
-// exponential backoff with a random delay.
+// WaitTime calculates the wait time in nanoseconds based off the attempt number based off
+// exponential backoff with a random delay. It caps at the constant maxWaitTime.
 func WaitTime(attempt uint, random *rand.Rand) int {
```
jwair, line 94: Add unit to the name please, return type is an int so it is not clear what the unit is when we use it in the code later.
Even better, can we use a type more suitable for time instead of returning an int?

retry/retry.go
wilkermichael: To me nil is the absence of something, so I equate nil with 0 which would be `no retries`

This was my reasoning for using -1 for infinite retries. I've also seen -1 used this way before

wilkermichael: Also I not sure why it was maxRetry, I left it from its original creation

retry/retry.go
wilkermichael: I'll change the name, and look into refactoring the WaitTime to use a duration

client/consul.go
mkam: If the key can't be found, though, does a user gain anything from seeing this full response? Especially since it's dense with information.

wilkermichael: Maybe a debug log for our purposes?

retry/retry.go
jwair: We don't have to change the method signature, we can define this a duration, and get the total nanoseconds from it right where we call it (assuming we don't call it in many places)

client/consul.go
```diff
@@ -0,0 +1,170 @@
...
+	isEnterprise, err := isConsulEnterprise(ctx, info)
+	if err != nil {
+		return false, fmt.Errorf("unable to parse sku: %v", err)
+	}
+	return isEnterprise, nil
+}
+
+func isConsulEnterprise(ctx context.Context, info ConsulAgentConfig) (bool, error) {
+	logger := logging.FromContext(ctx)
+	v, ok := info["Config"]["Version"].(string)
```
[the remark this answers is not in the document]
wilkermichael: @jwair this is an interesting goism that John brought to my attention last week. This will actually return false if the keys don't exist and if the casting to `string` were to fail

retry/retry_test.go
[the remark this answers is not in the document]
lornasong: gotcha, makes sense

client/consul.go
```diff
@@ -0,0 +1,170 @@
...
+	err = c.retry.Do(ctx, f, desc)
+
+	return license, err
+}
+
+// IsEnterprise queries Consul for information about itself, it then
+// parses this information to determine whether the Consul being
+// queried is Enterprise or OSS. Returns true if Consul is Enterprise.
+func (c *ConsulClient) IsEnterprise(ctx context.Context) (bool, error) {
+	c.logger.Debug("getting sku")
```
lornasong, line 124: There's still a couple references to "sku" here and in a couple other places. Would you mind doing a find and replace?

client/consul.go
lornasong: Thanks for looking into this. Wow, that payload is pretty large. Agree with @mkam - it'll be hard for the user to dig through the payload to find the information needed to look into this problem. If there is actually an issue with parsing out the version though, the user will probably need to go query and examine the response 🤔 . I'm 50/50 and ok with skipping logging the response object

retry/retry.go
lornasong: I named it "retry" - singular "retry" fits my intuition of how I think of this. If "retries" is better for other people, I'm ok to go with that.

I personally prefer -1 to avoid the additional check for nil when maxRetry is 0 or positive. It also would avoid any accidental nil pointer runtime error. Open to reasons for having a pointer though

retry/retry.go
```diff
@@ -91,19 +91,27 @@ func (r Retry) waitTimeInNs(attempt int) int {
...
+func WaitTimeInNs(attempt int, random *rand.Rand) time.Duration {
 	a := float64(attempt)
+
+	// Check if max attempts reached, by reversing the calculation
+	// for the wait time using the maxWaitTime
+	maxBaseTimeWithDelayInS := float64(maxWaitTimeInNs) / float64(time.Second)
+	maxA := (math.Log(2 * maxBaseTimeWithDelayInS / 3)) / math.Log(2)
+	if a >= maxA {
+		return maxWaitTimeInNs
+	}
```
lornasong, line 103: Curious about this new addition to calculate the number of attempts needed for the maxBaseTimeWithDelayInNS. Was there an issue with the previous calculation method? I found the previous way of handling this easier to read

client/consul.go
wilkermichael: I think a debug log would be okay, it would be useful for the case where it's failing and the user would be interested in performing the diagnostics/providing it to us a in a bug.

retry/retry.go
```diff
@@ -11,27 +11,31 @@ import (
 	"github.com/pkg/errors"
 )
 
-const taskSystemName = "task"
+const (
+	taskSystemName  = "task"
+	maxWaitTimeInNs = 15 * time.Minute // 15 minutes
```
lornasong, line 16: The switch to time.Duration is clean! Wanted to double check if it's still necessary to have `InNs` as part of this constant name? Similarly the related method that now return time.Duration?

wilkermichael: durations are always in nansoseconds, so that makes sense to me. What do you think @jwair

retry/retry.go
wilkermichael: With the change to duration, it became apparent that the attempt number passed in overflows really fast (like 30 attempts) due to the exponents and conversion to nanoseconds. So I added this calculation to eliminate this. I'll actually also add a check for the max int to make sure we don't ever overflow

client/consul.go
```diff
@@ -101,59 +103,68 @@ func (c *ConsulClient) GetLicense(ctx context.Context, q *consulapi.QueryOptions
...
 }
 
-func parseSKU(info Self) (string, bool) {
+func isConsulEnterprise(ctx context.Context, info ConsulAgentConfig) (bool, error) {
+	logger := logging.FromContext(ctx)
 	v, ok := info["Config"]["Version"].(string)
 	if !ok {
-		return "", ok
+		logger.Debug("expected keys map[Config][Version] to exist", "ConsulAgentConfig", info)
+		return false, errors.New("unable to parse map[Config][Version], keys do not exist")
```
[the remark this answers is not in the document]
jwair: This is similar to my question that Michael answered here https://github.com/hashicorp/consul-terraform-sync/pull/783/files#r842193597

retry/retry.go
jwair: Sounds good, thank you both for reflecting on this.

retry/retry.go
wilkermichael: Introduced this since I noticed the attempts was overflowing at really low integers (like 30) due to the exponents and converting to nanoseconds. This makes it so we can accept the full range of integers.

retry/retry.go
lornasong: Agree that time.Duration is always in nanoseconds. For me that feels like an implementation detail of the time.Duration package rather than CTS. That is, when something is a time.Duration type (vs. a float/int), its no longer necessary for CTS to clarify that its in nanoseconds.

With regards to this constant name (vs. method names), I think the nanosecond naming could be confusing with the minutes unit on the right.

wilkermichael: yeah, I can get rid of the nanoseconds since the time package is handling all the conversions. I think it makes sense not to include it.

client/consul.go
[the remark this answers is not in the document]
wilkermichael: I can explicitly check each scenario to provide a more meaningful error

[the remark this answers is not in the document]
lornasong: I should have worded my question clearer. My question isn't around chaining checks. My question is regarding the log/error message. map[Config][Version].(string) can fail due to the final string cast. In this case, the keys do exist but the log/error message says the keys don't exist. I want to confirm this and see if it makes sense to update the log/error message.

client/consul_test.go
```diff
@@ -1,57 +1,91 @@
...
+			expectError:      true,
+		},
+		{
+			name: "bad key",
+			info: ConsulAgentConfig{
+				"NoConfig": {"Version": "***"},
+			},
+			expectEnterprise: false,
+			expectError:      true,
+		},
```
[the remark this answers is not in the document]
wilkermichael: done, and also check for an explicit error message for this case

retry/retry_test.go
```diff
@@ -170,13 +170,19 @@ func TestWaitTime(t *testing.T) {
 			6 * time.Second,
 		},
 		{
-			"max attempt minimum",
+			"maximum attempt before max wait time",
```
lornasong, line 173: +1  to this test case

lornasong (approved):
thanks for taking the time to have all of these conversations. excited for the new consul client!

PR #788, opened 2022-04-05
Add Terraform Cloud workspace configuration to API

mkam:
Adds the enterprise-only `terraform_cloud_workspace` attribute to as an option when creating a task via the API. Also updated `terraform_version` to not be returned by the task API it's an empty string, which would always be the case when not using OSS.


api/openapi.yaml
```diff
@@ -446,6 +449,23 @@ components:
       required:
         - path
 
+    TerraformCloudWorkspace:
+      type: object
+      additionalProperties: false
+      description:  Enterprise only. Configuration values to use for the Terraform Cloud workspace associated with the task. This is only available when used with the Terraform Cloud driver.
```
jwair, line 455:
```suggestion
      description: Enterprise only. Configuration values to use for the Terraform Cloud workspace associated with the task. This is only available when used with the Terraform Cloud driver.
```

jwair: Just an extra space 
image

api/representation.go
```diff
@@ -374,7 +381,16 @@ func oapigenTaskFromConfigTask(tc config.TaskConfig) oapigen.Task {
 	}
 
 	// Enterprise
-	task.TerraformVersion = tc.TFVersion
+	if tc.TFVersion != nil && *tc.TFVersion != "" {
+		task.TerraformVersion = tc.TFVersion
+	}
+	if tc.TFCWorkspace != nil && !tc.TFCWorkspace.IsEmpty() {
```
jwair, line 387:
```suggestion

	if tc.TFCWorkspace != nil && !tc.TFCWorkspace.IsEmpty() {
```

jwair: Spacing out code blocks

api/openapi.yaml
jwair: Just an extra space

api/openapi.yaml
```diff
@@ -446,6 +449,23 @@ components:
...
+      properties:
+        execution_mode:
+          type: string
+          example: "agent"
+          description: Enterprise only. Execution mode to set for the Terraform Cloud workspace associated with the task. Can only be "remote" or "agent". Defaults to "remote".
+        agent_pool_id:
+          type: string
+          example: "apool-123"
+          description: Enterprise only. Agent pool ID to set for the Terraform Cloud workspace associated with the task when the execution mode is "agent". Either agent_pool_id or agent_pool_name is required if execution mode is "agent". If both are set, then agent_pool_id takes precedence.
+        agent_pool_name:
```
wilkermichael, line 465: can you add a description for agent_pool_name as well?

api/representation.go
```diff
@@ -195,6 +195,13 @@ func (tr TaskRequest) ToTaskConfig() (config.TaskConfig, error) {
 
 	// Enterprise
 	tc.TFVersion = tr.Task.TerraformVersion
+	if tr.Task.TerraformCloudWorkspace != nil {
```
wilkermichael, line 198: So I'm sure this is true, but just double checking

Since this is OSS, and since it gets converted to a config, we will catch the "error of an enterprise config" as part of the config validation right?

wilkermichael (approved):
Good job Melissa! Really minor comments

api/representation.go
```diff
@@ -374,7 +381,16 @@ func oapigenTaskFromConfigTask(tc config.TaskConfig) oapigen.Task {
 	}
 
 	// Enterprise
-	task.TerraformVersion = tc.TFVersion
+	if tc.TFVersion != nil && *tc.TFVersion != "" {
```
wilkermichael, line 384: I don't think you have to dereference tc here when checking the TFVersion?

api/representation.go
mkam: Yup, validation will be handled later by the server and will look like this:
```
{
    "error": {
        "message": "unsupported configuration 'terraform_cloud_workspace' for task \"mkam_test_1\". This option is available for Consul-Terraform-Sync Enterprise when using the Terraform Cloud driver"
    },
    "request_id": "8d1e47f1-d400-ef35-326c-84ebe93bf487"
}
```

wilkermichael: That message will only be outputted if we configure it in [processUnusedConfigKeys](https://github.com/hashicorp/consul-terraform-sync/blob/main/config/decode.go#L14-L40) right?

Issue #790, opened 2022-04-08
Add Terraform version to Terraform Cloud workspace configuration


mkam:
#### Description
Prerequisite: Support for Terraform Cloud Agents  (https://github.com/hashicorp/consul-terraform-sync/issues/772)

Task configurations now have a `terraform_cloud_workspace` block used in enterprise-only with the Terraform Cloud driver. This block sets Terraform Cloud workspace configurations that are specific to a task. CTS also has `task.terraform_version`, which controls the Terraform version used by the TFC workspace.

These configurations should be consolidated where `terraform_version` is added to `task.terraform_cloud_workspace` and the existing `task.terraform_version` is deprecated.

PR #793, opened 2022-04-11
Rename driver task TF version to deprecated

mkam:
Rename the TF version of a driver task to reflect that it has
been deprecated in favor of the TF version on the TFC workspace
configuration.

Missed this field when I previously was deprecating the task Terraform version.

Part of https://github.com/hashicorp/consul-terraform-sync/issues/790


wilkermichael (approved):
Looks good to me!

PR #794, opened 2022-04-11
POC for refactoring a "tasks manager" out of controllers

lornasong:
**This PR is _not_ intended to be reviewed or merged.**

This PR is a rough refactor poc for a high-level of where the refactor will "end up". Subsequent smaller PRs will be opened for review to accomplish this approximate refactor. This PR will serve as a reference for those PRs.

Notes:
 - E2e tests passing but unit/integration tests not yet refactored
 - No clean up, comments, logging, test improvements have been done. Will be done in the reviewed prs

High-level changes:
 - New file & struct controller/task_manager.go - task crud and execution code
 - New file & struct controller/condition_monitor.go - have the code around monitoring hcat and triggering task execution
 - New file & struct driver/factory.go - create drivers, previously baseController code

PR #796, opened 2022-04-12
Fix in-memory state store config shared references

lornasong:
Previously introduced storing the configuration in the in-memory state store: https://github.com/hashicorp/consul-terraform-sync/pull/785

Problem: the stored and returned config in NewInMemoryStore() and GetConfig() share references to the child fields
Fix: update these to store/return copies of the config

Changes:
 - Commits 1-4: fix some issues with configuration Copy() methods for finalized configs
 - Commit 5: fix shared references
 - Commit 6: cleanup redundant comments

### DIFF
--- a/config/task.go
+++ b/config/task.go
@@ -104,9 +104,15 @@ func (c *TaskConfig) Copy() *TaskConfig {
 	o.Description = StringCopy(c.Description)
 	o.Name = StringCopy(c.Name)
 
-	o.Providers = append(o.Providers, c.Providers...)
+	if c.Providers != nil {
+		o.Providers = make([]string, 0)
+		o.Providers = append(o.Providers, c.Providers...)
+	}
 
-	o.DeprecatedServices = append(o.DeprecatedServices, c.DeprecatedServices...)
+	if c.DeprecatedServices != nil {
+		o.DeprecatedServices = make([]string, 0)
+		o.DeprecatedServices = append(o.DeprecatedServices, c.DeprecatedServices...)
+	}
 
 	o.Module = StringCopy(c.Module)
 	o.DeprecatedSource = StringCopy(c.DeprecatedSource)
@@ -114,7 +120,10 @@ func (c *TaskConfig) Copy() *TaskConfig {
 	o.ModuleInputs = c.ModuleInputs.Copy()
 	o.DeprecatedSourceInputs = c.DeprecatedSourceInputs.Copy()
 
-	o.VarFiles = append(o.VarFiles, c.VarFiles...)
+	if c.VarFiles != nil {
+		o.VarFiles = make([]string, 0)
+		o.VarFiles = append(o.VarFiles, c.VarFiles...)
+	}
 
 	if c.Variables != nil {
 		o.Variables = make(map[string]string)
--- a/config/task.go
+++ b/config/task.go
@@ -105,12 +105,12 @@ func (c *TaskConfig) Copy() *TaskConfig {
 	o.Name = StringCopy(c.Name)
 
 	if c.Providers != nil {
-		o.Providers = make([]string, 0)
+		o.Providers = make([]string, 0, len(c.Providers))
 		o.Providers = append(o.Providers, c.Providers...)
 	}
 
 	if c.DeprecatedServices != nil {
-		o.DeprecatedServices = make([]string, 0)
+		o.DeprecatedServices = make([]string, 0, len(c.DeprecatedServices))
 		o.DeprecatedServices = append(o.DeprecatedServices, c.DeprecatedServices...)
 	}
 
@@ -121,7 +121,7 @@ func (c *TaskConfig) Copy() *TaskConfig {
 	o.DeprecatedSourceInputs = c.DeprecatedSourceInputs.Copy()
 
 	if c.VarFiles != nil {
-		o.VarFiles = make([]string, 0)
+		o.VarFiles = make([]string, 0, len(c.VarFiles))
 		o.VarFiles = append(o.VarFiles, c.VarFiles...)
 	}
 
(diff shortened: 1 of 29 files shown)


state/in_memory_store.go
```diff
@@ -20,7 +20,7 @@ type InMemoryStore struct {
 // configStorage is the storage for the configuration with its own mutex lock
 type configStorage struct {
 	mu   sync.RWMutex
-	conf config.Config
+	conf *config.Config
```
wilkermichael, line 23: I'm a bit confused on your point 

`Store config.Config as a pointer - this allows modifications to the config
 down the line`

Can you give me a scenario? If you don't store the config as a pointer, but the method altering it takes a pointer receiver (i.e. `s *InMemoryStory), won't it still alter the underlying config?

state/in_memory_store_test.go
```diff
@@ -48,31 +48,65 @@ func Test_NewInMemoryStore(t *testing.T) {
...
+
+		// Confrm that input and stored config reference different objects
+		assert.NotSame(t, finalizedConf, actual.conf.conf)
+
+		// Confrm that input and stored config fields reference different objects
+		assert.NotSame(t, finalizedConf.Tasks, actual.conf.conf.Tasks)
+	})
 }
 
 func Test_InMemoryStore_GetConfig(t *testing.T) {
```
wilkermichael, line 68: would it make sense to combine the tests from the table with the t.Run() test?

It seems like the second test (the t.Run() test) tests the first test scenario and more

wilkermichael: same with the above test? I don't think the extra checks add a lot of overhead

wilkermichael (changes requested):
@lornasong glad you caught these cases, great work! Just some minor comments.

config/task.go
```diff
@@ -104,17 +104,26 @@ func (c *TaskConfig) Copy() *TaskConfig {
 	o.Description = StringCopy(c.Description)
 	o.Name = StringCopy(c.Name)
 
-	o.Providers = append(o.Providers, c.Providers...)
+	if c.Providers != nil {
+		o.Providers = make([]string, 0)
```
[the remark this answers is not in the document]
lornasong: Sure, good idea

state/in_memory_store.go
lornasong: Yes, you're right. Thank you for catching this

state/in_memory_store_test.go
lornasong: If I understand you correctly, I think you're suggesting to combine the separated dereference test-case with the table test?

If so, I think it is possible. As you mentioned, it would require some extra handling for the some test cases. I'm also planning to add more read methods which will also require extra handling:
 - [example GetTask test](https://github.com/hashicorp/consul-terraform-sync/pull/794/files#diff-646d14a48a3be7e665abc90317bb268d1064583c15147b61fd7c4236c2f96754R263-R273) would require an extra if-statement
 - [example of GetAllTask test](https://github.com/hashicorp/consul-terraform-sync/pull/794/files#diff-646d14a48a3be7e665abc90317bb268d1064583c15147b61fd7c4236c2f96754R141-R145) also requires an extra if statement
 - [example of GetConfig test](https://github.com/hashicorp/consul-terraform-sync/pull/794/files#diff-646d14a48a3be7e665abc90317bb268d1064583c15147b61fd7c4236c2f96754L70-R100) doesn't require an extra handling

I ended up making it outside the table-test so that it would be consistent across tests and easier to identify.

lornasong:
Thanks for reviewing @wilkermichael -this is ready for re-review when you have a chance

state/in_memory_store_test.go
wilkermichael: ok, sure, with the extensions it makes sense to leave it separate, I'm okay with this :)

state/in_memory_store.go
```diff
@@ -36,26 +36,21 @@ func NewInMemoryStore(conf *config.Config) *InMemoryStore {
 	}
 }
 
-// GetConfig returns the config
 func (s *InMemoryStore) GetConfig() config.Config {
```
wilkermichael, line 39: relooking at this again, why did we get rid of the comments for these public methods? Shouldn't we always include a comment for a public method? Even if it is part of the interface.

state/in_memory_store.go
```diff
@@ -20,7 +20,7 @@ type InMemoryStore struct {
 // configStorage is the storage for the configuration with its own mutex lock
 type configStorage struct {
 	mu   sync.RWMutex
-	conf *config.Config
+	conf config.Config
```
wilkermichael, line 23: since this defines a configStorage object, which is effectively a config + mutex, would it make sense to embed the config instead? That way you won't need `conf.conf` calls.

wilkermichael (approved):
Thanks for making those changes Lorna!

One question, for the Get call now, do we still need to make a copy of the config? I just want to double check, we might need to because the config itself contains pointers to values.

state/in_memory_store.go
lornasong: Yes, I think you're right. My linter stopped warning me about this for some reason. The comments are a bit redundant but I think it's still the Go standard to have them for public methods. Thank you, adding them back

state/in_memory_store.go
lornasong: I always forget about embedding! Great idea, cleans up the `conf.conf` and makes readability a lot better

lornasong:
Sorry, missed your question earlier:

> One question, for the Get call now, do we still need to make a copy of the config? I just want to double check, we might need to because the config itself contains pointers to values.

Correct, a copy is needed since the config struct's fields contains pointers. If we don't make a copy, the below assert at line 107 will fail:

```
// Confirm returned config field reference different object from stored
assert.NotSame(t, storedConf.Port, actual.Port)
```
```
 Error Trace:	in_memory_store_test.go:107
 Error:      	Expected and actual point to the same object: 0xc0002223d8 (*int)(0xc0002223d8)
 Test:       	Test_InMemoryStore_GetConfig/returned_config_is_dereferenced
```

Issue #798, opened 2022-04-13
Task Status API response's provider field does not match configuration when alias used


lornasong:
### Describe the bug
The Task API payload returns the provider information in two places (see below). The provider that is returned is the provider name rather than the provider id. The provider id is what a user would have configured the task with.

Impact:
 - This only impacts users who have configured [multiple of the same providers](https://www.consul.io/docs/nia/configuration#multiple-provider-configurations) using aliases.
 - Task Status API response provider field is deprecated [#569](https://github.com/hashicorp/consul-terraform-sync/issues/569)

The provider id format is:
 - If alias is not provided: `<provider-name>` e.g. myfirewall
 - If alias is provided: `<provider-name>.<provider-alias> `e.g. myfirewall.east

The provider returned in the Task API payload should have the provider id to include alias information for accuracy

```
{
  "task_b": {
    "task_name": "task_b",
    "status": "errored",
    "enabled": false,
    "providers": [
      "null"  // here ***
    ],
    "events_url": "/v1/status/tasks/task_b?include=events",
    "events": [
      {
        "id": "44137ba2-8fc9-6cbe-0e0e-e9305ee4f7f9",
        "success": false,
        "config": {
          "providers": [
            "null" // here ***
          ],
  // ...
```
### Versions
**Consul Terraform Sync**
```
v0.1.0-techpreview2
(since Task Status API was introduced)
```

### Configuration File(s)

<details>
<summary>Click to toggle contents of config file </summary>

```hcl
terraform_provider "local" {
	alias = "east"
}

task {
 name = "ls_testing_services"
 version = "0.0.1"
 module = "lornasong/cts_file/local"
 condition "services" {
   names = ["api"]
 }
 providers = ["local.east"]
}
```
</details>

### Expected Behavior

I expect the Task Status API response provider field to match the providers information that was I configured for the task
```
curl localhost:8558/v1/status/tasks/ls_testing_services | jq .
{
  "ls_testing_services": {
    "task_name": "ls_testing_services",
    "status": "successful",
    "enabled": true,
    "events_url": "/v1/status/tasks/ls_testing_services?include=events",
    "providers": [
      "local.east" // matches
    ],
    "services": []
  }
}
```

### Actual Behavior

I expect the Task Status API response provider field to match the providers information that was I configured for the task i.e. `local.east`

```
curl localhost:8558/v1/status/tasks/ls_testing_services | jq .
{
  "ls_testing_services": {
[...]

PR #799, opened 2022-04-13
Support managing task information in state store

lornasong:
Storing configuration was added in https://github.com/hashicorp/consul-terraform-sync/pull/785. This PR supports accessing  and updating task configuration in the state store.

Also fixes https://github.com/hashicorp/consul-terraform-sync/issues/798 that was uncovered from converting to using state store

Overview:
 - Commit 1-2: Add new methods to state store
 - Commit 3-6: Update controller to use state store, particularly controller server
 - Commit 7: Fix an issue with providers that was uncovered by updating the controller server Task() to using the state stored configuration rather than using the driver (see [circleci broken test](https://app.circleci.com/pipelines/github/hashicorp/consul-terraform-sync/3835/workflows/d82e951b-043c-4c31-923b-b96191766a71/jobs/10873))

Sincere apologies for such a large PR. I tried splitting out access (Get methods) vs. updating (Set and Delete methods) task configuration but they have a dependency on each other that caused e2e tests to break. Please let me know if it would be helpful to review in person.


controller/server.go
```diff
@@ -20,18 +20,26 @@ func (rw *ReadWrite) Events(ctx context.Context, taskName string) (map[string][]
...
+		return conf, nil
 	}
 
-	conf, err := configFromDriverTask(d.Task())
-	if err != nil {
-		return config.TaskConfig{}, err
+	return config.TaskConfig{}, fmt.Errorf("a task with name '%s' does not exist or has not been initialized yet", taskName)
+}
+
+func (rw *ReadWrite) Tasks(ctx context.Context) ([]config.TaskConfig, error) {
```
lornasong, line 32: Note: Tasks() returning an error is no longer necessary at this point. I have a task to revisit this at the end of refactoring to remove returning an error. I decided not to include it in this PR since its already so large

state/in_memory_store_test.go
```diff
@@ -106,7 +106,240 @@ func Test_InMemoryStore_GetConfig(t *testing.T) {
...
+				{Name: config.String("existing_task")},
+			},
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			conf := &config.Config{
+				Tasks: &config.TaskConfigs{
+					{Name: config.String("existing_task")},
```
mkam, line 334: Could you add the case of removing a task when there are multiple existing tasks?

controller/server.go
```diff
@@ -29,6 +29,19 @@ func (rw *ReadWrite) Task(ctx context.Context, taskName string) (config.TaskConf
 	return config.TaskConfig{}, fmt.Errorf("a task with name '%s' does not exist or has not been initialized yet", taskName)
 }
 
+func (rw *ReadWrite) Tasks(ctx context.Context) ([]config.TaskConfig, error) {
+	// TODO handle ctx while waiting for state lock if it is currently active
+	tasks := rw.state.GetAllTasks()
+
+	// convert config.TaskConfigs => []config.TaskConfig
```
mkam, line 36: Would it make sense in the future to have `Tasks()` return `config.TaskConfigs`? I haven't looked into where this all is called, so not sure the impact of this potential change.

mkam (approved):
Changes look good to me!

state/in_memory_store_test.go
lornasong: Sure, added

controller/server.go
lornasong: Thanks for thinking of this! I gave it a try and it trickles down some minor changes mostly in the api package. Since it also involves updating the server interface and regenerating the mock, I'd like to make this change at the same time as updating Tasks() to no longer return an error ([earlier comment](https://github.com/hashicorp/consul-terraform-sync/pull/799#discussion_r849699691)) if that works for you?

lornasong: Thanks for thinking of this! I tried it out and it was mostly minor trickle down changes in the `api` package.

I'd like to make this change together later when I update Tasks() to remove returning an error ([earlier comment](https://github.com/hashicorp/consul-terraform-sync/pull/799#discussion_r849699691)) since they similarly update the return signature of `Tasks()` if that works for you?

mkam: That sounds good to me!

lornasong: Following up: this is addressed here https://github.com/hashicorp/consul-terraform-sync/pull/846/commits/2931a2a135d1471bd16a1735af21d50812aae9a9

controller/server.go
lornasong: Following up, this is addressed here https://github.com/hashicorp/consul-terraform-sync/pull/846/commits/2931a2a135d1471bd16a1735af21d50812aae9a9

PR #805, opened 2022-04-15
Simplified contexts used in Consul client unit tests

jwair:


lornasong (approved):
I like the improvement to directly call context.Background() in the method

PR #806, opened 2022-04-18
Health API Endpoint OLD

jwair:


api/handler.go
```diff
@@ -19,10 +19,11 @@ const (
 // the handler to adhere to the generated server interface
 type Handlers struct {
 	*TaskLifeCycleHandler
+	*HealthHandler
 }
 
-//go:generate oapi-codegen  -package oapigen -generate types -o oapigen/types.go openapi.yaml
```
jwair, line 24: extra spaces

api/task_create.go
```diff
@@ -70,8 +73,7 @@ func (h *TaskLifeCycleHandler) CreateTask(w http.ResponseWriter, r *http.Request
 	logger.Trace("task created", "create_task_response", resp)
 }
 
-func (h *TaskLifeCycleHandler) createDryRunTask(w http.ResponseWriter, r *http.Request,
```
jwair, line 73: single line for better readability

api/task_delete.go
```diff
@@ -15,7 +15,7 @@ func (h *TaskLifeCycleHandler) DeleteTaskByName(w http.ResponseWriter, r *http.R
 
 	ctx := r.Context()
 	requestID := requestIDFromContext(ctx)
-	logger := logging.FromContext(r.Context()).Named(deleteTaskSubsystemName).With("task_name", name)
+	logger := logging.FromContext(ctx).Named(deleteTaskSubsystemName).With("task_name", name)
```
jwair, line 18: inlining ctx

client/consul.go
```diff
@@ -21,7 +21,7 @@ const (
 	consulSubsystemName       = "consul"
 )
 
-//go:generate mockery --name=ConsulClientInterface --filename=consul_client.go --output=../mocks/client --tags=enterprise
```
jwair, line 24: I thought I fixed this already but I guess I didn't, this is just to fix the filename for the consul client mock

controller/readonly.go
```diff
@@ -42,16 +42,16 @@ func NewReadOnly(conf *config.Config) (*ReadOnly, error) {
 		return nil, err
 	}
 
-	state := state.NewInMemoryStore(conf)
```
jwair, line 45: variable name conflicts with imported package name

controller/readwrite.go
```diff
@@ -37,16 +38,16 @@ func NewReadWrite(conf *config.Config) (*ReadWrite, error) {
 		return nil, err
 	}
 
-	state := state.NewInMemoryStore(conf)
```
jwair, line 40: variable name conflicts with imported package name

controller/tasks_manager.go
```diff
@@ -75,11 +75,11 @@ func (tm *TasksManager) Config() config.Config {
 	return tm.state.GetConfig()
 }
 
-func (tm *TasksManager) Events(ctx context.Context, taskName string) (map[string][]event.Event, error) {
+func (tm *TasksManager) Events(_ context.Context, taskName string) (map[string][]event.Event, error) {
```
jwair, line 78: fixing warnings, those params are not used

controller/tasks_manager_test.go
```diff
@@ -672,20 +672,20 @@ func Test_TasksManager_CheckApply(t *testing.T) {
 			}
 
 			assert.Equal(t, 1, len(events))
-			event := events[0]
```
jwair, line 675: variable name conflicts with imported package name

api/task_delete.go
jwair: this change is made in other places in this PR

controller/tasks_manager.go
jwair: this change is made in other places in this PR

api/openapi.yaml
```diff
@@ -22,6 +22,25 @@ servers:
       port:
         default: '8558'
 paths:
+  /v1/health:
```
wilkermichael, line 25: I would add a tag for `health` similar to the tag for `tasks`

``` yaml
      tags:
        - health
```

api/openapi.yaml
```diff
@@ -500,3 +519,29 @@ components:
           example: "this is an error message"
       required:
         - message
+
+    Timestamp:
```
wilkermichael, line 523: Not in scope for this PR, but I think this is something we should add to all of our responses.

api/openapi.yaml
```diff
@@ -500,3 +519,29 @@ components:
...
+      type: string
+      description: Date and time (RFC 3339) representing the UTC timestamp of the health check
+      example: 2021-10-23T15:05:37Z
+
+    GoodHealthCheckResponse:
```
wilkermichael, line 528: Something we should cleanup.

Currently our task endpoints also include the error block, we just expect this to be empty when a response is successful. I think the original intent behind this was to allow for a single object to be consumed by a client. We should decide if this is the way we'd like to handle response objects going forward, or if we want two distinct objects (one for success and one for errors).

api/openapi.yaml
```diff
@@ -500,3 +519,29 @@ components:
           example: "this is an error message"
       required:
         - message
+
+    Timestamp:
+      type: string
+      description: Date and time (RFC 3339) representing the UTC timestamp of the health check
+      example: 2021-10-23T15:05:37Z
```
wilkermichael, line 526: missing quotes around example string

wilkermichael (commented):
Just took a quick look at the openapi.yaml file

api/openapi.yaml
jwair: Is it because it is a string or for consistency? because it is not required for valid YAML

api/openapi.yaml
wilkermichael: removing timestamp for 0.6.0 as discussed

wilkermichael:
TODO: Create ticket for adding timestamp to endpoints

api/openapi.yaml
wilkermichael: TODO: add ticket for adding timestamp to endpoints

api/openapi.yaml
wilkermichael: TODO: create spike ticket to re-evaluate this

api/openapi.yaml
wilkermichael: removed timestamp so no longer required for 0.6.0

api/openapi.yaml
jwair: If we add error here, there's not a lot of value in having two return types `Good...` and `Bad...`

api/openapi.yaml
jwair: Ticket created

jwair:
Closing in favor of https://github.com/hashicorp/consul-terraform-sync/pull/823

PR #808, opened 2022-04-19
Split out baseController from TasksManager pt2

lornasong:
Currently, baseController is embedded in TasksManager

This is the second PR in a series of changes to split out baseController from TasksManager. baseController logic is mostly around creating and initializing drivers for tasks. Currently the goal is:
 - Convert `controller.baseController` to something like `driver.Factory`
 - TasksManager will have a `driverFactory` field and call `driverFactory.Make()` for new tasks (naming input welcome)

Previous PR: https://github.com/hashicorp/consul-terraform-sync/pull/804

This PR focuses on detangling `drivers` from baseController and moving it to TasksManager. A major resulting change is refactoring Once-mode and Inspect-mode to use existing TasksManager methods (TaskCreateAndRun and TaskInspect) to create drivers rather than using the baseController.

Commits
 - 1: Add e2e test for inspect-mode and once-mode in anticipation of refactor (passing)
 - 2-4: Move various driver calls from baseController to TasksManager
 - 5-8: Resulting once-mode refactor + tests
 - 9-11: Resulting inspect-mode refactor + tests
 - 12-14: Fixes some e2e test issues (see failed test at commit 11 and 13)

Future work: move / rename baseController struct and methods. clean up and remaining TODOs mentioned in code

Note: this PR got large from cascading broken things 😭 . The individual commits are intended to make reviewing a little easier. Please let me know if it would be helpful to go over anything in person


mkam (approved):
Looks good to me! Had a non-blocking comment around the error returned by `WatchDep()`.

controller/readonly.go
```diff
@@ -29,11 +30,6 @@ type ReadOnly struct {
 
 // NewReadOnly configures and initializes a new ReadOnly controller
 func NewReadOnly(conf *config.Config) (*ReadOnly, error) {
-	// Run the driver with logging to output the Terraform plan to stdout
-	if tfConfig := conf.Driver.Terraform; tfConfig != nil && !MuteReadOnlyController {
-		tfConfig.Log = config.Bool(true)
```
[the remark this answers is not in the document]
lornasong: Good question! Thanks for asking. You're right, it was no longer used. For example if you go to an earlier commit (before all this refactoring) and hardcode this to false, the plan still outputs to the terminal. This is because we switched to reading the plan into a buffer and logging it in the [old checkInspect code](https://github.com/hashicorp/consul-terraform-sync/pull/808/files#diff-8e409a7329d9bad51e5c0d25844fa2d15602caab402947e2d4e8b60951e93e3bL328-L332) I removed. `MuteReadOnlyController` is so far only used by our benchmarks but I haven't checked how those are outputting.

That makes me realize I'm not outputting the plan URL like the old inspect code! I'll add a commit to make this the same

controller/readwrite.go
```diff
@@ -99,7 +99,38 @@ func (rw *ReadWrite) Run(ctx context.Context) error {
...
-	return rw.tasksManager.Once(ctx)
+	rw.logger.Info("executing all tasks once through")
+
+	// Stop watchiing dependencies after once-mode completes
+	ctxDep, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// start watching dependencies in order to render templates to apply tasks
+	// once
+	go rw.tasksManager.WatchDep(ctxDep)
```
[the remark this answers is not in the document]
lornasong: Oh yes, thanks for calling that out. That's a good point. Let me add a check for an error in ReadWrite and ReadOnly and add some tests. Let me try it out!

controller/readonly.go
```diff
@@ -60,16 +60,40 @@ func (ro *ReadOnly) Init(ctx context.Context) error {
...
+	go func() {
+		exitCh <- ro.inspectConsecutive(ctxInspect)
+		cancelWatch()
+	}()
+
+	counter := 0
+	for {
+		err := <-exitCh
+		counter++
+		if err != context.Canceled {
```
mkam, line 88: Will this be an issue if `err` is nil in the happy path case?

lornasong: If the err is nil, the counter will still increment. Once both WatchDep and inspectOnce finish, they'll both send a nil to the exitCh. Then Run() should exit at line 94 below. `Test_ReadOnly_Run` also should capture the happy test path and looks to be still passing (I did break the equivalent `Test_ReadWrite_Once_WatchDep_errors_Terraform` at one point while developing :D ). Let me know if you have concerns

mkam: This is the scenario I was thinking of:

1. inspectConsecutive finishes, returns nil on exitCh and cancels watch
2. This loop gets the nil on the exitCh, increments counter to 1, but since nil isn't equal to `context.Canceled`, it returns nil on line 90.
3. Watch is stopped because of the cancel, but the function has already returned and didn't necessarily wait for the watch to finish cancelling.

But I also need to learn more about contexts and cancelling contexts so I might be way off base here!

lornasong: Ah, I get what you're saying. Yes, I agree. You're absolutely right. To confirm, I ran Test_ReadOnly_run with coverage and saw that it exited early and did not wait for WatchDep:

Screen Shot 2022-04-20 at 1 07 57 PM

I added a commit to fix this for rw.Once and ro.Run (and added a cautionary check in rw.Run - from a previous change). The coverage now looks something like:
Screen Shot 2022-04-20 at 1 52 49 PM

Thank you for catching this! Let me know if you have any other feedback

PR #813, opened 2022-04-25
Update e2e test WaitForAPI

lornasong:
Context:
 - Currently the API is served once daemon once-mode was complete
 - A lot of tests (e2e in particular) rely on API availability to determine daemon once-mode is complete

Goal: decouple serving the API from daemon once-mode. Serve the API before daemon once-mode

This PR removes test reliance on API availability to determine daemon once-mode is complete. Instead determine once-mode completion by examining the status response rather than API availability.

Future work: move when the API is served to be before daemon once-mode

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -176,26 +176,40 @@ func (c *Client) Scheme() string {
 	return c.url.Scheme
 }
 
-// WaitForAPI polls the /v1/status endpoint to check when the CTS API is
-// available. The API is started when CTS is run as a daemon and completes
-// all tasks once.
-func (c *Client) WaitForAPI(timeout time.Duration) error {
+// WaitForOnce checks if all the current enabled tasks have been run once by
+// polling the /v1/status/tasks endpoint. Tasks that have not run will have an
+// unknown status. Should only use if no runtime changes are being actively made.
+func (c *Client) WaitForOnce(timeout time.Duration) error {
 	polling := make(chan struct{})
 	stopPolling := make(chan struct{})
 	statusAPI := c.Status()
 
 	go func() {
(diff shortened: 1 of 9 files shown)


api/client.go
```diff
@@ -176,26 +176,40 @@ func (c *Client) Scheme() string {
...
 }
 
-// WaitForAPI polls the /v1/status endpoint to check when the CTS API is
-// available. The API is started when CTS is run as a daemon and completes
-// all tasks once.
-func (c *Client) WaitForAPI(timeout time.Duration) error {
+// WaitForOnce checks if all the current enabled tasks have been run once by
+// polling the /v1/status/tasks endpoint. Tasks that have not run will have an
+// unknown status. Should only use if no runtime changes are being actively made.
+func (c *Client) WaitForOnce(timeout time.Duration) error {
```
wilkermichael, line 182: Rather than `WaitForOnce` or `WaitForAPI` maybe we can make the name more about what we're waiting to be able to do, rather than waiting for? That way in the future if this were to change again, we wouldn't have to rename the function.

I can't think of a good name at the moment, but something that conveys "waiting to be able to process tasks" or "waiting to start the nominal workflow"

wilkermichael (approved):
@lornasong looks great, one minor comment around the name `WaitForOnce` but i'm also okay with keeping it.

api/client.go
lornasong: Notes from chatting offline: went with "WaitForTestReadiness"

Other brainstormed ideas:

- WaitForReadiness
- WaitForTestReadiness
- WaitForTesting
- WaitToTest

Issue #812, opened 2022-04-22
syslog configuration doesn't work with provided docker_image


mcousens:
### Describe the bug
Enabling the advertised `syslog` integration in the global config options https://www.consul.io/docs/nia/configuration#global-config-options while running CTS via the provided docker image results in the below error:
```
[ERROR] cli: error setting up logging: error="error setting up syslog logger: Unix syslog delivery error"
```

Of note, this error only occurs when explicitly enabling the integration (when you don't pass `enabled = true` the daemon will start but the syslog integration fails silently and doesn't do anything - despite "Specifying other option also enables syslog logging")

```
syslog {
  enabled = true
  facility = "local7"
}
```

### Versions
**Consul Terraform Sync**
```
consul-terraform-sync v0.5.2
From image: hashicorp/consul-terraform-sync-enterprise:0.5.2-ent
```

**Consul Version**
```
Consul 1.8.4
```

**Terraform Version**
```
Terraform v0.15.0
```

**Other Details**
- docker image: `hashicorp/consul-terraform-sync-enterprise:0.5.2-ent`
- driver: terraform-cloud
- orchestrating daemon via Nomad
- using Terraform Enterprise

### Configuration File(s)

<details>
<summary>Click to toggle contents of config file </summary>

```hcl
log_level = "DEBUG"

buffer_period {
  enabled = true
  min     = "5s"
}

syslog {
  enabled = true
  facility = "local7"
}

# Consul connection setup
consul {
  address = "<redacted>"
  tls {
    enabled = true
    verify  = false
  }
}

# CTS task: monitors Consul service catalog and apply TF code
# in specified module whenever service layout changes
task {
  name           = "core_alb_core_backend-qa"
  description    = "Dynamically manage target groups in qa via Consul-Terraform-Sync (CTS)"
  enabled        = true
  providers      = ["aws"]
  module         = "<redacted>/core-alb-backend/cts"
  # if you don't pin module version, latest will be used
  # version        = "1.0.1"
  variable_files = ["/local/core_alb_core_backend_qa.tfvars"]

  condition "services" {
    names = ["<redacted>"]
    filter     = "Service.Tags contains \"<redacted>\""
  }
}


# TFE connection setup
driver "terraform-cloud" {
  hostname         = "<redacted>"
  organization     = "<redacted>"
  token            = "<redacted>"
  workspaces {
    prefix = "cts-"
    tags   = ["source:cts"]
  }

  required_providers {
    aws = {
[...]

PR #815, opened 2022-04-26
Fix syslog enabled value

lornasong:
Context: syslog should be automatically enabled if other fields (e.g. facility
or name) values are provided. This is documented in CTS

Issue: syslog enabled field has to be specifically set to true. Otherwise it is
false even though other config fields are provided.

This was caused by the default was set to enabled=false. When CTS builds the
config, the user-provided config is merged with the default config. If the user
left enabled unconfigured, it would merge with the default enabled=false. This
results in enabled being set to false in a way that looks as though the user set
enabled to false.

Related: https://github.com/hashicorp/consul-terraform-sync/issues/812

### DIFF
--- a/config/syslog.go
+++ b/config/syslog.go
@@ -26,7 +26,8 @@ type SyslogConfig struct {
 // DefaultSyslogConfig returns the default configuration struct.
 func DefaultSyslogConfig() *SyslogConfig {
 	return &SyslogConfig{
-		Enabled: Bool(false),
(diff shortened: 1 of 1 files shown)


wilkermichael (approved):

wilkermichael:
@lornasong looks good, good catch

lornasong:
no changes, just rebase latest on main

PR #816, opened 2022-04-26
Fix deprecated services warning

mkam:
Warning should only be logged if there are a non-zero number of services in the
deprecated field.

Noticed that this recently started logging even though I don't have the deprecated field set.

Checked with older versions, and previously when Finalize was called, DeprecatedServices was nil when no `services` configuration was set. At some point, it changed to being an empty but initialized list of strings. This new `else if` is the more appropriate behavior either way, so didn't really look into what made this change.


lornasong (approved):
thanks for fixing this!

PR #817, opened 2022-04-27
Refactor controllers for each "mode"

lornasong:
This PR does two things
 - Refactors ReadWrite and ReadOnly controllers into 3 controllers Daemon, Once, and Inspect controllers*
 - This refactor is necessary in order to start serve the API before daemon-once-mode in the Daemon controller

* Note: Our team has initially discussed naming of these modes particularly the usage of the word "mode" and "daemon". "Daemon" is used in our docs and code but "long-running" mode is also used. The naming in this PR is a placeholder until we figure out the best naming to move ahead with

Changes:
 - Commit 1-4: refactor out Inspect and Once controller
 - Commit 5: Updates ReadWrite controller to add back once-mode after serving the API
 - Commit 6-8: refactor out Daemon controller and update start command
 - Commit 9-10: fixes to e2e tests


lornasong:
@mkam fyi this PR refactors the controllers and moves the serve API. I think these can potentially impact development on self-registration. I should have gotten to these earlier! Apologies for the delay 🙏

controller/daemon.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+		if counter >= exitBufLen {
+			// Wait for all contexts to cancel
+			return ctx.Err()
+		}
+	}
+}
+
+// Once runs the tasks once. Intended to only be called by Run() or outside of
+// Run() for the case of benchmarks
```
mkam, line 108: Thanks for calling this out!

command/start.go
```diff
@@ -281,24 +274,6 @@ func (c *startCommand) Run(args []string) int {
 			return
 		}
 
-		switch ctrl := ctrl.(type) {
```
mkam, line 284: This block of code has definitely confused me in the past, so kind of nice to see it removed and the logic simplified! 😄

mkam (approved):
Looks good to me! ✨

PR #818, opened 2022-04-27
Endpoint error handling for valid error scenarios and improved retries

wilkermichael:
## Summary
This PR serves to improve handling of potentially valid error scenarios, and to improve retries by not performing retries in certain scenarios by

- Removing dependency on /agent/self for version check. This was found to be a workaround, and may not be the most dependable way to check if CTS is communicating with Consul Enterprise or OSS
- Add logic to retries to only retry on 5xx and 429 (too many requests) errors
- Rely on 404’s returned from Consul api to determine if connected to Enterprise or Consul

## Solution
To accomplish the above, I have added two new Error types:

1. `NonRetryableError` which can be used to signal to the retry logic to halt retries, for example in the case where the error returned was a 4xx (except 429) error
2. `NonEnterpriseConsulError` which is returned by the Consul Client to indicate to the caller that the error was a result of not connecting to a Consul Enterprise Server

The only API call currently implemented `GetLicense` Consul Client will process response codes and wrap errors with `NonRetryableError` to indicate to the retry code to stop retries. If a response code is a `404`, this will be interpreted as a result of connecting to an OSS Consul server, where the `license` endpoint would not exist. In this case, wrap the error in the `NonEnterpriseConsulError` and leave it to the caller to handle it.


retry/retry_test.go
```diff
@@ -142,6 +142,53 @@ func TestWithRetry_client(t *testing.T) {
...
+
+	cases := []struct {
+		name                string
+		maxRetry            int
+		nonRetryAbleErrorOn int
+		expected            error
+	}{
+		{
+			"non retryable error on first attempt",
+			2, // max retries is 2, but will succeed on first retry
```
wilkermichael, line 156: I'll fix this comment

retry/retry_test.go
```diff
@@ -142,6 +142,53 @@ func TestWithRetry_client(t *testing.T) {
...
+			0,
+			&NonRetryableError{
+				Err: errors.New("error on 0"),
+			},
+		},
+		{
+			"non retryable error after enters retry loop",
+			2, // max retries is 2, but will succeed on first retry
```
wilkermichael, line 164: I'll fix this comment

client/consul.go
```diff
@@ -39,8 +61,8 @@ type ConsulClient struct {
 	logger logging.Logger
 }
 
-// ConsulAgentConfig represents the response body from Consul /v1/agent/self API endpoint.
-// The response contains configuration and member information of the requested agent.
+// ConsulAgentConfig represents the responseCode body from Consul /v1/agent/self API endpoint.
+// The responseCode contains configuration and member information of the requested agent.
```
wilkermichael, line 65: I'll fix the responseCode, should be response

client/consul.go
```diff
@@ -100,15 +124,73 @@ func (c *ConsulClient) GetLicense(ctx context.Context, q *consulapi.QueryOptions
...
+		if errors.As(err, &nonEnterpriseConsulError) {
+			c.logger.Warn("Unable to get license, this is most likely caused by CTS connecting to OSS Consul")
+		}
+	}
 
 	return license, err
 }
+
+func getStatusCodeFromError(ctx context.Context, err error) int {
+	// Extract the unexpected responseCode substring
```
wilkermichael, line 163: fix comment to be `response code`

client/consul.go
```diff
@@ -100,15 +124,73 @@ func (c *ConsulClient) GetLicense(ctx context.Context, q *consulapi.QueryOptions
...
+	re := regexp.MustCompile("Unexpected response code: ([0-9]{3})")
+	s := re.FindString(err.Error())
+	if s == "" {
+		return 0
+	}
+
+	// Extract the status code substring from the unexpected responseCode substring
```
wilkermichael, line 170: fix comment to be `response code`

client/consul_test.go
```diff
@@ -15,15 +16,53 @@ import (
...
 	t.Parallel()
-	path := "/v1/operator/license"
 
-	intercepts := []*testutils.HttpIntercept{
-		{Path: path, ResponseStatusCode: http.StatusInternalServerError},
+	var nonRetryableError *retry.NonRetryableError
+	var nonEnterpriseConsulError *NonEnterpriseConsulError
+	cases := []struct {
+		name                 string
+		responseCode         int
```
wilkermichael, line 24: change to `statusCode` to be consistent

client/consul.go
```diff
@@ -100,15 +124,73 @@ func (c *ConsulClient) GetLicense(ctx context.Context, q *consulapi.QueryOptions
 	var license string
 
 	f := func(context.Context) error {
+		var err error
 		license, err = c.Operator().LicenseGetSigned(q)
+
+		// Process the error by wrapping it in the correct error types
```
wilkermichael, line 130: I thought about separating this processing logic it's own function, but I think the interpretation of `404` in this case might not be universal. For now just leaving the logic in place, we can re-evaluate as we add more supported API calls

lornasong (commented):
Really like the simplification to use one Consul API request! the commit organization is super readable ✨ . I saw your comments on a couple to-dos; I can review again once they're done

retry/retry.go
```diff
@@ -64,14 +84,19 @@ func (r Retry) Do(ctx context.Context, f func(context.Context) error, desc strin
 				return nil
 			}
 
-			err = fmt.Errorf("retry attempt #%d failed '%s'", attempt, err)
+			err = fmt.Errorf("retry attempt #%d failed '%w'", attempt, err)
```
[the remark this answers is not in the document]
wilkermichael: yeah, the unwrap stuff introduced in 1.13 is pretty interesting https://go.dev/blog/go1.13-errors

client/consul_test.go
wilkermichael: going with responseCode to be consistent with Consul

client/consul.go
wilkermichael: going with response code to be consistent with Consul

client/consul.go
wilkermichael: going with response code to be consistent with Consul

client/consul.go
```diff
@@ -100,15 +124,73 @@ func (c *ConsulClient) GetLicense(ctx context.Context, q *consulapi.QueryOptions
...
+			if statusCode == http.StatusNotFound {
+				err = &NonEnterpriseConsulError{Err: err}
+			}
+
+			// non-retryable errors allows for termination of retries
+			if isStatusCodeRetryable(statusCode) {
+				return err
+			} else {
+				return &retry.NonRetryableError{Err: err}
+			}
```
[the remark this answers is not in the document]
wilkermichael: To be complete the parity, what about this?

```Go
			// If we get a StatusNotFound assume that this is because CTS
			// is connected to OSS Consul where this endpoint isn't available
			// wrap in the appropriate error
			if statusCode == http.StatusNotFound {
				err = &NonEnterpriseConsulError{Err: err}
			}

			// non-retryable errors allows for termination of retries
			if !isResponseCodeRetryable(statusCode) {
				err = &retry.NonRetryableError{Err: err}
			}

			return err
```

client/consul.go
```diff
@@ -100,15 +124,73 @@ func (c *ConsulClient) GetLicense(ctx context.Context, q *consulapi.QueryOptions
...
+			c.logger.Warn("Unable to get license, this is most likely caused by CTS connecting to OSS Consul")
+		}
+	}
 
 	return license, err
 }
+
+func getStatusCodeFromError(ctx context.Context, err error) int {
+	// Extract the unexpected responseCode substring
+	re := regexp.MustCompile("Unexpected response code: ([0-9]{3})")
```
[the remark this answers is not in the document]
wilkermichael: For the MustCompile conversation, I was thinking it should be used since i'm providing the string literal statically. Therefore, if it failed, that would indicate a programmer error, so should panic?

client/consul.go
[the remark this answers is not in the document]
lornasong: Yes exactly ✨

client/consul.go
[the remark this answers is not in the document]
wilkermichael: Talked about with Lorna, will make this a global

wilkermichael:
@lornasong ready for another review

lornasong (approved):
looks good to me!

PR #819, opened 2022-04-28
Add sleep for long running WaitForTestReadiness

lornasong:
When WaitForTestReadiness is used by e2e tests, the wait can be long. There was
a change to serve the API before once-mode. This caused WaitForTestReadiness
to use more resources, which in turn caused Consul start up to frequently
timeout in e2e tests.

Add a 500ms sleep when the timeout is expected to be 1s or more.


mkam (approved):
Glad you figured out what the issue was!

PR #821, opened 2022-04-29
added tests for the error types

wilkermichael:
Added unit tests for the new error types, want to be consistent and test all public methods exported by packages

### DIFF
--- a/retry/retry_test.go
+++ b/retry/retry_test.go
@@ -246,3 +246,30 @@ func TestWaitTime(t *testing.T) {
 		})
 	}
 }
+
+func TestNonRetryableError_Error(t *testing.T) {
+	err := NonRetryableError{Err: errors.New("some error")}
+	var nonRetryableError *NonRetryableError
+
+	assert.True(t, errors.As(&err, &nonRetryableError))
+	assert.Equal(t, "this error is not retryable: some error", err.Error())
+}
+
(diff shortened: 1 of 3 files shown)


retry/retry_test.go
```diff
@@ -246,3 +246,30 @@ func TestWaitTime(t *testing.T) {
 		})
 	}
 }
+
+func TestNonEnterpriseConsulError_Error(t *testing.T) {
```
mkam, line 250: Should this be `TestNonRetryableError_Error`?

retry/retry_test.go
```diff
@@ -246,3 +246,30 @@ func TestWaitTime(t *testing.T) {
...
+	err := NonRetryableError{Err: errors.New("some error")}
+	var nonRetryableError *NonRetryableError
+
+	assert.True(t, errors.As(&err, &nonRetryableError))
+	assert.Equal(t, "this error is not retryable: some error", err.Error())
+}
+
+func TestNonEnterpriseConsulError_Unwrap(t *testing.T) {
```
mkam, line 258: Same as above, I think this was a copy/paste miss!

mkam (approved):

retry/retry_test.go
wilkermichael: Thanks for the catch!

PR #823, opened 2022-04-29
Health Endpoint

wilkermichael:
## Summary
This PR serves to introduce a new health endpoint to CTS. The health endpoint uses the concept of a `health.manager` interface to determine the health of the system. Currently the only health manager implemented has a very simple check that always just returns `error == nil`, so as long as the system can be reached it will respond `healthily`.

### Considerations

- This PR is based off of https://github.com/hashicorp/consul-terraform-sync/pull/806 which has been left to preserve some of the code around timestamps.
- Logging has been added for the health endpoint, but will leave it up to future discussion to determine if it should be removed
- An new error type was introduced to separate internal errors from those errors diagnosing the system as "unhealthy." Another possible approach would be to have the Checker return an `isHealthy` boolean status. Up for discussion on this one, but the error might allow future checks and subchecks to also utilize this error type when returning.

## Not in Scope
- E2E tests will be covered in a separate PR
- Logging will be revisited in a separate PR


wilkermichael:
@mkam I decided to create a new branch and open up a new PR to preserve some of @jwair code around timestamps separately

api/openapi.yaml
```diff
@@ -500,3 +522,19 @@ components:
           example: "this is an error message"
       required:
         - message
+
+    GoodHealthCheckResponse:
+      type: object
+      additionalProperties: false
+      properties:
+        error:
```
wilkermichael, line 530: I added error to the good health check response to be consistent with our response other objects, I will create a task to revisit these objects

health/manager.go
```diff
@@ -0,0 +1,17 @@
+package health
+
+//go:generate mockery --name=Manager --filename=manager.go --output=../mocks/health
+
+type Manager interface {
```
wilkermichael, line 5: I also thought about calling this a checker, maybe not everything needs to necessarily be a manager?

I also thought about defining the interface in the api package, to remove the api packages dependency on this package, but we don't really do that elsewhere in our code. Might be something to look at in the future if we should define the interfaces where they're used to mitigate the number of dependencies in certain packages?

api/health.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+}
+
+// GetHealth returns the health status
+func (hh *HealthHandler) GetHealth(w http.ResponseWriter, r *http.Request) {
+	hh.mu.RLock()
+	defer hh.mu.RUnlock()
+
+	logger := logging.FromContext(r.Context()).Named(getHealthSubsystemName)
+	logger.Trace("get health")
```
wilkermichael, line 36: revisit logging in separate PR

wilkermichael:
@jwair wanted to tag you as well since you opened the original PR. I didn't change very much, mostly some cleanup stuff and a new error type to allow for internal errors as well as 503's  to conform with the design doc:

`if CTS has a scenario where it is considered unhealthy but the API is still being served, we will return a 503 service unavailable error.`

api/health.go
```diff
@@ -0,0 +1,62 @@
...
+
+	err := hh.hm.Check()
```
wilkermichael, line 39: Now I'm leaning towards returning an "isHealthy" boolean check over the error type :S

mkam: I think either would be fine, especially since we're realistically never going to return an error for now! I think the main crux of the decision would be whether we want to treat unhealthy different from not being able to determine the health. 

It may even be that you can return both a boolean and an error and be able to distinguish between the two that way. `false, nil` being we successfully checked the health and it's unhealthy, and `false, "some error"` could be for errors checking the health.

mkam (approved):
Looks good to me! Left some minor comments on naming, but nothing blocking!

api/health.go
wilkermichael: I'll just leave it as an error type for now, we can revisit when we add more robust health checking

api/openapi.yaml
```diff
@@ -500,3 +522,19 @@ components:
           example: "this is an error message"
       required:
         - message
+
+    GoodHealthCheckResponse:
```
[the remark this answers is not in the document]
wilkermichael: @jwair I'll leave this for your PR when you combine these into one

controller/daemon.go
```diff
@@ -60,6 +61,7 @@ func (ctrl *Daemon) Run(ctx context.Context) error {
 	conf := ctrl.tasksManager.state.GetConfig()
 	s, err := api.NewAPI(api.Config{
 		Controller: ctrl.tasksManager,
+		Health:     &health.BasicManager{},
```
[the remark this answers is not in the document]
wilkermichael: That's a really good point! Should we document this somewhere?

PR #824, opened 2022-04-29
Self-Registration with Consul: Configurations, Consul client changes, manager

mkam:
First part of implementing self-registration, mostly setting the groundwork and not actually doing any registration yet. 

**Overview of changes**
- Adds configuration for self-registration, which includes a top-level `id` config and a `self_registration` config under the existing `consul` config.
  - If `id` is not provided, it is automatically generated in the format of `cts-<uuid>`.
  -  A `name` configuration was still under discussion when I started this implementation and will be added in a follow up PR.
- Adds service registration and deregistration methods to the Consul client
  - These methods are not currently taking advantage of the improved retry logic added in https://github.com/hashicorp/consul-terraform-sync/pull/818
- Adds a service registration manager, which can register a service with a default HTTP health check
  - Not yet being used by CTS. Plan to add a `Start()` method that will be called in the daemon controller `Run()`
  - HTTP check is using `/v1/status` for now until `/v1/health` is implemented. Main advantage of `/v1/health` is that requests to it will not be logged, so health check polling will not clutter the logs. (see https://github.com/hashicorp/consul-terraform-sync/pull/823)

### DIFF
--- a/registration/manager.go
+++ b/registration/manager.go
@@ -1,0 +1,129 @@
+package registration
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/consul-terraform-sync/client"
+	"github.com/hashicorp/consul-terraform-sync/config"
+	"github.com/hashicorp/consul-terraform-sync/logging"
+	consulapi "github.com/hashicorp/consul/api"
+)
+
+const (
+	// Service defaults
+	defaultServiceName = "Consul-Terraform-Sync"
+	defaultNamespace   = ""
+
+	// Check defaults
+	defaultCheckName                      = "CTS Health Status"
+	defaultCheckNotes                     = "Check created by Consul-Terraform-Sync"
+	defaultDeregisterCriticalServiceAfter = "30m"
+	defaultCheckStatus                    = consulapi.HealthCritical
+
+	// HTTP-specific check defaults
+	defaultEndpoint      = "/v1/status" // TODO: temporary until /v1/health is implemented
+	defaultMethod        = "GET"
+	defaultInterval      = "10s"
+	defaultTimeout       = "2s"
+	defaultTLSSkipVerify = true
+
+	logSystemName = "registration"
+)
+
+var defaultServiceTags = []string{"cts"}
+
+// RegistrationManager handles the registration of a service and the health checks for that service
+// to Consul.
+type RegistrationManager struct {
+	client  client.ConsulClientInterface
+	service *service
+
+	logger logging.Logger
+}
+
+type RegistrationManagerConfig struct {
+	ID               string
+	Port             int
+	TLSEnabled       bool
+	SelfRegistration *config.SelfRegistrationConfig
+}
+
+type service struct {
+	name      string
+	id        string
+	tags      []string
+	port      int
+	namespace string
+
+	checks []*consulapi.AgentServiceCheck
+}
+
+func NewRegistrationManager(conf *RegistrationManagerConfig, client client.ConsulClientInterface) *RegistrationManager {
+	logger := logging.Global().Named(logSystemName)
+
+	name := defaultServiceName
+
+	ns := defaultNamespace
+	if conf.SelfRegistration.Namespace != nil {
+		ns = *conf.SelfRegistration.Namespace
+	}
+
+	var checks []*consulapi.AgentServiceCheck
+	checks = append(checks, defaultHTTPCheck(conf))
+
+	return &RegistrationManager{
+		client: client,
+		logger: logger,
+		service: &service{
+			name:      name,
+			id:        conf.ID,
+			tags:      defaultServiceTags,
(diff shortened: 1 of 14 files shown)


registration/manager_test.go
```diff
@@ -0,0 +1,221 @@
...
+			assert.Equal(t, tc.expectedService.tags, m.service.tags)
+			assert.Equal(t, tc.expectedService.port, m.service.port)
+			assert.Equal(t, tc.expectedService.namespace, m.service.namespace)
+
+			// Verify service has default health check
+			assert.NotNil(t, m.service.checks)
+			assert.Equal(t, 1, len(m.service.checks))
+			check := m.service.checks[0]
+			defaultCheck := defaultHTTPCheck(tc.conf)
+			assert.Equal(t, defaultCheck, check)
```
lornasong, line 83: Wondering if it's possible to consolidate some of these asserts to something like:

```
assert.NotNil(t, m.service)
tc.expectedService.checks = []*consulapi.AgentServiceCheck{defaultHTTPCheck(tc.conf)}
assert.Equal(t, tc.expectedService, m.service)
```

Maybe it's not very extensible for future test cases though?

registration/manager_test.go
```diff
@@ -0,0 +1,221 @@
...
+	testcases := []struct {
+		name      string
+		setup     func(*mocks.ConsulClientInterface)
+		expectErr bool
+	}{
+		{
+			"success",
+			func(cMock *mocks.ConsulClientInterface) {
+				cMock.On("RegisterService", mock.Anything,
+					mock.MatchedBy(func(r *consulapi.AgentServiceRegistration) bool {
```
lornasong, line 180: Nice! I didn't know about the MatchedBy function, I'll have to try it out!

registration/manager.go
```diff
@@ -0,0 +1,129 @@
...
+	name      string
+	id        string
+	tags      []string
+	port      int
+	namespace string
+
+	checks []*consulapi.AgentServiceCheck
+}
+
+func NewRegistrationManager(conf *RegistrationManagerConfig, client client.ConsulClientInterface) *RegistrationManager {
```
lornasong, line 62: Similar here, would you mind adding comments on this struct and the methods below that are exported?

registration/manager.go
```diff
@@ -0,0 +1,129 @@
...
+	r := &consulapi.AgentServiceRegistration{
+		ID:        s.id,
+		Name:      s.name,
+		Tags:      s.tags,
+		Port:      s.port,
+		Checks:    s.checks,
+		Namespace: s.namespace,
+	}
+
+	m.logger.Debug("registering service with Consul", "name", s.name, "id", s.id)
```
lornasong, line 100: _Super_ minor nit: I was curious on your thoughts on updating the language from "registering service" to "registering CTS" (or "registering self"?!)  in some places - e.g. the method name "RegisterService", this log, and the error below. "registering service" initially gives me the impression that CTS is registering other (non-CTS) services.

This is definitely a personal thing - I think it's easily clarified since service name and id are included. Feel free to take or leave!

lornasong (approved):
Thanks for the helpful PR description! Really like how the code is organized in a RegistrationManager. Comments are minor and non-blocking

registration/manager.go
mkam: Yeah, I think you're right that this would be better to have CTS/self specific wording! 

I had this more general service language because I realized that this manager could technically be used for any kind of service. I didn't really think we'd ever need to use it to register a service other than CTS, but it just worked out like that. It doesn't make sense to keep it vague for a use case that will likely never exist, though, when having CTS specific logs will be a better experience for the user.

registration/manager_test.go
mkam: I like your consolidation, I will update!

registration/manager_test.go
mkam: It's nifty for checking that a request was called with the parameters you expected! I do bang my head a lot when I'm writing them, though, because I tend to get the types wrong (like pointer vs. value) and it's hard to debug why it didn't match. It ends up being a lot of commenting out each attribute check until I figure out which one is messing up!

registration/manager.go
mkam: Going to rename basically everything to have `SelfRegistration` so that it also better matches the `self_registration` config! I also have realized that I've already baked in some self registration specific stuff (like passing a `config.SelfRegistrationConfig` and having the HTTP default check using CTS specific attributes), so it extra doesn't make sense to keep it as just `Registration`. 🙂

mkam:
@lornasong I added new commits to address PR feedback! I rebased on main which is why all of the commits look new.

mkam:
Oh, forgot to update test names, will make that update too!

lornasong (approved):
thanks for the changes, looks good to me!

PR #826, opened 2022-05-03
Unified naming of health checker in code and tests + Task creation refactoring

jwair:
- Unified naming of health checker in code and tests
- Updated oapi-codegen module


health/checker.go
```diff
@@ -2,7 +2,7 @@ package health
 
 import "fmt"
 
-//go:generate mockery --name=Manager --filename=manager.go --output=../mocks/health
+//go:generate mockery --name=Checker --filename=checker.go --output=../mocks/health
```
jwair, line 5: This is the main change in the PR, to unify the use of 'Checker' over 'Manager'

wilkermichael: ah thanks for catching this, I did the refactor but then didn't catch these

e2e/condition_catalog_service_test.go
```diff
@@ -369,10 +369,6 @@ func TestCondition_CatalogServices_SharedDependency_NoServices(t *testing.T) {
 		name    string
 		runMode string
 	}{
-		{
```
jwair, line 372: Removed this invalid test case. Empty string for `run` is not a valid value.

api/task_lifecycle_client.go
```diff
@@ -59,29 +59,12 @@ func (c *TaskLifecycleClient) Scheme() string {
...
-		return nil, err
-	}
-
-	resp, err := c.CreateTaskWithResponse(ctx, &oapigen.CreateTaskParams{Run: &run}, oapigen.CreateTaskJSONRequestBody(req))
-	if err != nil {
-		return nil, err
-	}
-
-	return resp, nil
+	run := oapigen.CreateTaskParamsRun(runOption)
```
jwair, line 66: Swagger validation happens as part of `c.CreateTaskWithResponse()` and that validates the run parameter is in the enum of supported options, no need for our additional validation logic.

api/task_lifecycle_client.go
```diff
@@ -59,31 +58,6 @@ func (c *TaskLifecycleClient) Scheme() string {
 	return c.url.Scheme
 }
 
-// CreateTaskSimple takes a task request and run option and sends this information to the client. It then returns
-// a task response object and any errors to the caller.
-// TODO: remove this to conform to interface
-func (c *TaskLifecycleClient) CreateTask(ctx context.Context, runOption string, req TaskRequest) (*oapigen.CreateTaskResponse, error) {
-	var run oapigen.CreateTaskParamsRun
-	switch runOption {
```
jwair, line 67: Swagger validation happens in the generated client code, and that validates the run parameter is in the enum of supported options, no need for our additional validation logic.

e2e/condition_catalog_service_test.go
wilkermichael: actually in this case, we should verify that the nil case works. This test is verifying that the task creates without issue, and creating a task without a `run` parameter is a valid scenario from the API

wilkermichael (approved):
Comment about the nil `run` parameter case, otherwise everything else looks good

e2e/condition_catalog_service_test.go
jwair: Good catch.
Added back a test case for creating a task without run mode: https://github.com/hashicorp/consul-terraform-sync/pull/826/commits/e8737b9bfd9fe4ec549f9e9dbe9b91a8a21d744d

PR #827, opened 2022-05-03
Add config key to the promote-staging event

claire-labry:
We need to pass the `release-metadata.hcl` filename to the promote-staging workflow prior to the RelAPI launch on May 9th.

Note: it looks like my `.hcl` linter found a missing `}` at the end of the `promote-production-packaging` event -- LMK if it was intentional to leave that out.


lornasong (approved):
makes sense to me!

mdeggies (approved):

PR #831, opened 2022-05-03
Refactor monitoring/triggering out of TasksManager (pt1)

lornasong:
First PR of two to refactor monitoring/triggering out of TasksManager. The intent is to keep TasksManager scoped to managing tasks (crud + run). A separate struct (currently named planning to be named ConditionMonitor) will be used to monitor changes and trigger task runs using the TasksManager.

The changes to disentangle monitoring/triggering code out of TasksManager is pretty large. This first PR contains miscellaneous commits that could be separated out without breaking tests and compilation. The bulk of the refactor will be in the next PR.

### DIFF
--- a/controller/daemon.go
+++ b/controller/daemon.go
@@ -4,10 +4,12 @@ import (
 	"context"
 
 	"github.com/hashicorp/consul-terraform-sync/api"
+	"github.com/hashicorp/consul-terraform-sync/client"
 	"github.com/hashicorp/consul-terraform-sync/config"
 	"github.com/hashicorp/consul-terraform-sync/health"
 	"github.com/hashicorp/consul-terraform-sync/logging"
 	"github.com/hashicorp/consul-terraform-sync/state"
+	"github.com/hashicorp/consul-terraform-sync/templates"
 )
 
 var (
@@ -25,6 +27,7 @@ type Daemon struct {
 
 	state        state.Store
 	tasksManager *TasksManager
+	watcher      templates.Watcher
 
 	// whether or not the tasks have gone through once-mode. intended to be used
 	// by benchmarks to run once-mode separately
@@ -38,7 +41,13 @@ func NewDaemon(conf *config.Config) (*Daemon, error) {
 
 	s := state.NewInMemoryStore(conf)
 
-	tm, err := NewTasksManager(conf, s)
+	logger.Info("initializing Consul client and testing connection")
+	watcher, err := newWatcher(conf, client.ConsulDefaultMaxRetry)
+	if err != nil {
+		return nil, err
+	}
+
+	tm, err := NewTasksManager(conf, s, watcher)
 	if err != nil {
 		return nil, err
 	}
@@ -47,6 +56,7 @@ func NewDaemon(conf *config.Config) (*Daemon, error) {
 		logger:       logger,
 		state:        s,
 		tasksManager: tm,
+		watcher:      watcher,
 	}, nil
 }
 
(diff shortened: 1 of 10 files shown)


lornasong:
rebased to latest on main to resolve conflicts in a child branch

lornasong:
tests passing!

wilkermichael:
@lornasong I wasn't able to do this today, I will take a look first thing tomorrow

wilkermichael:
@lornasong taking a look now!

controller/daemon.go
```diff
@@ -126,7 +136,7 @@ func (ctrl *Daemon) Once(ctx context.Context) error {
 }
 
 func (ctrl *Daemon) Stop() {
```
wilkermichael, line 138: I really like how you pulled the watcher out of taskManager

controller/daemon.go
```diff
@@ -38,7 +41,13 @@ func NewDaemon(conf *config.Config) (*Daemon, error) {
 
 	s := state.NewInMemoryStore(conf)
 
-	tm, err := NewTasksManager(conf, s)
+	logger.Info("initializing Consul client and testing connection")
```
wilkermichael, line 44: Can we add a debug log to log which controller was initialized?

wilkermichael: same for other controllers

controller/tasks_manager.go
```diff
@@ -47,16 +46,9 @@ type TasksManager struct {
 }
 
 // NewTasksManager configures a new tasks manager
-func NewTasksManager(conf *config.Config, state state.Store) (*TasksManager, error) {
-
+func NewTasksManager(conf *config.Config, state state.Store, watcher templates.Watcher) (*TasksManager, error) {
```
wilkermichael, line 49: Now that the inputs are expanding, do you think this warrants a constructor config object? I think 3 is okay, but if we add more it might be worth it

controller/tasks_manager.go
```diff
@@ -404,6 +404,15 @@ func (tm *TasksManager) checkApply(ctx context.Context, d driver.Driver, retry,
 	return rendered, nil
 }
 
+// EnableTestMode is a helper for testing which tasks were triggered and
+// executed. Callers of this method must consume from TaskNotify channel to
+// prevent the buffered channel from filling and causing a dead lock.
+func (tm *TasksManager) EnableTestMode() <-chan string {
```
wilkermichael, line 410: So this method sets the channels for `tm.taskNotify`. I know that this method is used for testing, but there isn't anything being done here that feels like it couldn't be used for production. What I mean is, if in the future, if we wanted to have some entity that monitored live task runs, this method could be used for that.

Maybe a name that doesn't include Test would be good? `SetNotificationChannels` or `EnableTaskNotify`?

wilkermichael (commented):
@lornasong looks good, nice refactor!

Just some minor things

controller/tasks_manager.go
lornasong: Good idea, I could rename this to something like `EnableTaskRunNotify` (in a subsequent PR I added something like a `EnableTaskDeleteNotify`). If it's ok with you I'd like to do this renaming in a separate PR as it will create conflicts with another PR I already opened.

controller/daemon.go
lornasong: There is a log that lets the user know which controller was initialized ~5 lines above. Do you think that would be sufficient?
`logger.Info("setting up controller", "type", "daemon")`

wilkermichael: oh yep, all good!

controller/tasks_manager.go
wilkermichael: that works for me!

wilkermichael (approved):
Since we agreed on no further changes, I'm okay with approving

PR #832, opened 2022-05-04
Self-register CTS as a service with Consul in daemon mode

mkam:
Adds support for CTS to register itself with Consul when running in daemon mode.

- Updates the health check to use `/v1/health` now that has been implemented
- Adds a Start method to the self-registration manager which will register the service and then wait for an initiated shutdown to deregister the service
- Calls Start() in the daemon controller's Run() to start the registration and monitoring.
  - Does not error and exit if there are registration issues, relies on the error logging from the self-registration manager.
  - Explicitly deregister CTS if once-mode fails

End-to-end tests will be added in a follow up PR after the remaining work is complete (new configurations, improved retry logic).

### DIFF
--- a/controller/daemon.go
+++ b/controller/daemon.go
@@ -76,6 +76,19 @@ func (ctrl *Daemon) Run(ctx context.Context) error {
 		return err
 	}
 
+	// Serve API
+	go func() {
+		err := s.Serve(ctx)
+		exitCh <- err
+	}()
+
+	// Run tasks once through once-mode
+	if !ctrl.once {
+		if err := ctrl.Once(ctx); err != nil {
+			return err
+		}
+	}
+
 	var rm *registration.SelfRegistrationManager
 	if *conf.Consul.SelfRegistration.Enabled {
 		// Expect one more long-running goroutine
@@ -101,28 +114,11 @@ func (ctrl *Daemon) Run(ctx context.Context) error {
 				SelfRegistration: conf.Consul.SelfRegistration,
 			},
 			ctrl.consulClient)
-			
+
 		go func() {
 			rm.Start(ctx)
 			exitCh <- nil // registration errors are logged only
 		}()
-	}
-
-	// Serve API
-	go func() {
-		err := s.Serve(ctx)
-		exitCh <- err
-	}()
-
-	// Run tasks once through once-mode
-	if !ctrl.once {
-		if err := ctrl.Once(ctx); err != nil {
-			if rm != nil {
-				// Deregister CTS service from Consul
-				rm.Deregister(ctx)
-			}
-			return err
-		}
 	}
 
 	// Run tasks in long-running mode
--- a/controller/daemon.go
+++ b/controller/daemon.go
@@ -4,9 +4,11 @@ import (
 	"context"
 
 	"github.com/hashicorp/consul-terraform-sync/api"
+	"github.com/hashicorp/consul-terraform-sync/client"
 	"github.com/hashicorp/consul-terraform-sync/config"
 	"github.com/hashicorp/consul-terraform-sync/health"
 	"github.com/hashicorp/consul-terraform-sync/logging"
+	"github.com/hashicorp/consul-terraform-sync/registration"
 	"github.com/hashicorp/consul-terraform-sync/state"
 )
 
@@ -25,6 +27,8 @@ type Daemon struct {
 
 	state        state.Store
 	tasksManager *TasksManager
+
+	consulClient client.ConsulClientInterface
 
 	// whether or not the tasks have gone through once-mode. intended to be used
 	// by benchmarks to run once-mode separately
@@ -57,7 +61,10 @@ func (ctrl *Daemon) Init(ctx context.Context) error {
 }
 
 func (ctrl *Daemon) Run(ctx context.Context) error {
-	// Serve API
+	exitBufLen := 2 // api & run tasks exit
+	exitCh := make(chan error, exitBufLen)
+
+	// Configure API
 	conf := ctrl.tasksManager.state.GetConfig()
 	s, err := api.NewAPI(api.Config{
 		Controller: ctrl.tasksManager,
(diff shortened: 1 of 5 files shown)


mkam:
@lornasong I've left this in draft because I'm not sure why, but `TestE2ERestartConsul` is failing intermittently with these changes where it doesn't detect the change in service after Consul restarts. 😞

registration/manager_test.go
```diff
@@ -141,7 +143,140 @@ func TestSelfRegistrationManager_defaultHTTPCheck(t *testing.T) {
...
+		logger: logging.NewNullLogger(),
+		// mock client will be set per test case
+	}
+
+	t.Run("success", func(t *testing.T) {
+		mockClient := new(mocks.ConsulClientInterface)
+		mockClient.On("DeregisterService", mock.Anything,
+			mock.MatchedBy(func(actual string) bool {
+				return actual == id
+			})).Return(nil)
```
jwair, line 258:
```suggestion
		mockClient.On("DeregisterService", mock.Anything, id).Return(nil)
```

controller/daemon.go
```diff
@@ -69,8 +76,38 @@ func (ctrl *Daemon) Run(ctx context.Context) error {
...
+		// Configure and start self-registration manager
+		rm = registration.NewSelfRegistrationManager(
+			&registration.SelfRegistrationManagerConfig{
+				ID:               *conf.ID,
+				Port:             *conf.Port,
+				TLSEnabled:       (conf.TLS != nil && *conf.TLS.Enabled),
+				SelfRegistration: conf.Consul.SelfRegistration,
+			},
+			ctrl.consulClient)
+		go func() {
```
jwair, line 104: For better readability, I suggest spacing out code blocks specially if indentations and parenthesis are involved :)

jwair, line 104:
```suggestion
			ctrl.consulClient)
			
		go func() {
```

jwair: For better readability, I suggest spacing out code blocks specially if indentations and parenthesis are involved :)

registration/manager_test.go
```diff
@@ -141,7 +143,140 @@ func TestSelfRegistrationManager_defaultHTTPCheck(t *testing.T) {
...
+
+		ctx, cancel := context.WithCancel(context.Background())
+		errCh := make(chan error)
+		go func() {
+			if err := manager.Start(ctx); err != nil {
+				errCh <- err
+			}
+		}()
+		cancel()
+		select {
```
jwair, line 178:
```suggestion
		cancel()
		
		select {
```

registration/manager_test.go
```diff
@@ -141,7 +143,140 @@ func TestSelfRegistrationManager_defaultHTTPCheck(t *testing.T) {
...
+
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		errCh := make(chan error)
+		go func() {
+			if err := manager.Start(ctx); err != nil {
+				errCh <- err
+			}
+		}()
+		select {
```
jwair, line 202:
```suggestion
		}()
		
		select {
```

registration/manager_test.go
mkam: Oh, sorry I misread your suggestion! That makes sense to me, `MatchedBy` would be better for when you don't need an exact match but an approximate custom match instead.

jwair: Actually, even passing ctx as the first arg would work.
i.e.
`mockClient.On("DeregisterService", ctx, id).Return(nil)`

lornasong (commented):
Did an initial review, make sense to me! I can review again once the e2e test gets sorted out

controller/daemon.go
```diff
@@ -95,8 +136,6 @@ func (ctrl *Daemon) Run(ctx context.Context) error {
 		counter++
 		if err != nil && err != context.Canceled {
 			// Exit if an error is returned
-			// Not expecting any routines to send a nil error because they run
-			// until canceled. Nil check is just to be safe
```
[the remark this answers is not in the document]
mkam: Yeah, because we don't want CTS to exit if registration errors, I'm returning always sending `nil` on to the exit channel when the self-manager goroutine ends. This would either be on a graceful exit where no error is returned by Start() or if registration or deregistration fails in which case an error is returned by Start() but we're ignoring it.

https://github.com/hashicorp/consul-terraform-sync/pull/832/files#diff-557dd1981617dc7e0ec1c5d81fda203c5dc5325fa3c59b5ff23083c8b39bdd95R106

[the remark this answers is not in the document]
lornasong: Ah, thank you for explaining and pointing out the nil to exitCh. That makes sense

registration/manager_test.go
mkam: Will update other tests with this format where relevant!

registration/manager_test.go
```diff
@@ -141,7 +143,139 @@ func TestSelfRegistrationManager_defaultHTTPCheck(t *testing.T) {
...
+	t.Run("start registers, cancel deregisters", func(t *testing.T) {
+		mockClient := new(mocks.ConsulClientInterface)
+		mockClient.On("RegisterService", mock.Anything,
+			mock.MatchedBy(func(r *consulapi.AgentServiceRegistration) bool {
+				return r.ID == id && r.Name == defaultServiceName
+			})).Return(nil)
+		mockClient.On("DeregisterService", mock.Anything,
+			mock.MatchedBy(func(actual string) bool {
+				return actual == id
+			})).Return(nil)
```
jwair, line 167:
```suggestion
		mockClient.On("DeregisterService", mock.Anything, id).Return(nil)
```

This one can potentially be switched to a simpler mock?

mkam: Yup, will review all of them and make changes where needed!

controller/daemon_test.go
```diff
@@ -2,12 +2,14 @@ package controller
...
 import (
 	"context"
+	"errors"
 	"testing"
 	"time"
 
 	"github.com/hashicorp/consul-terraform-sync/config"
 	"github.com/hashicorp/consul-terraform-sync/driver"
 	"github.com/hashicorp/consul-terraform-sync/logging"
+	mocks "github.com/hashicorp/consul-terraform-sync/mocks/client"
```
jwair, line 12: for consistency with the two following imports, can we change this to `mocksC` or something like that?

registration/manager.go
```diff
@@ -88,9 +88,28 @@ func NewSelfRegistrationManager(conf *SelfRegistrationManagerConfig, client clie
...
+	<-ctx.Done()
+	err = m.deregister(ctx)
+	if err != nil {
+		return err
+	}
+	return ctx.Err()
+}
+
+// register registers Consul-Terraform-Sync with Consul
+func (m *SelfRegistrationManager) register(ctx context.Context) error {
```
mkam, line 110: @lornasong I renamed this to be simply `register()` from `SelfRegisterService()` to simplify it and keep parity with `deregister()`, but left all the logs that still clarify that it's a self-registration of a service. Additionally, I was thinking that now that it's a `SelfRegistrationManager`, then the `Self` part was taken care of. Open to discussion, though, since I know you had some concerns earlier!

lornasong: I like `register()`! I think it removes redundancy as you're saying

lornasong (approved):
looks good to me!

PR #834, opened 2022-05-04
Health check responses consolidation + adding request ID to responses

jwair:
Consolidated health check response objects
Added missing request ID to the response


wilkermichael:
@mkam AJ and I were talking about this PR, so I can go ahead and review it

api/health_test.go
```diff
@@ -30,11 +30,9 @@ func Test_HealthHandler_GetHealth(t *testing.T) {
 			validate: func(t *testing.T, resp *httptest.ResponseRecorder) {
```
wilkermichael, line 30: I think we should also add checks that we are setting the requestID as expected

You can define some request ID outside the for loop and then just add it to the context of the request
```Go
path := "/v1/health"
req, err := http.NewRequest(http.MethodGet, path, nil)
require.NoError(t, err)
resp := httptest.NewRecorder()

req = req.WithContext(requestIDWithContext(req.Context(), reqID))
```

wilkermichael (approved):
One comment on testing, otherwise looks good!

Thanks for doing this!

api/health_test.go
jwair: Thanks, I'll add that

jwair: Addressed in https://github.com/hashicorp/consul-terraform-sync/pull/834/commits/9aea60d770d2b39bcc18b4495bd173cb5e46dfe1

wilkermichael (approved):

PR #838, opened 2022-05-04
Remove logging from health endpoint

wilkermichael:
# Summary
Remove logging from `GET v1/health` endpoint call. This is required since we expect the health endpoint to get a lot of "hits" and we do not want to flood the logs

# Challenges
Oapi-codegen generated servers take an all or nothing approach to applying middleware to URIs. We can separate middleware application into groups, however, application of middleware to the oapi-codegen handler is done all at once, with no real way to exclude certain URIs from the middleware (i.e. `v1/health`)

### How we apply middleware to the group
https://github.com/hashicorp/consul-terraform-sync/blob/1bcc8ac9c73847da6f0bc256bd73f249fdcafbbd/api/api.go#L116-L129

### How routing is handled in the generated server
https://github.com/hashicorp/consul-terraform-sync/blob/1bcc8ac9c73847da6f0bc256bd73f249fdcafbbd/api/oapigen/server.go#L258-L292

# Solution
I've opted to handle the exclusion in the middleware itself, the logging middleware now has an associated object and constructor. The constructor can take a list of URIs to be excluded. 

Ideally it would be better to apply the middleware only to the paths that require it, however, with the code generation we are limited in how we can handle this.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -94,11 +94,13 @@ func NewAPI(conf Config) (*API, error) {
 	r := chi.NewRouter()
 
 	// add the middleware for all endpoints
-	r.Use(withLogging)
 	r.Use(withCORS)
+	r.Use(withRequestID)
 
 	// add the base path route then mount the endpoints
 	r.Route(fmt.Sprintf("/%s", defaultAPIVersion), func(r chi.Router) {
+		lm := newLoggingMiddleware(nil)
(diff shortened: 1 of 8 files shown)


mkam (approved):
Looks good to me! Thanks for all the details on the challenges, I'm sure they'll be a good reference in the future!

PR #840, opened 2022-05-06
Improved 'default' values output in CLI help

jwair:
Improved 'default' values output in CLI help
Simplified CLI help generation, improved naming

### DIFF
--- a/command/cli.go
+++ b/command/cli.go
@@ -78,10 +78,8 @@ func (cli *CLI) Run(args []string) int {
 		Commands:                   Commands(),
 		Autocomplete:               true,
 		AutocompleteNoDefaultFlags: true,
-		HelpFunc: cli.groupedHelpFunc(
-			mcli.BasicHelpFunc("cts"),
-		),
-		HelpWriter: tabwriter.NewWriter(os.Stdout, 0, 2, 4, ' ', tabwriter.AlignRight),
+		HelpFunc:                   help,
+		HelpWriter:                 tabwriter.NewWriter(os.Stdout, 0, 2, 4, ' ', tabwriter.AlignRight),
 	}
 
 	if subcommands.IsVersion() {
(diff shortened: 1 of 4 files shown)


wilkermichael:
Can you remove the portion of the usage that describes the default now that you've added it in the help formatting?

https://github.com/hashicorp/consul-terraform-sync/blob/9b483385f8d4ae3b359f9659f760cfe2e1e8cf6e/command/meta.go#L62-L66

command/cli.go
```diff
@@ -98,18 +96,16 @@ func (cli *CLI) Run(args []string) int {
...
+	c := make(map[string]string)
+	for _, v := range commonCommands {
+		c[v] = generateCommandHelp(v, commands[v])
 	}
+
+	return generateHelp(c, "Usage CLI: consul-terraform-sync <command> [-help] [options]\n", nil)
 }
 
-func helpFunc(commands map[string]string, usage string, omitFlags []string) string {
+func generateHelp(commands map[string]string, usage string, omitFlags []string) string {
```
wilkermichael, line 108:
```suggestion
// To support usage by the start command, this function takes a map[string]string and 
// allows for omission of flags. This functionality will not be needed once running CTS 
// without command arguments is deprecated in a future major release.
func generateHelp(commands map[string]string, usage string, omitFlags []string) string {
```

wilkermichael: to add more context on why the function signature includes a map[string]string instead of the command factory

wilkermichael (approved):
Looks good AJ! Nice work.

jwair:
> Can you remove the portion of the usage that describes the default now that you've added it in the help formatting?
> 
> https://github.com/hashicorp/consul-terraform-sync/blob/9b483385f8d4ae3b359f9659f760cfe2e1e8cf6e/command/meta.go#L62-L66

Addressed in https://github.com/hashicorp/consul-terraform-sync/pull/840/commits/b6aae0fb3963b123f0275d8a3937528ba98e2aa4

PR #846, opened 2022-05-09
TasksManager & ConditionMonitor refactor cleanup

lornasong:
Context: previously merged in a number of prs refactoring TasksManager and ConditionMonitor out of controllers.

Changes: various cleanup from feedback and minor issues noticed during refactoring

### DIFF
--- a/api/overall_status.go
+++ b/api/overall_status.go
@@ -84,10 +84,6 @@ func (h *overallStatusHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 		}
 
 		tasks := h.ctrl.Tasks(ctx)
-		if err != nil {
-			jsonErrorResponse(ctx, w, http.StatusInternalServerError, err)
-			return
-		}
 		for _, task := range tasks {
 			// look for any tasks that have a driver but no events
 			if _, ok := data[*task.Name]; !ok {
--- a/api/overall_status.go
+++ b/api/overall_status.go
@@ -83,7 +83,7 @@ func (h *overallStatusHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 			}
 		}
 
-		tasks, err := h.ctrl.Tasks(ctx)
+		tasks := h.ctrl.Tasks(ctx)
 		if err != nil {
 			jsonErrorResponse(ctx, w, http.StatusInternalServerError, err)
 			return
--- a/api/representation.go
+++ b/api/representation.go
@@ -216,10 +216,10 @@ func (tr TaskRequest) String() string {
 
 type TasksResponse oapigen.TasksResponse
 
-func tasksResponseFromTaskConfigs(tcs []config.TaskConfig, requestID oapigen.RequestID) TasksResponse {
+func tasksResponseFromTaskConfigs(tcs config.TaskConfigs, requestID oapigen.RequestID) TasksResponse {
 	tasks := make([]oapigen.Task, len(tcs))
 	for i, tc := range tcs {
-		tasks[i] = oapigenTaskFromConfigTask(tc)
+		tasks[i] = oapigenTaskFromConfigTask(*tc)
 	}
 
 	return TasksResponse{
--- a/controller/condition_monitor.go
+++ b/controller/condition_monitor.go
@@ -73,11 +73,8 @@ func (cm *ConditionMonitor) Run(ctx context.Context) error {
 	if cm.watcherCh == nil {
 		// Size of channel is the number of CTS tasks configured at initialization
 		// +10 for any additional tasks created during runtime. 10 arbitrarily chosen
-		tasks, err := cm.tasksManager.Tasks(ctx)
-		if err != nil {
-			return err
-		}
-		cm.watcherCh = make(chan string, len(tasks)+10)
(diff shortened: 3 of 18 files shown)


api/overall_status.go
```diff
@@ -83,7 +83,7 @@ func (h *overallStatusHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 			}
 		}
 
-		tasks, err := h.ctrl.Tasks(ctx)
+		tasks := h.ctrl.Tasks(ctx)
 		if err != nil {
```
jwair, line 87: Is this error check still providing value, since the error assignment was removed from line 86?

api/representation.go
```diff
@@ -216,10 +216,10 @@ func (tr TaskRequest) String() string {
 
 type TasksResponse oapigen.TasksResponse
 
-func tasksResponseFromTaskConfigs(tcs []config.TaskConfig, requestID oapigen.RequestID) TasksResponse {
+func tasksResponseFromTaskConfigs(tcs config.TaskConfigs, requestID oapigen.RequestID) TasksResponse {
 	tasks := make([]oapigen.Task, len(tcs))
 	for i, tc := range tcs {
-		tasks[i] = oapigenTaskFromConfigTask(tc)
+		tasks[i] = oapigenTaskFromConfigTask(*tc)
```
jwair, line 222: Nit, should this method name now be `oapigenTaskFromTaskConfig` ?

controller/condition_monitor.go
```diff
@@ -59,15 +59,14 @@ func (cm *ConditionMonitor) WatchDep(ctx context.Context) error {
...
-// Run runs the controller in read-write mode by continuously monitoring Consul
-// catalog and using the driver to apply network infrastructure changes for
-// any work that have been updated.
+// Run runs a continuous loop monitoring Consul catalog and, on changes, uses
+// the TasksManager to execute tasks to update network infrastructure
 //
-// Blocking call runs the main consul monitoring loop, which identifies triggers
+// Blocking call runs the main Consul monitoring loop, which identifies triggers
 // for dynamic tasks. Scheduled tasks use their own go routine to trigger on
 // schedule.
```
wilkermichael, line 67: Just a suggestion on rephrasing some of this method comment:

```Go
// Run runs a continuous loop which monitors the Consul catalog. On changes, run
// uses the TasksManager to execute tasks to update network infrastructure.
// 
// The blocking call runs the main Consul monitoring loop, which identifies triggers
// for dynamic tasks. Scheduled tasks use their own go routine to trigger on
// schedule.
```

wilkermichael (commented):
Great work Lorna, some minor comments

api/overall_status.go
lornasong: Thank you for catching

controller/tasks_manager.go
```diff
@@ -442,13 +442,13 @@ func (tm TasksManager) TaskByTemplate(tmplID string) (string, bool) {
 	return driver.Task().Name(), true
 }
 
-// EnableTestMode is a helper for testing which tasks were triggered and
-// executed. Callers of this method must consume from TaskNotify channel to
-// prevent the buffered channel from filling and causing a dead lock.
-func (tm *TasksManager) EnableTestMode() <-chan string {
+// EnableTaskRanNotify is a helper for testing which tasks were triggered and
```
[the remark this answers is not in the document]
lornasong: I actually prefer to keep the testing context within the comment. A while back, it originally wasn't there and this method and its associated channel took me some extra effort to understand while making a somewhat related change. Someone added that context in the comment and I personally found it helpful.

If it's better, I can add it at the end e.g. "[comment text without testing language]. Typically used for testing."

api/representation.go
lornasong: Would you say more about why the method should be renamed now? I believe the naming is the result of the thought "`opaigen.Task` from `config.Task`". The function content and signature also hasn't really changed in this PR, so would prefer not to rename the it as part of this PR to keep it focused on TasksManager / ConditionMonitor refactoring

controller/tasks_manager.go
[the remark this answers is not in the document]
wilkermichael: yeah, I like that! Let's add the context of a typical usecase instead

lornasong:
@wilkermichael and @jwair this is ready for re-review when you have time

controller/tasks_manager.go
```diff
@@ -435,18 +435,20 @@ func (tm TasksManager) TaskByTemplate(tmplID string) (string, bool) {
 	return driver.Task().Name(), true
 }
 
-// EnableTaskRanNotify is a helper for testing which tasks were triggered and
-// finished executing. Callers of this method must consume from ranTaskNotify
-// channel to prevent the buffered channel from filling and causing a dead lock.
+// EnableTaskRanNotify is a helper for enabling notifications when a task has
+// finished executing after being triggered. Callers of this method must consume
+// from ranTaskNotify channel to prevent the buffered channel from filling and
+// causing a dead lock. EnableTaskRanNotify is typically used for testing.
```
wilkermichael, line 441: Perfect!

wilkermichael (approved):
Looks good @lornasong thanks for addressing!

PR #851, opened 2022-05-10
Added dependabot config for go modules version updates

jwair:
This PR adds a dependabot config that targets `dependabot-test` branch for gomod version management. It is experimental for now and will start targeting `main` after evaluating the results.


lornasong (approved):
makes sense to me

PR #858, opened 2022-05-10
End-to-end tests for service registration

mkam:
Adds e2e tests for service registration and various additional Consul testutil methods to verify the registered service and health check.

Sorry for the extensive PR, there were a lot of changes needed to support the various use cases. An overview of the non-test changes are:

Consul testutil additions:
- Listing all instances for a service and filtering by name (for the case where ID is generated and not known)
- Retrieving a health check and waiting until it has a certain status
- Getting a specific service instance by ID
- Waiting for a service to be registered and deregistered

Config changes:
- Add options to the basic Consul config
- Add append method for global `id` configuration

Module additions:
- Added a module that fails on runtime (rather than on initialization)

And this wasn't  really a part of the tests, but I wanted to make some improvements to the retry logging which I noted while testing the use case where registration fails with a non-retryable error.
- Fixed the log system name
- Added more details for logging a non-retryable error
- Removed `this error is not retryable` since the user

Closes https://github.com/hashicorp/consul-terraform-sync/issues/860


e2e/registration_test.go
```diff
@@ -155,49 +155,68 @@ func TestE2E_ServiceRegistration_DeregisterWhenStopped(t *testing.T) {
 	assert.False(t, registered)
 }
 
-// TestE2E_ServiceRegistration_RegistrationFailure tests that if CTS is unable
-// to register itself as a service, it will not exit.
-func TestE2E_ServiceRegistration_RegistrationFailure(t *testing.T) {
+// TestE2E_ServiceRegistration_RegistrationSkipped tests scenarios where registration does
+// not happen, but CTS continues to run.
+func TestE2E_ServiceRegistration_RegistrationSkipped(t *testing.T) {
```
mkam, line 160: The diff for this commit is hard to read, but I basically just put the old test in a for loop where the only difference between the test cases is the `service_registration` config. From a test perspective, all the setup and validation was the same whether the config disabled the registration or the config caused an error with registration.

testutils/consul.go
```diff
@@ -144,6 +210,67 @@ func DeleteKV(tb testing.TB, srv *testutil.TestServer, key string) {
...
+
+	require.Equal(tb, resp.StatusCode, 200, string(b))
+
+	var checks map[string]Check
+	err = json.Unmarshal(b, &checks)
+	require.NoError(tb, err)
+	return checks
+}
+
+func WaitForCheckStatus(tb testing.TB, srv *testutil.TestServer, checkID, status string, wait time.Duration) (Check, error) {
```
mkam, line 241: Something I've noted is that we have a lot of WaitFor methods that all have the same structure. I'd like to refactor this in the future so that they all use a common method for polling, kind of in the same way we have a generic retry structure in the CTS code. I'll be making an issue for this.

mkam: Opened https://github.com/hashicorp/consul-terraform-sync/issues/859 to address this.

testutils/consul.go
```diff
@@ -118,6 +119,27 @@ func registerConsulService(tb testing.TB, srv *testutil.TestServer,
...
 }
 
+func ListServices(tb testing.TB, srv *testutil.TestServer, filter string) map[string]testutil.TestService {
+	u := fmt.Sprintf("http://%s/v1/agent/services", srv.HTTPAddr)
+	if filter != "" {
+		params := url.Values{}
+		params.Add("filter", filter)
+		u = fmt.Sprintf("%s?%s", u, params.Encode())
+	}
+	resp := RequestHTTP(tb, http.MethodGet, u, "")
```
wilkermichael, line 148: You don't have to for this PR, but might be useful to wrap this in a retry

e2e/config.go
```diff
@@ -37,16 +37,21 @@ func (c hclConfig) appendPort(port int) hclConfig {
 	return c.appendString(fmt.Sprintf("port = %d", port))
 }
 
-func (c hclConfig) appendConsulBlock(consul *testutil.TestServer) hclConfig {
+func (c hclConfig) appendConsulBlock(consul *testutil.TestServer, opts ...string) hclConfig {
+	var optsConfig string
+	if len(opts) > 0 {
+		optsConfig = "\n" + strings.Join(opts, "\n")
```
wilkermichael, line 43: You add a newline when you do the `Sprintf` so I don't think you need the first `"\n"`

Makes it slightly easier to read

e2e/registration_test.go
```diff
@@ -61,6 +62,63 @@ func TestE2E_ServiceRegistration_Defaults(t *testing.T) {
...
+		address = "%s"
+	}
+}`, serviceName, address)
+
+	configPath := filepath.Join(tempDir, configFile)
+	config := baseConfig(tempDir).appendID(id).appendPort(port).
+		appendConsulBlock(srv, rConfig).appendTerraformBlock().appendDBTask()
+	config.write(t, configPath)
+
+	cmd := exec.Command("consul-terraform-sync", fmt.Sprintf("--config-file=%s", configPath))
```
wilkermichael, line 98: It took me a few minutes to figure out why the test was using a CTS command vs using the `ctsSetup` method. Can you add a comment just explaining that it's using a custom configuration file to test the `service_registration` parameters and can't use the default config

Something like that

wilkermichael: Side note, any reason not to use `api.StartCTS(t, configPath)` instead?

e2e/registration_test.go
```diff
@@ -73,7 +131,8 @@ func TestE2E_ServiceRegistration_DeregisterWhenStopped(t *testing.T) {
 	id := "cts-01"
 	configPath := filepath.Join(tempDir, configFile)
 	config := baseConfig(tempDir).appendID(id).
-		appendConsulBlock(srv).appendTerraformBlock().appendDBTask()
+		appendConsulBlock(srv).appendTerraformBlock().
+		appendModuleTask("disabled_task", "mkam/hello/cts", "enabled = false")
```
wilkermichael, line 140: Why does the task need to be disabled?

retry/retry.go
```diff
@@ -60,7 +60,7 @@ func (r Retry) Do(ctx context.Context, f func(context.Context) error, desc strin
 	if err == nil || r.maxRetry == 0 {
 		return err
 	} else if errors.As(err, &nonRetryableError) {
-		r.logger.Debug("received non retryable error")
+		r.logger.Debug("received non-retryable error", "method", desc)
```
wilkermichael, line 63: Should other instances of `description` be changed to `method` for consistency?

wilkermichael: and should these be included in all logs in this function?

e2e/registration_test.go
```diff
@@ -150,6 +151,51 @@ func TestE2E_ServiceRegistration_DeregisterWhenStopped(t *testing.T) {
 	assert.False(t, registered)
 }
 
+// TestE2E_ServiceRegistration_RegistrationFailure tests that if CTS is unable
+// to register itself as a service, it will not exit.
+func TestE2E_ServiceRegistration_RegistrationFailure(t *testing.T) {
```
wilkermichael: Maybe include the test case being tested in the name? 

`TestE2E_ServiceRegistration_ContinueOnRegistrationFailure` or something like that

e2e/registration_test.go
```diff
@@ -200,3 +201,32 @@ func getServiceInstancesByName(t testing.TB, srv *testutil.TestServer, serviceNa
...
+
+func waitForServiceByName(t testing.TB, srv *testutil.TestServer, serviceName string) (map[string]testutils.Service, error) {
+	var instances map[string]testutils.Service
+	polling := make(chan struct{})
+	stopPolling := make(chan struct{})
+	go func() {
+		for {
+			select {
+			case <-stopPolling:
+				return
```
wilkermichael, line 283: maybe add a tiny sleep to free up resources

e2e/registration_test.go
```diff
@@ -197,6 +200,54 @@ service_registration {
...
+	select {
+	case <-exitCh:
+		break
+	case <-time.After(defaultWaitForTestReadiness):
+		t.Fatal("timed out waiting for CTS initialization to fail")
+	}
+
+	// Verify no attempt at registration and no service registered
+	output := buf.String()
+	assert.NotContains(t, output, "registering Consul-Terraform-Sync as a service with Consul")
```
wilkermichael, line 265: I was wondering how you would know it didn't try to register vs deregistered on exit. Nice!

e2e/registration_test.go
```diff
@@ -197,6 +200,54 @@ service_registration {
...
+func TestE2E_ServiceRegistration_InitError(t *testing.T) {
+	setParallelism(t)
+	srv := newTestConsulServer(t)
+	defer srv.Stop()
+
+	tempDir := fmt.Sprintf("%s%s", tempDirPrefix, "registration_init_failure")
+	cleanup := testutils.MakeTempDir(t, tempDir)
+	defer cleanup()
+
+	// Use a nonexistent module
```
wilkermichael, line 233: The module exists, it's just made to fail right?

e2e/registration_test.go
mkam: That's a good point, will add comments to explain! I can't use `api.StartCTS()` because it sets the port, but I need to know what port to use beforehand so that I can configure `address`.

e2e/registration_test.go
mkam: It doesn't, this was an optimization since whether the task runs or not doesn't affect the test case.

e2e/registration_test.go
mkam: Oh, whoops, that was an old comment that I forgot about! I started with a nonexistent module, but realized that fails at the init stage and not at the plan/apply stage.

e2e/registration_test.go
mkam: Sure thing! I've got some other polls, I will add them there too.

retry/retry.go
mkam: Oh, I didn't notice that we had logged description later.  I didn't think `description` makes sense without the context of this method (as in, if you see this log and don't know about how we do retries, you don't know what this is a description of).

But `method` isn't really all that better when I think about it and this really should have been a separate PR anyway, so I'll change it to `description` and that can be some future evaluation.

testutils/consul.go
```diff
@@ -165,6 +166,31 @@ func ServiceRegistered(tb testing.TB, srv *testutil.TestServer, serviceID string
 	return resp.StatusCode == http.StatusOK
 }
 
+type Service struct {
+	ID      string
+	Service string
```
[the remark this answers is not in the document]
mkam: Discussed this offline, this is to match the response from Consul which uses `Service`.

e2e/registration_test.go
mkam: Added a comment with the same information!

wilkermichael (approved):
Thanks for addressing everything Melissa!

PR #861, opened 2022-05-10
Upgrade to terraform-exec v0.16.1

lornasong:
Upgrades `terraform-exec` and handles a breaking change in v0.16 ([changelog](https://github.com/hashicorp/terraform-exec/blob/main/CHANGELOG.md#0160-january-31-2022)) where `tfinstall` package was removed and requires importing the `hc-install` project as replacement.

Resources:
- [Relevant hc-install code](https://github.com/hashicorp/hc-install/blob/main/releases/exact_version.go#L20-L37)
- [tfexec code to replace tfinstall](https://github.com/hashicorp/terraform-exec/pull/235/commits/1dfc3aff0fb53c06949e7bef5968eadc78e7198c#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L38-R48)

### DIFF
--- a/driver/terraform_download.go
+++ b/driver/terraform_download.go
@@ -13,8 +13,9 @@ import (
 	ctsVersion "github.com/hashicorp/consul-terraform-sync/version"
 	"github.com/hashicorp/go-checkpoint"
 	goVersion "github.com/hashicorp/go-version"
+	"github.com/hashicorp/hc-install/product"
+	"github.com/hashicorp/hc-install/releases"
 	"github.com/hashicorp/terraform-exec/tfexec"
-	"github.com/hashicorp/terraform-exec/tfinstall"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


driver/terraform_download.go
```diff
@@ -189,7 +190,12 @@ func installTerraform(ctx context.Context, conf *config.TerraformConfig) (*goVer
 	// Create path if one doesn't already exist
 	os.MkdirAll(*conf.Path, os.ModePerm)
 
-	installedPath, err := tfinstall.ExactVersion(v.String(), *conf.Path).ExecPath(ctx)
+	installer := &releases.ExactVersion{
+		Product:    product.Terraform,
+		Version:    v,
+		InstallDir: *conf.Path,
+	}
+	installedPath, err := installer.Install(ctx)
```
mkam, line 198: Do we have any tests that check that Terraform is installed on the expected path? Probably isn't necessary if we don't, but was just wondering!

mkam (approved):
LGTM!

driver/terraform_download.go
lornasong: You are right! It looks like the benchmark and compatibility tests exercise the terraform download code and recently failed. I'll mention your comment on the fix for the benchmark and compatibility. It sounds like it would be good to have a test for this.

PR #872, opened 2022-05-13
Fixed issue with flags when running CTS without start command

### LINKED ISSUE #870: Command-line execution fails when utilizing space-delimited arguments
### Describe the bug

Command-line execution fails when utilizing space-delimited arguments rather than the `=` symbol in some circumstances. This occurs when excluding the `start` subcommand.

The demo application at the following documentation URL fails with this behavior:

https://learn.hashicorp.com/tutorials/consul/consul-terraform-sync-run-and-inspect?in=consul/network-infrastructure-automation

### Example Commands

Fails: `consul-terraform-sync -config-file cts-config.hcl -inspect`
Succeeds: `consul-terraform-sync -config-file=cts-config.hcl -inspect`
Succeeds: `consul-terraform-sync start -config-file cts-config.hcl -inspect`
Succeeds: `consul-terraform-sync start -config-file=cts-config.hcl -inspect`

### Versions
**Consul Terraform Sync**
```
consul-terraform-sync 0.6.0-beta1 (2d04123b)
Compatible with Terraform >= 0.13.0, < 1.2.0
```

wilkermichael:
# Background
When running CTS without a command, CLI space separate flag/value is not being interpreted correctly.

eg. `consul-terraform-sync -flag foo` `foo` is being interpreted as a command instead of as a value for `-flag`

When the the CLI library is invoke, it will attempt to find the command anywhere in the passed in args. That means that if no command is passed, it will interpret config.hcl as the command if `consul-terraform-sync -config-file config.hcl` is used to start the daemon.

We did not run into this issue previously as we used to only use the cli library if the first argument was not a flag (contained `-`). Now we use the CLI library for everything, and it will run the start command as the default command.

# Solution
Pre-process the args and tack on the default command (start). Not ideal, but also support for the command-less option will eventually go away (edited) 

Resolves #870

### DIFF
--- a/command/start.go
+++ b/command/start.go
@@ -28,6 +28,7 @@ const (
 	flagAutocompleteInstall   = "autocomplete-install"
 	flagAutocompleteUninstall = "autocomplete-uninstall"
 	flagClientType            = "client-type"
+	flagDeprecatedStartUp     = "deprecated-start-up"
 )
 
 // startCommand handles the `start` command
@@ -43,7 +44,11 @@ type startCommand struct {
 	autocompleteInstall   *bool
 	autocompleteUninstall *bool
 
-	isDefault bool
+	// isDeprecatedStartUp is set to true when 'start' is used because Consul-Terraform-Sync (CTS)
+	// CLI was not provided with a command, which is the deprecated way to run CTS as a Daemon.
+	// The flag is needed to change some behavior when CTS is started in this way,
+	// for example help messaging
+	isDeprecatedStartUp *bool
 
 	clientType *string
 }
@@ -52,7 +57,7 @@ func (c *startCommand) startFlags() *flag.FlagSet {
 	flags := flag.NewFlagSet(cmdStartName, flag.ContinueOnError)
 
 	var configFiles, inspectTasks config.FlagAppendSliceValue
-	var isInspect, isOnce, autocompleteInstall, autocompleteUninstall bool
+	var isInspect, isOnce, autocompleteInstall, autocompleteUninstall, isDeprecatedStartup bool
 	var clientType string
 
 	// Parse the flags
@@ -89,6 +94,14 @@ func (c *startCommand) startFlags() *flag.FlagSet {
 	flags.BoolVar(&autocompleteUninstall, flagAutocompleteUninstall, false, "Uninstall the autocomplete")
 	c.autocompleteUninstall = &autocompleteUninstall
 
+	// Automatically set by CTS when CTS is invoked without a command, and we are
+	// using `start` as a default. Not printed with -h, -help
+	flags.BoolVar(&isDeprecatedStartup, flagDeprecatedStartUp, false,
+		"Use only when running CTS without a command. "+
+			"\n\t\tSignals any different actions CTS should perform when "+
+			"\n\t\tthe start command is used as the default.")
+	c.isDeprecatedStartUp = &isDeprecatedStartup
+
 	// Development only flags. Not printed with -h, -help
 	flags.StringVar(&clientType, flagClientType, "",
 		"Use only when developing Consul-Terraform-Sync binary. "+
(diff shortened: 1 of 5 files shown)


wilkermichael:
@lornasong fixing the unit tests

wilkermichael:
@lornasong good to go

hashi-derek (approved):

command/start.go
```diff
@@ -89,6 +90,13 @@ func (c *startCommand) startFlags() *flag.FlagSet {
 	flags.BoolVar(&autocompleteUninstall, flagAutocompleteUninstall, false, "Uninstall the autocomplete")
 	c.autocompleteUninstall = &autocompleteUninstall
 
+	// Sets that we are using start as the default command. Not printed with -h, -help
+	flags.BoolVar(&isDefault, flagDefault, false,
+		"Use only when running CTS without the start command. "+
+			"\n\t\tSignals any different actions CTS should perform when "+
+			"\n\t\tthis command is used as the default.")
```
lornasong, line 97: From the flag description, I initially though this PR was adding a new flag for users to set (it's not, I think?). What are your thoughts on language something like, this flag is set automatically by CTS when starting up without the start command....

lornasong: I got a little flip-flopped understanding "default" - my intuition was the `start` command was the default (this is what we want users to do) but if I understand correctly, no `start` cmd is the default.

I was wondering what your thoughts are on renaming the flag and related variables to maybe `-deprecated` or `-deprecated-start-up`? I know this work will be removed in a few releases so want to be mindful of not taking up much of your time. I think having "deprecated" in the language will make it easy and clear what to clean up.

command/start.go
```diff
@@ -43,7 +44,7 @@ type startCommand struct {
 	autocompleteInstall   *bool
 	autocompleteUninstall *bool
 
-	isDefault bool
+	isDefault *bool
```
lornasong, line 47: I was initially wondering why a flag was used. If I understand correctly, a flag is used because it is the easiest (only?) way to pass the default information to the command.

Let me know if my understanding is incorrect. Either way, would it be possible to add a quick comment with this context?

lornasong (commented):
great catch! left a couple questions to check my understanding with potential follow up suggestions. let me know if it would be easier to chat in person

command/start.go
wilkermichael: yep, you're understanding correctly, this is the easiest (probably only?) way I could pass this information

command/start.go
wilkermichael: yeah, my reasoning for using `default` was because if you don't include a command, it's using the `start` command as default, and this signals to the command that it's using `start` as the default command, which changes some of the behaviour...

That might be unclear though, so I'm okay with deprecated-start-up?

lornasong: Ah interesting. Thanks for sharing your thought process. That makes sense. Yea, it looks like default can be interpreted different ways. `deprecated-start-up` works for me if it makes sense to you as well

command/start.go
lornasong: Gotcha, thanks for confirming

wilkermichael:
@lornasong ready for you to look at again

command/start.go
```diff
@@ -89,6 +94,13 @@ func (c *startCommand) startFlags() *flag.FlagSet {
 	flags.BoolVar(&autocompleteUninstall, flagAutocompleteUninstall, false, "Uninstall the autocomplete")
 	c.autocompleteUninstall = &autocompleteUninstall
 
+	// Sets that we are using start as the because CTS invoked without a command. Not printed with -h, -help
```
lornasong, line 97: Looks like there might be a typo in "using start as the because"

command/start.go
```diff
@@ -89,6 +94,13 @@ func (c *startCommand) startFlags() *flag.FlagSet {
...
+	flags.BoolVar(&isDeprecatedStartup, flagDeprecatedStartUp, false,
+		"Use only when running CTS without a command. "+
```
lornasong, line 99: In an earlier comment, mentioned how this reads as though is a flag that users pass in. Wasn't sure what your thoughts were on updating it to be "Automatically set by CTS only when running..."?

lornasong (approved):
thanks for the clarifications! makes sense to me. left a couple of non-blocking suggestion

Issue #873, opened 2022-05-16
Service deregistration errors when namespace configured


lornasong:
### Describe the bug
Issue occurs particularly when configuring CTS's Consul `service_registration` with namespace. When CTS gracefully shuts down, it deregisters itself from Consul. When `namespace` is configured, an error occurs:

> 2022-05-13T15:05:44.645-0400 [ERROR] registration: error deregistering Consul-Terraform-Sync from Consul: id=cts-0ebcb286-efcf-96e4-9a41-a7840c83be6e service_name=Consul-Terraform-Sync error="Unexpected response code: 404 (Unknown service ID "default/default/cts-0ebcb286-efcf-96e4-9a41-a7840c83be6e". Ensure that the service ID is passed, not the service name.)"

It looks like this issue occurs because CTS is being deregistered from the `default` namespace rather than the configured namespace.

### Versions
**Consul Terraform Sync**

consul-terraform-sync v0.6.0-beta1

**Consul**

consul enterprise (no specific version)

### Configuration File(s)

<details>
<summary>Click to toggle contents of config file </summary>

```hcl
log_level = "trace"

id = "cts-lorna"

consul {
 service_registration {
   namespace = "lornans"
 }
}

driver "terraform" {
  log = true
}

task {
 name = "a_task"
 version = "0.0.1"
 source = "lornasong/cts_file/local"
  condition "services" {
   names = ["api", "web"]
 }
}
```
</details>

### Expected Behavior
CTS successfully deregisters from Consul on CTS graceful shutdown.

### Expected Behavior
CTS logs an error when deregistering from Consul on CTS graceful shutdown. CTS remains  registered with Consul.

### Steps to Reproduce
1. Start Consul Enterprise: `consul agent -dev`
2. Add a namespace: `consul namespace create -name=lornans`
3. Start CTS with `consul.service_registration.namespace` configured
4. Wait until CTS finishes once-mode
5. Send sigint to CTS e.g. Ctrl+C
6. See error message in logs
7. In Consul UI, the CTS service registration can still be seen

### Additional context
When manually deregistering service, I encountered the same error as that logged in CTS:
`consul services deregister -id cts-0ebcb286-efcf-96e4-9a41-a7840c83be6e`

When adding a namespace flag, then I was able to successfully manually deregister the service
`consul services deregister -id cts-0ebcb286-efcf-96e4-9a41-a7840c83be6e -namespace=lornans`

PR #876, opened 2022-05-17
Fix issue with service deregistration using namespace

mkam:
Updates Consul API to the latest version (v1.12.0) to get the `ServiceDeregisterOpts` method that was introduced in v1.9.0. This allows us to set the namespace query parameter on a deregister service request.

Overview of Changes:
- Upgraded the Consul version
- Included query parameters when determining intercept path for `HttpIntercept`
- Updated registration manager to use namespace query parameter on deregistration
- Added compatibility test for checking CTS service registration successful
- Added v1.12.0 and updated 1.11.0-beta1-> 1.11.5 for Consul compatibility versions
- Removed the node check that was breaking with the new version of the client. `ExposedPort` was added to checks in [v1.10.0](https://github.com/hashicorp/consul/releases/tag/v1.10.0), and Catalog service registration requests were failing against versions older than 1.10.0 with `400 (Request decode failed: json: unknown field "ExposedPort")`. Looking at the test case, the node check wasn't being validated anyway and was grouped with the node name update.

Ran with compatibility test enabled on a separate branch, results can be seen [here](https://app.circleci.com/pipelines/github/hashicorp/consul-terraform-sync?branch=deregister-ns-compat&filter=all).

Fixes https://github.com/hashicorp/consul-terraform-sync/issues/873


client/consul_test.go
```diff
@@ -49,7 +49,7 @@ func Test_GetLicense_API_Failure(t *testing.T) {
 
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
-			path := "/v1/operator/license"
+			path := "/v1/operator/license?signed=1"
```
wilkermichael, line 52: can also set `signed=true` which might be more explicit?

testutils/utils.go
```diff
@@ -322,7 +322,12 @@ type HttpIntercept struct {
 
 func NewHttpClient(t *testing.T, intercepts []*HttpIntercept) *http.Client {
 	f := func(req *http.Request) *http.Response {
-		intercept, err := find(intercepts, req.URL.Path)
+		path := req.URL.Path
+		query := req.URL.RawQuery
```
wilkermichael, line 326: ah nice 👍

e2e/compatibility/consul_test.go
```diff
@@ -537,6 +577,24 @@ func registerCatalog(t *testing.T, entity *capi.CatalogRegistration, port int) {
...
+	conf.Address = fmt.Sprintf("localhost:%d", port)
+	client, err := capi.NewClient(conf)
+	require.NoError(t, err)
+
+	checks, err := client.Agent().Checks()
+	if err != nil {
+		return nil, err
+	}
+	c, ok := checks[checkID]
+	if !ok {
```
wilkermichael, line 579: I noticed that when using getCheck and getService, the caller requires no error. Should we just use the passed in *testing.T to require true?
```Go
require.True(t, ok)
```

wilkermichael (approved):

wilkermichael:
@mkam great work, just some minor comments

e2e/compatibility/consul_test.go
mkam: I was leaving it to the caller for the future if we need to verify the opposite, where the desired behavior is that a check or service doesn't exist.

But I probably should have done an error assertion right before this in the case where getting the agent checks failed! We should always expect to get a list of checks back.

client/consul_test.go
wilkermichael: ah, all good then, when Consul team sent me the `uri` they used `signed=true`

PR #877, opened 2022-05-17
Remove `RequestId` field from health checks.

hashi-derek:
The health-check endpoint was generating an unnecessary `RequestId` on each call. Request logs cannot be fetched by id for this endpoint, so the field did not provide value and was removed.


lornasong (approved):
makes sense to me! left a non-blocking comment that's more of a heads up

PR #878, opened 2022-05-17
Add e2e test for argument parsing variants.

hashi-derek:
Ensures that both `=` and space-delimited argument values are parsed correctly for subcommands and the "default" command.

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -123,7 +123,7 @@ func TestE2EArgumentParsing(t *testing.T) {
 		appendDBTask().appendWebTask()
 	config.write(t, configPath)
 
-	// Execute CTS with these extra parameters to ensure they parse correctly. For example:
+	// Execute CTS with these extra parameters to ensure they parse correctly.
 	testCases := [][]string{
 		{"start", "-config-dir", emptyDir, "-config-dir=" + emptyDir},
 		// TODO remove this line after the deprecated "default" implied subcommand is no longer supported.
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -107,14 +107,14 @@ func TestE2EArgumentParsing(t *testing.T) {
 	srv := newTestConsulServer(t)
 	defer srv.Stop()
 
-	tempDir := fmt.Sprintf("%s%s", tempDirPrefix, "cmd_parsing")
+	tempDir := fmt.Sprintf("%s%s", tempDirPrefix, "argument_parsing")
 	cleanupTemp := testutils.MakeTempDir(t, tempDir)
 	defer cleanupTemp()
 
 	// Create an extra, empty directory that will be searched for config files.
 	// This is done because the `StartCTS` function always includes its own
 	// argument of `-config-file=` and we cannot reference the same files twice.
-	emptyDir := fmt.Sprintf("%s%s", tempDirPrefix, "cmd_parsing_emptydir")
+	emptyDir := fmt.Sprintf("%s%s", tempDirPrefix, "argument_parsing_emptydir")
 	cleanupEmpty := testutils.MakeTempDir(t, emptyDir)
 	defer cleanupEmpty()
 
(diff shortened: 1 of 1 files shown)


e2e/e2e_test.go
```diff
@@ -99,6 +99,50 @@ func TestE2EBasic(t *testing.T) {
...
+	config.write(t, configPath)
+
+	// Execute CTS with these extra parameters to ensure they parse correctly:
+	// `-config-dir <empty-dir>`
+	// `-config-dir=<empty-dir>`
+	extraParams := []string{"-config-dir", emptyDir, "-config-dir=" + emptyDir}
+
+	// Execute the command with no subcommand
+	// TODO remove this after the "default" command is no longer supported.
+	cts, stop := api.StartCTS(t, configPath, extraParams...)
```
jwair, line 131: Is the goal of this test to validate the behavior of the CLI when the same arg `-config-dir` is set in two different ways? Or to test that either of these two ways (with and without `=`) work correctly?

e2e/e2e_test.go
```diff
@@ -99,6 +99,50 @@ func TestE2EBasic(t *testing.T) {
 	_ = cleanup()
 }
 
+// TestE2EArgumentParsing runs the CTS binary and checks to ensure that
+// argument parsing works properly for both `=` and space-delimited argument values.
+// For example, both `--a=b` and `--a b` should be acceptable ways to specify args.
+func TestE2EArgumentParsing(t *testing.T) {
+	srv := newTestConsulServer(t)
```
jwair, line 106: This test seems to be missing `setParallelism(t)` (first line)

e2e/e2e_test.go
```diff
@@ -99,6 +99,50 @@ func TestE2EBasic(t *testing.T) {
...
+
+	tempDir := fmt.Sprintf("%s%s", tempDirPrefix, "cmd_parsing")
+	cleanupTemp := testutils.MakeTempDir(t, tempDir)
+
+	// Create an extra, empty directory that will be searched for config files.
+	// This is done because the `StartCTS` function always includes its own
+	// argument of `-config-file=` and we cannot reference the same files twice.
+	emptyDir := fmt.Sprintf("%s%s", tempDirPrefix, "cmd_parsing_emptydir")
+	cleanupEmpty := testutils.MakeTempDir(t, emptyDir)
+	defer cleanupEmpty()
```
jwair, line 117: `cleanupTemp()` and `cleanupEmpty()` are called in two different ways, one is in a `defer` here, the other is in `_ = cleanupTemp()` at the end of the test. Is that intentional?

hashi-derek: Yeah, this was mainly a clone of the `TestE2EBasic` function. It has:
```
	cleanup := testutils.MakeTempDir(t, tempDir)
	// no defer to delete directory: only delete at end of test if no errors
```
for its cleanup approach, which allowed developers to debug by looking at the directory in the event that the configuration was bad. I was replicating that behavior, but since we're using the same config file as the basic test, I can likely just defer it and not worry about need to debug in that way.

e2e/e2e_test.go
```diff
@@ -99,6 +99,50 @@ func TestE2EBasic(t *testing.T) {
...
+// For example, both `--a=b` and `--a b` should be acceptable ways to specify args.
+func TestE2EArgumentParsing(t *testing.T) {
+	srv := newTestConsulServer(t)
+	defer srv.Stop()
+
+	tempDir := fmt.Sprintf("%s%s", tempDirPrefix, "cmd_parsing")
+	cleanupTemp := testutils.MakeTempDir(t, tempDir)
+
+	// Create an extra, empty directory that will be searched for config files.
+	// This is done because the `StartCTS` function always includes its own
```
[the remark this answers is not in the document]
wilkermichael: I'm leaving the above comment so that you're aware that `runSubcommand` exists, but I'm realising that it is a blocking call, and removes the ability to check if CTS started up correctly. Not a great fit for this particular scenario, so your workaround is the better option.

e2e/e2e_test.go
```diff
@@ -99,6 +99,50 @@ func TestE2EBasic(t *testing.T) {
...
+	// `-config-dir <empty-dir>`
+	// `-config-dir=<empty-dir>`
+	extraParams := []string{"-config-dir", emptyDir, "-config-dir=" + emptyDir}
+
+	// Execute the command with no subcommand
+	// TODO remove this after the "default" command is no longer supported.
+	cts, stop := api.StartCTS(t, configPath, extraParams...)
+	defer stop(t)
+	err := cts.WaitForTestReadiness(defaultWaitForTestReadiness)
+	require.NoError(t, err)
```
wilkermichael, line 136: we usually reserve `require` for errors that will make subsequent steps fail, i.e. an error decoding json

For these kinds of steps, use `assert` instead so that the test continues and we can catch other errors too

wilkermichael (approved):
@hashi-derek great work, some minor comments

PR #882, opened 2022-05-18
version not outputted anymore after fix running CTS without a command

wilkermichael:
## Summary
CTS does not behave as expected when attempting to get the version through the cli:

`consul-terraform-sync -version` 

outputs:
```linux
flag provided but not defined: -version
```

No variations of the version flag work correctly `-v`, `--version`, `--version`

This bug was introduced after fixing #870

## Investigation Results
To support pre-existing behavior in CTS where CTS can be run as a daemon without the `start` command, some preprocessing to arguments needed to be performed in order to make the CLI library believe a `start` command was used. This caused the CLI library to interpret `consul-terraform-sync -version` as `consul-terraform-sync start -version` which is invalid.

## Solution
Parse passed in args for `-v`, `--version` and `-version` and do not pre-process the args if present. Let the CLI library handle detection of the version after this.

### DIFF
--- a/command/cli.go
+++ b/command/cli.go
@@ -73,11 +73,14 @@ func NewCLI(out, err io.Writer) *CLI {
 // status from the command.
 func (cli *CLI) Run(args []string) int {
 	processedArgs := args[1:]
-	// If the command is not the first argument, assume no command was provided. Prepend the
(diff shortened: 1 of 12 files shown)


wilkermichael:
@mkam first few commits are to support unit testing

mkam (approved):

PR #886, opened 2022-05-19
move registration before once mode

### LINKED ISSUE #884: First service change may be ignored after restarting Consul
### Describe the bug
If Consul is stopped and started while CTS is still running, then sometimes the first monitored service change that happens after Consul has started again is ignored and the task does not trigger. The issue tends to happen more if there is a service registered before CTS is fully initialized.

This was tested with services and not confirmed if it is also happening with other changes like Consul KV.

### Versions
**Consul Terraform Sync**
Confirmed in v0.5.2 and v0.6.0-beta1. Confirmed _not_ in v0.4.3.

### Expected Behavior
After CTS reconnects with Consul, it should monitor changes and trigger tasks.

### Actual Behavior
The first service change is intermittently being ignored. Subsequent changes will cause the task to trigger. 

### Steps to Reproduce
Have not been able to reproduce manually, but was able to reproduce by modifying `TestE2ERestartConsul` to have a service registration done before CTS has initialized. See https://github.com/hashicorp/consul-terraform-sync/commit/dff2d615e3ce81a425e980ee5d4af17e6cb4d72f

### Additional Context
- This was first seen during development of CTS service registration. At that point, CTS was being registered before all tasks had run, and `TestE2ERestartConsul` was failing pretty consistently.
- I'm able to trigger the bug both by having a service registered before starting CTS and registering a service soon after starting CTS (presumably before CTS has fully initialized)
- Timing seems to be a big factor since I'm not able to recreate manually.
- When Consul restarts in the tests, the services are not persisted. This could also be a factor.

wilkermichael:
Registration moved before once mode runs

To accommodate this change, HCAT has been upgraded with a fix for #884 as placing the registration before once mode made this issue occur more frequently.

This PR resolves #884

### DIFF
--- a/e2e/registration_test.go
+++ b/e2e/registration_test.go
@@ -140,7 +140,7 @@ func TestE2E_ServiceRegistration_DeregisterWhenStopped(t *testing.T) {
 	config := baseConfig(tempDir).appendID(id).
 		appendConsulBlock(srv).appendTerraformBlock().
 		appendModuleTask("disabled_task", "mkam/hello/cts",
-			"enabled = false") // optimization since task running is not relevant to test
+			"enabled = true")
 	config.write(t, configPath)
 
 	// Start CTS, verify that service is registered
@@ -223,7 +223,7 @@ service_registration {
 }
 
 // TestE2E_ServiceRegistration_InitError tests that if the initial run
-// of all tasks fails, then CTS is not registered.
+// of all tasks fails, then CTS is still registered.
 func TestE2E_ServiceRegistration_InitError(t *testing.T) {
 	setParallelism(t)
 	srv := newTestConsulServer(t)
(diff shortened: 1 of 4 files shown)


e2e/registration_test.go
```diff
@@ -140,7 +140,7 @@ func TestE2E_ServiceRegistration_DeregisterWhenStopped(t *testing.T) {
 	config := baseConfig(tempDir).appendID(id).
 		appendConsulBlock(srv).appendTerraformBlock().
 		appendModuleTask("disabled_task", "mkam/hello/cts",
-			"enabled = false") // optimization since task running is not relevant to test
+			"enabled = true")
```
wilkermichael, line 143: This test used to work with false because registration was performed after once mode. The test waits for the registration to complete before trying to stop CTS. With registration after once, and all the tasks disabled, the `cts.WaitForTestReadiness(defaultWaitForTestReadiness)` function is not working as expected.

For now I am enabling this task so that the `WaitForTestReadiness` works as intended. Will have to revisit why this doesn't work with all disabled tasks later

mkam: I poked around because I was curious as to what's going on and here's my findings:

`WaitForTestReadiness` looks through all tasks and checks only enabled tasks. Since the only task is disabled, it will essentially return immediately. Because it's ignoring the disabled tasks, it doesn't wait for the task to finish initializing, which is preventing the graceful shutdown.

```
        2022-05-20T09:42:04.151-0500 [ERROR] driver.terraform: error initializing workspace for task: task_name=disabled_task
        2022-05-20T09:42:04.151-0500 [ERROR] ctrl: error initializing task: task_name=disabled_task
        2022-05-20T09:42:04.151-0500 [DEBUG] ctrl: cleaned up task that errored initializing: task_name=disabled_task
        2022-05-20T09:42:04.151-0500 [ERROR] cli: error running controller: error="error tf-init for 'disabled_task': context canceled"
```
This wasn't an issue when we registered tasks after once mode because of `WaitForConsulServiceRegistered`. This would wait for CTS to be registered, which would only happen after the task was fully initialized.

mkam: So having it set to true is the way to go since I don't think we can determine whether a task is initialized or not in the API. It was a very minor optimization anyway!

mkam (approved):

PR #893, opened 2022-05-20
Add configuration option for CTS service address

mkam:
Adds a new `address` configuration to the `consul.service_registration` block. Note that this configuration is purposefully  different than the `address` under `default_check`.

```
consul {
  service_registration {
    address = "123.456.789"
  }
}
```

### DIFF
--- a/config/service_registration.go
+++ b/config/service_registration.go
@@ -11,6 +11,7 @@ const (
 type ServiceRegistrationConfig struct {
 	Enabled      *bool               `mapstructure:"enabled"`
 	ServiceName  *string             `mapstructure:"service_name"`
+	Address      *string             `mapstructure:"address"`
 	Namespace    *string             `mapstructure:"namespace"`
 	DefaultCheck *DefaultCheckConfig `mapstructure:"default_check"`
 }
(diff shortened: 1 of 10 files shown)


config/service_registration.go
```diff
@@ -127,11 +138,13 @@ func (c *ServiceRegistrationConfig) GoString() string {
 	return fmt.Sprintf("&ServiceRegistrationConfig{"+
 		"Enabled:%v, "+
 		"ServiceName:%s, "+
+		"Address: %s, "+
```
wilkermichael, line 141: remove space between `:` and `%s` for consistency

registration/manager_test.go
```diff
@@ -48,6 +49,7 @@ func TestNewServiceRegistrationManager(t *testing.T) {
 				TLSEnabled: false,
 				ServiceRegistration: &config.ServiceRegistrationConfig{
 					ServiceName: config.String("cts-service"),
+					Address:     config.String("172.1.2.3"),
```
wilkermichael, line 52: since this is shared between the service and the service registration, consider making this string a variable? Might make it slightly easier to change in the future if we need to without missing one of them.

wilkermichael (approved):
Great work Melissa, you got this in so quickly! 

Just some really minor comments

PR #911, opened 2022-05-25
Add codeowners file for releng dirs

modrake:
This locks down changes to release directories so that they require approval from release engineering, and so they have service account rights for pipeline runs and backport-assistant functionality.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,4 +1,0 @@
-# release configuration
-
-/.release/                              @hashicorp/release-engineering @hashicorp/github-consul-ecosystem
-/.github/workflows/build.yml            @hashicorp/release-engineering @hashicorp/github-consul-ecosystem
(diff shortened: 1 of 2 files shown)


CODEOWNERS
```diff
@@ -0,0 +1,4 @@
+# release configuration
```
mkam, line 1: We've got an existing [CODEOWNERS file](https://github.com/hashicorp/consul-terraform-sync/blob/main/.github/CODEOWNERS) under the `.github` directory, could this be added to that file or does it need to be at the root level?

modrake: My apologies, I overlooked this subdir in a couple of repos - I've updated this to use the existing CODEOWNERS and will proceed to verify I haven't repeated the mistake in other consul ecosystem repos.

claire-labry (approved):

PR #921, opened 2022-06-01
Require indirect v3.0.1 gopkg.in/yaml.v3

lornasong:
Security scanner found a vulnerability in gopkg.in/yaml.v3
https://github.com/advisories/GHSA-hp87-p4gw-j4gq

 - go-yaml has released v3.0.1 to patch this issue
 - CTS uses github.com/deepmap/oapi-codegen@v1.10.1 which imports go-yaml
 - The latest version of oapi-codegen is still on go-yaml v3.0 and not yet
 upgraded to v3.0.1

This PR requires go-yaml with "// indirect" to require oapi-codegen to use a
later the specific version of go-yaml

Testing: used a [test branch](https://github.com/hashicorp/consul-terraform-sync/compare/main...lorna-go-yaml) and saw that [security scanner](https://github.com/hashicorp/consul-terraform-sync/runs/6676916628) passed

Preemptively added backport labels in case we need to release patch versions in the future


lornasong:
From investigation, we have decided not to do a security patch release for this issue. Our usage of go-yaml come from the oapigen dependency. We tested that using `0: [:!00 \xef` in the body of a request did not cause CTS to panic and exit.

Testing details:
- We tested using `0: [:!00 \xef` as the request body for Create Task API. CTS errored: ```{"error":{"message":"request body has an error: doesn't match the schema: Field must be set to object or not be present"},"request_id":"3eff7473-bcef-1085-9285-b93a7b71e4e9"}```
- We tested using `0: [:!00 \xef` as a description value for a task config for Create Task API. CTS errored: ```{"error":{"message":"request body has an error: failed to decode request body: invalid character 'x' in string escape code"},"request_id":"21222387-644e-ad9d-ff60-c3b75eff977c"}```

PR #932, opened 2022-06-04
GitHub Actions

jwair:
Introduced GitHub Actions for CI and weekly tests
Switched to v3 of setup-go and actions/checkout GH Actions
Bumped Consul versions to latest patch version in compatibility tests
Removed storing test results in CircleCI


.github/workflows/ci.yml
```diff
@@ -0,0 +1,186 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Install Consul
```
hashi-derek, line 76: Is there some way to abstract these installs to a step or bash script, since we're using it in multiple places?

hashi-derek (approved):

.github/workflows/ci.yml
jwair: Hi @hashi-derek, I switched to re-using actions that already exist and extracted out the other ones.

hashi-derek: @jwair That looks really nice

mdeggies (approved):

PR #933, opened 2022-06-06
Remove incorrect path lookup for terraform install.

hashi-derek:
The configuration provided to hc-install caused issues
whenever a version of terraform was located on the PATH
but not the application workdir. Since logic already
exists to check the workdir and version of the executable
prior to hc-install execution, the fs-check can be removed.

### DIFF
--- a/e2e/compatibility/terraform_download_test.go
+++ b/e2e/compatibility/terraform_download_test.go
@@ -1,0 +1,80 @@
+//go:build e2e
+// +build e2e
+
+package compatibility
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path"
+	"testing"
+	"time"
+
+	"github.com/hashicorp/consul-terraform-sync/config"
+	"github.com/hashicorp/consul-terraform-sync/driver"
+	"github.com/hashicorp/consul-terraform-sync/testutils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// This technically isn't an e2e test, but it downloads a file from the internet,
+// so it likely shouldn't be run frequently with other tests.
+func TestCompatibility_TFDownload(t *testing.T) {
+	tempDir := fmt.Sprintf("%s%s", tempDirPrefix, "Compatibility_TFDownload")
+	cleanup := testutils.MakeTempDir(t, tempDir)
+	defer cleanup()
+
+	conf := config.DefaultTerraformConfig()
+	conf.Path = &tempDir
+	// Use blank as a default value. Nil results in a panic.
+	tfVersion := ""
+	conf.Version = &tfVersion
+
+	binaryPath := path.Join(tempDir, "terraform")
+	info, err := os.Stat(binaryPath)
+	// We want to ensure the file does not exist and remove it if found.
+	if !os.IsNotExist(err) {
+		require.NoError(t, err, "An unexpected error occurred while checking for old terraform executable.")
+	}
+
+	if info != nil {
+		err = os.Remove(binaryPath)
+		require.NoError(t, err, "An unexpected error occurred while removing an old terraform executable.")
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
+	defer cancel()
+	err = driver.InstallTerraform(ctx, conf)
+	require.NoError(t, err, "Unable to download terraform executable.")
+
+	// Check to ensure the terraform binary exists.
+	info, err = os.Stat(binaryPath)
+	require.NoError(t, err, "Unable to find terraform executable that was expected to be installed.")
+
+	ensureTFVersionExecutes(t, binaryPath)
+
+	// Trigger install again. It shouldn't re-download the file.
+	firstInstallTime := info.ModTime()
+	ctx, cancel = context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	err = driver.InstallTerraform(ctx, conf)
+	require.NoError(t, err, "Unexpected error while checking for existing terraform install.")
+	info, err = os.Stat(binaryPath)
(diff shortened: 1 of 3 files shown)


jwair:
Is this going to cause issues for e2e tests? Because they all run at the same time and skip the TF download when they find TF in PATH. Even though they end up using TF that's in ./e2e (which we put in place as part of the test setup)

hashi-derek:
> Is this going to cause issues for e2e tests? Because they all run at the same time and skip the TF download when they find TF in PATH. Even though they end up using TF that's in ./e2e (which we put in place as part of the test setup)

I believe it will replicate the same behavior that existed prior to my hc-install changes.

The `InstallTerraform()` method does this sequence of checks first.
```
if isTFInstalled(path) { // Checks the workdir for the executable
    tfVersion, compatible, err := verifyInstalledTF(ctx, conf) // Checks the version of the executable
    ...
    return nil // Early return if a compatible tf was found in the workdir.
}
...
tfVersion, err := installTerraform(ctx, conf) // Download file to workdir.
```

If there are any race conditions with overlapping downloads, they should be the same behavior as before my hc-install change. The old tfinstall logic here doesn't have any PATH checks that I can tell. https://github.com/hashicorp/terraform-exec/blob/43ec6e17578bf2fd1e1037f9df5f1399d983eca8/tfinstall/exact_version.go#L34

jwair:
Makes sense, thanks @hashi-derek

mkam (approved):

e2e/compatibility/terraform_download_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/hashicorp/consul-terraform-sync/driver"
+	"github.com/hashicorp/consul-terraform-sync/testutils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// This technically isn't an e2e test, but it downloads a file from the internet,
+// so it likely shouldn't be run frequently with other tests.
+func TestCompatibility_TFDownload(t *testing.T) {
+
```
jwair, line 24:
```suggestion
```

e2e/compatibility/terraform_download_test.go
```diff
@@ -0,0 +1,65 @@
...
+	tfVersion := ""
+	conf.Version = &tfVersion
+
+	binaryPath := path.Join(tempDir, "terraform")
+	info, err := os.Stat(binaryPath)
+	// We want to ensure the file does not exist and remove it if found.
+	if !os.IsNotExist(err) {
+		require.NoError(t, err, "An unexpected error occurred while checking for old terraform executable.")
+	}
+	if info != nil {
```
jwair, line 41: Spacing out the two if blocks so the second one doesn't get confused with an `else` 
```suggestion

	if info != nil {
```

e2e/compatibility/terraform_download_test.go
```diff
@@ -0,0 +1,65 @@
...
+		err = os.Remove(binaryPath)
+		require.NoError(t, err, "An unexpected error occurred while removing an old terraform executable.")
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
jwair, line 45: Spacing out code block
```suggestion

	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```

e2e/compatibility/terraform_download_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	// Check to ensure the terraform binary exists.
+	info, err = os.Stat(binaryPath)
+	require.NoError(t, err, "Unable to find terraform executable that was expected to be installed.")
+	// Check for an arbitrary 1MB+ terraform executable size.
+	assert.Greater(t, info.Size(), int64(1_000_000), "The terraform file size was unexpectedly small. This is likely an issue with installing.")
+
+	// Trigger install again. It shouldn't re-download the file.
+	firstInstallTime := info.ModTime()
+	ctx, cancel = context.WithTimeout(context.Background(), 3*time.Second)
```
jwair, line 58: Is this too aggressive of a timeout? in case the CPU is busy for instance

hashi-derek: I suppose it could be. It's hitting the local disk, so I figured 3s was enough. I can bump it to something higher if you want, though.

e2e/compatibility/terraform_download_test.go
```diff
@@ -0,0 +1,66 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	err = driver.InstallTerraform(ctx, conf)
+	require.NoError(t, err, "Unable to download terraform executable.")
+
+	// Check to ensure the terraform binary exists.
+	info, err = os.Stat(binaryPath)
+	require.NoError(t, err, "Unable to find terraform executable that was expected to be installed.")
+	// Check for an arbitrary 1MB+ terraform executable size.
+	assert.Greater(t, info.Size(), int64(1_000_000), "The terraform file size was unexpectedly small. This is likely an issue with installing.")
```
jwair, line 55: This check is neat but it is a bit indirect, is it better to test that `terraform version` runs successfully?

e2e/compatibility/terraform_download_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	// Trigger install again. It shouldn't re-download the file.
+	firstInstallTime := info.ModTime()
+	ctx, cancel = context.WithTimeout(context.Background(), 3*time.Second)
```
jwair, line 59: Is this timeout a bit aggressive for installing TF?

e2e/compatibility/terraform_download_test.go
hashi-derek: Yeah, that's fair. Do you think looking for the success status code is sufficient, or should I be looking for a version string from stdout?

jwair (approved):
LGTM

PR #941, opened 2022-06-07
Updated the default service name to lower case

noejbrown:
Service name is used in service registration it was previously proper case. Changed to lower case because of feedback; service names should never be capitalized in Consul so that they are valid hostnames.


mkam (approved):
Looks good to me!

PR #945, opened 2022-06-08
Update CHANGELOG.md

noejbrown:
Updated changelog to reflect default service name change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 IMPROVEMENTS:
 * Add `openssh` command to Docker image to support git over ssh for Terraform modules [[GH-940](https://github.com/hashicorp/consul-terraform-sync/issues/940)]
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 IMPROVEMENTS:
 * Add `openssh` command to Docker image to support git over ssh for Terraform modules [[GH-940](https://github.com/hashicorp/consul-terraform-sync/issues/940)]
+* Changed default service name in service registration configuration from `Consul-Terraform-Sync` to `consul-terraform-sync` (https://github.com/hashicorp/tutorials/pull/402#discussion_r883929477)
```
mkam, line 5: Since changing a default value changes the behavior of a user's configuration if they upgrade CTS while using the default value, can you create a new for `BREAKING CHANGES:` and put it above the improvements section? We like to call out breaking changes first since we want to make sure users see them.

Also, I made a GitHub issue that you can link (https://github.com/hashicorp/consul-terraform-sync/issues/946)! You can use the format:

```
[[GH-946](https://github.com/hashicorp/consul-terraform-sync/issues/946)]
```

noejbrown: Thank you will change this!

PR #947, opened 2022-06-09
Parameterize task name for enterprise.

hashi-derek:
This helper method currently uses the same task name for both enterprise and oss testing. Changing the task name to a parameter helps prevent concurrent test execution from failing unexpectedly.


mkam (approved):
Thanks for taking this on!

PR #954, opened 2022-06-10
Fix e2e test that fails due to a race condition.

hashi-derek:
The logic in this test assumed that task deletion was
a synchronous process and could fail to see the task
disappear from a listing after a delete call.


api/task_delete.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/hashicorp/consul-terraform-sync/logging"
 )
 
-// DeleteTaskByName deletes an existing task and its events. Does not delete
+// DeleteTaskByName deletes an existing task and its events asynchronously. Does not delete
 // if the task is active.
```
mkam, line 11:
```suggestion
// until the task is inactive and not running.
```
This line is a little out of date, I forgot to update it when it was changed from error if active to waiting until it was inactive.

controller/tasks_manager.go
```diff
@@ -122,11 +122,11 @@ func (tm *TasksManager) TaskCreateAndRun(ctx context.Context, taskConfig config.
 func (tm *TasksManager) TaskDelete(ctx context.Context, name string) error {
 	logger := tm.logger.With(taskNameLogKey, name)
 	if tm.drivers.IsMarkedForDeletion(name) {
-		logger.Debug("task is already marked for deletion")
+		logger.Debug("task is already marked for deletion", "taskName", name)
```
mkam, line 125: This actually has the task name as part of the logging already, as set on line 123.

```
tm.logger.With(taskNameLogKey, name)
```

```
2022-06-13T09:12:21.691-0500 [DEBUG] tasksmanager: task marked for deletion: task_name=test-task
2022-06-13T09:12:21.691-0500 [DEBUG] tasksmanager: task deleted: task_name=test-task
```

controller/tasks_manager.go
```diff
@@ -122,11 +122,11 @@ func (tm *TasksManager) TaskCreateAndRun(ctx context.Context, taskConfig config.
...
 		return nil
 	}
 	tm.drivers.MarkForDeletion(name)
-	logger.Debug("task marked for deletion")
+	logger.Debug("task marked for deletion", "taskName", name)
```
mkam, line 129: Same here, this should already be logging the task name.

mkam (approved):
Left a note on the changes to logging, but the test changes looks good to me!

PR #960, opened 2022-06-13
Simpler No-Duplicate CI Runs in GH Actions

jwair:
Instead of using a custom GH action to skip duplicate builds, we are relying on native GH config to only run the CI workflow on push (any branch) and PR open/reopen to main.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,7 +159,7 @@ jobs:
           test-results: .build/test-results.json
 
   slack-notification:
-    name: Slack notification
+    name: Slack Notification
     if: ${{ failure() && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release')) }}
     needs:
       - checks
--- a/.github/workflows/weekly-tests.yml
+++ b/.github/workflows/weekly-tests.yml
@@ -17,7 +17,7 @@ env:
 
 jobs:
   compatability:
-    name: Compatability tests
+    name: Compatability Tests
     runs-on: ubuntu-latest
     timeout-minutes: 30
     steps:
@@ -33,7 +33,7 @@ jobs:
           make test-compat
 
   vault-integration:
-    name: Vault integration tests
+    name: Vault Integration Tests
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
@@ -53,7 +53,7 @@ jobs:
           make test-vault-integration
 
   benchmarks:
-    name: Benchmarks tests
+    name: Benchmarks Tests
     runs-on: ubuntu-latest
     timeout-minutes: 150
     steps:
@@ -97,7 +97,7 @@ jobs:
           path: .build/benchmarks-results*.csv
 
   slack-notification:
-    name: Slack notification
+    name: Slack Notification
     if: ${{ failure() }}
     needs:
       - compatability


.github/workflows/ci.yml
```diff
@@ -4,6 +4,9 @@ on:
   workflow_dispatch:
   push:
   pull_request:
+    types:
+      - opened
+      - reopened
```
mkam, line 9: If I pushed to a branch, waited for the build to finish, and then opened a PR, would it run the build again?

.github/workflows/ci.yml
```diff
@@ -4,6 +4,9 @@ on:
...
     branches:
       - main
```
lornasong, line 11: I think this `pull_request` block is saying, run GHA when opening/reopening PRs against main? Is there a separate flow for PRs against the release branches? I might be misunderstanding

lornasong (commented):
Just want to make sure I understand - the problem that was solved (and is being simplified now) is when a new commit is pushed to an open PR, GHA would run twice (once for the open PR and another for the push). Is that understanding right?

.github/workflows/ci.yml
jwair: > Just want to make sure I understand - the problem that was solved (and is being simplified now) is when a new commit is pushed to an open PR, GHA would run twice (once for the open PR and another for the push). Is that understanding right?

Yes, that is correct.

jwair:
> Just want to make sure I understand - the problem that was solved (and is being simplified now) is when a new commit is pushed to an open PR, GHA would run twice (once for the open PR and another for the push). Is that understanding right?

Yes, that is correct.

.github/workflows/ci.yml
jwair: No there isn't a separate flow. Pinning `main` is to ensure we trigger a CI run when there's a PR from a forked repo.
For release branches, we don't create a PR, we create a new branch and might push to it. When we push to it, a PR will be triggered.
Should we add a release branch pattern for pull_request triggers? (in addition to main) @lornasong That would help IF we open PRs to release branches.

.github/workflows/ci.yml
mkam: I'd prefer the confusing approach, to be honest, because my normal workflow is that I push to a branch and only when the branch is passing do I open a pull request. There's always going to be an initial push to a branch that triggers a build (excluding the scenario where the PR is from a fork), so we're always going to have a duplicate run.

lornasong: Just sharing my workflow, I similarly create a PR after I see that my branch is passing.

jwair: Just to clarify, CI will run when you push to your branch (with or without a PR). This optimization is to improve what happens after we open a PR.
So the only duplicate CI run will be once when we first open a PR. After that, CI will trigger for any pushes to your branch per usual.

.github/workflows/ci.yml
jwair: Addressed in https://github.com/hashicorp/consul-terraform-sync/pull/960/commits/0adbf39fb055f958271851d9ae063d990050e782

.github/workflows/ci.yml
mkam: I guess what I'm saying is I don't want that duplicate CI run to happen when you open the PR.

 Because our tests have issues with flakiness, my general approach is to push the branch, make sure everything is passing so that I know it doesn't need either further changes because of a bug or retries because of intermittent failures, and then open the PR. Then at that point I can be done with this PR until it's reviewed and move on to another task.

If it runs again after the PR is opened, I'll have to check back again and make sure everything is still passing.

lornasong: Would it make sense to remove the `pull_request` block? I think with CircleCI, we typically ran tests only on push. I don't recall if any tests were run on pull requests opening / reopening. Would this solve the duplicate test runs? Or are there new actions that specifically need to be run on pull request?

.github/workflows/ci.yml
lornasong: Thanks for addressing even though it's outside the original PR change

.github/workflows/ci.yml
jwair: 👍 @mkam, understood.
@lornasong, yes it's possible but we need CI to run when we get a PR from a forked repo. In that case, without the PR block, we won't have any PR checks.

lornasong: Thanks for that context. To confirm, it sounds like the only way to run tests on PRs from forked repos is to have a duplicate run when opening a PR against the original repo?

jwair: @lornasong I just had a discussion with Melissa, sharing here for context. I have removed the pull_request trigger, this covers the majority of cases, CAT team creating a PR from a branch. I will test after merging this what happens when a PR is made from a forked repo and adjust as needed.

jwair: > Thanks for that context. To confirm, it sounds like the only way to run tests on PRs from forked repos is to have a duplicate run when opening a PR against the original repo?

No, there's a way to only trigger on PRs from forked repos but it requires adding an `if` block to our workflow jobs. I wanted to see if we have a simpler solution first.

PR #962, opened 2022-06-14
Simpler ci gh action

jwair:


.github/workflows/ci.yml
```diff
@@ -160,8 +197,12 @@ jobs:
 
   slack-notification:
     name: Slack Notification
-    if: ${{ failure() && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release')) }}
+    if: |
+      needs.skip-check.result == 'success' &&
```
lornasong, line 201: nit: the `skip-check.result == 'success'` intuitively reads to me "we should skip this job". so the if-statement here initially reads as something like "if the job was successfully skipped and there was a failure and the branches match, then send notification in slack" (which I think is not the intention).

suggestion to reframe the check as when to run a job vs. when to skip a job and maybe rename `skip-check` to something like `check-run-job`, `should-run-job`, `should-run`

I might be misunderstanding this all together, please don't hesitate to let me know

.github/workflows/ci.yml
```diff
@@ -14,9 +18,25 @@ env:
   TERRAFORM_VERSION: 1.2.2
 
 jobs:
+  skip-check:
+    name: Skip Check
+    runs-on: ubuntu-latest
+    # This job only runs if the action trigger is manual, a commit push, or a PR from a different repo (fork).
+    # Other jobs can check if this job ran successfully to determine if they should run as well.
```
lornasong, line 25: helpful comment!

lornasong (approved):
makes sense to me. glad that prs from forks will be able to run tests! left a non-blocking naming suggestion

.github/workflows/ci.yml
jwair: Great suggestion, I'll update the name

jwair: Switched the name to `run-check` in https://github.com/hashicorp/consul-terraform-sync/pull/962/commits/1bf2375d211729674195bdc1355a7dd205fd171c

PR #967, opened 2022-06-22
Various state store improvements

hashi-derek:
In order to better support different state stores in the future, error returns needed to be introduced into the store interface when saving data.


wilkermichael (approved):
@hashi-derek looks good to me!

Issue #849, opened 2022-05-10
Terraform CLI requests cause leaked go-routines


lornasong:
### Describe the bug
Root issue: Currently CTS usage of Terraform CLI causes a leaked go-routine each time the CLI is requested.

Symptoms: When Terraform logs are enabled, users can observe `error from kill: os: process already finished"`in the logs (more details / samples below)

Impact:
 - Only Terraform driver (does not impact TFC driver)
 - Increased resource usage as more Terraform requests are made by CTS daemon

Workarounds: no known workarounds

Prior to 0.6, users can see the "error from kill" log when CTS stops gracefully:
```
^C2022-05-02T15:39:43.469-0400 [INFO]  cli: signal received to initiate graceful shutdown: signal=interrupt
2022/05/02 15:39:43 error from kill: os: process already finished
2022/05/02 15:39:43 error from kill: os: process already finished
2022-05-02T15:39:43.469-0400 [INFO]  ctrl: stopping dependency monitoring
2022-05-02T15:39:43.469-0400 [INFO]  ctrl: stopping controller
2022/05/02 15:39:43 error from kill: os: process already finished
2022/05/02 15:39:43 error from kill: os: process already finished
2022/05/02 15:39:43 error from kill: os: process already finished
2022/05/02 15:39:43 error from kill: os: process already finished
2022-05-02T15:39:43.469-0400 [INFO]  api: shutdown api server
2022-05-02T15:39:43.469-0400 [INFO]  cli: graceful shutdown
```

With 0.6 beta, users can _additionally_ see the "error from kill" log  after once-mode is completed due to some refactoring
```
2022-05-10T11:55:45.304-0400 [INFO]  ctrl: task completed: task_name=ls_testing_services
2022-05-10T11:55:45.304-0400 [INFO]  ctrl: all tasks completed once
2022/05/10 11:55:45 error from kill: os: process already finished
2022/05/10 11:55:45 error from kill: os: process already finished
2022/05/10 11:55:45 error from kill: os: process already finished
2022/05/10 11:55:45 error from kill: os: process already finished
2022/05/10 11:55:45 error from kill: os: process already finished
2022/05/10 11:55:45 error from kill: os: process already finished
2022-05-10T11:55:45.307-0400 [INFO]  registration: registering Consul-Terraform-Sync as a service with Consul: id=cts-a3807435-f398-534c-cb9e-a6c1551da9fc service_name=Consul-Terraform-Sync
```
### Versions
[...]

PR #969, opened 2022-06-23
Upgrade to terraform-exec v0.17

lornasong:
Fixes: https://github.com/hashicorp/consul-terraform-sync/issues/849

Commits:
 - Adds a test to show issue with leaked goroutines (test fails)
 - Pulls latest tf-exec version (test passes)


mkam (approved):
Looks good to me! Thanks for putting in all the effort to fix this in terrafrom-exec!

PR #973, opened 2022-06-27
Support once-mode failing silently

lornasong:
We want to be able to support when once-mode does not exit on error. One use-case for this is when the tasks have been known to run successfully before. When CTS starts up, we expect the tasks to be successful and any errors are intermittent. In this case, we want CTS to not exit and just log the error.

Commits
 - 1 & 2: Split out error test case for `Test_Once_Run_Terraform` in preparation for new error case in commit 4
 - 3: Add TaskFailsSilently (open to better naming ideas) in tasks manager
 - 4: Update once-mode with a failSilently flag that calls TaskFailsSilently when true

Edit: from PR feedback changed "fail silently" language to "allow fail"

### DIFF
--- a/controller/tasks_manager.go
+++ b/controller/tasks_manager.go
@@ -215,13 +215,14 @@ func (tm *TasksManager) TaskUpdate(ctx context.Context, updateConf config.TaskCo
 	return plan.ChangesPresent, plan.Plan, "", nil
 }
 
-// TaskFailSilently creates, runs, and adds a new task. It expects that this
-// task is highly unlikely to error because it has previously been created and
-// run before. Therefore it does not handle the error beyond logging.
+// TaskCreateAndRunAllowFail creates, runs, and adds a new task. It expects that
+// this task is highly unlikely to error because it has previously been created
+// and run before. Therefore it allows failure and does not handle error beyond
+// logging
 //
 // This method is used when we do not want the caller to error and exit when
 // creating, running, and adding a new task.
-func (tm *TasksManager) TaskFailSilently(ctx context.Context, taskConfig config.TaskConfig) {
+func (tm *TasksManager) TaskCreateAndRunAllowFail(ctx context.Context, taskConfig config.TaskConfig) {
 	logger := tm.logger.With(taskNameLogKey, *taskConfig.Name)
 
 	actionSteps := `
--- a/controller/tasks_manager.go
+++ b/controller/tasks_manager.go
@@ -110,7 +110,7 @@ func (tm *TasksManager) TaskCreateAndRun(ctx context.Context, taskConfig config.
 		return config.TaskConfig{}, err
 	}
 
-	if err := tm.runNewTask(ctx, d); err != nil {
+	if err := tm.runNewTask(ctx, d, false); err != nil {
 		return config.TaskConfig{}, err
 	}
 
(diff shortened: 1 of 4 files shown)


controller/tasks_manager.go
```diff
@@ -215,6 +215,42 @@ func (tm *TasksManager) TaskUpdate(ctx context.Context, updateConf config.TaskCo
 	return plan.ChangesPresent, plan.Plan, "", nil
 }
 
+// TaskFailSilently creates, runs, and adds a new task. It expects that this
+// task is highly unlikely to error because it has previously been created and
+// run before. Therefore it does not handle the error beyond logging.
+//
+// This method is used when we do not want the caller to error and exit when
+// creating, running, and adding a new task.
+func (tm *TasksManager) TaskFailSilently(ctx context.Context, taskConfig config.TaskConfig) {
```
mkam, line 224: Was thinking that this would benefit from adding `Create` in the name, which then led me to think that it's technically `TaskCreateAndRunFailSilently` if we want parity with the other methods. So then on this note, wanted to get your thoughts on updating `TaskCreateAndRun` to have a boolean for whether errors are ignored or not instead of a separate method.

Also this is like a big nit, but if we're logging the errors, is it technically failing silently? 🤔 I think it gets the point across either way, so not really advocating a change!

controller/tasks_manager_test.go
```diff
@@ -850,6 +850,51 @@ func Test_ConditionMonitor_EnableTaskRanNotify(t *testing.T) {
...
+			return mockD, nil
+		}
+
+		tm.TaskFailSilently(ctx, validTaskConf)
+
+		_, ok := tm.drivers.Get(validTaskName)
+		assert.True(t, ok, "driver is added even if run is unsuccessful")
+
+		events := tm.state.GetTaskEvents(validTaskName)
+		assert.Len(t, events, 1, "event is stored even on failed run")
```
mkam, line 894: Could you also assert that the event has a failure status?

mkam (approved):
Looks good to me! Left some thoughts on the name, happy to discuss it more if needed!

controller/tasks_manager.go
lornasong: Thanks for the feedback!

I like having one methods with a boolean because it makes naming easier. I went with separate methods to reduce complexity within the method, to be more explicit about the two behaviors to avoid any accidents, to avoid having the API have access to the fail-silently option. 🤔let me try out combining again!

I get what you mean that "fail-silently" is not intuitive because the logs aren't very silent 😆 . From looking around initially, fail-silently seemed like it referred to only when errors are aren't returned/thrown (they may or may not be logged - logging seemed like a grey zone) ([ex](https://enterprisecraftsmanship.com/posts/fail-fast-principle/)). Since it's not intuitive, maybe "continue on failure" or "allow failure"?

lornasong: Hello! I tried out some iterations to combine the methods ([gist](https://gist.github.com/lornasong/247a0d96803c06bb8447ee183faa9ad1)). From that, I'm thinking of 2 options:
1. One combined method: iteration 2 in gist and use "allowFail" instead of "failSilently"
2. Two separate methods: with "allowFail" e.g. TaskCreateAndRunAllowFail

I think I'm leaning towards Op 1. Please let me know if you have any thoughts!

mkam: Yeah, I think iteration 1 is what I was thinking of originally, but it's more difficult to read than anticipated! I'd be good with either of these two options! I might be feeling more the separate methods now since iteration 2 of the combined method doesn't really reduce code duplication, which I thought could have been the advantage of a combined method.

lornasong: Thanks for the feedback! I went with separate methods (op 2) and "allow fail" type of naming. Agree with your point about reducing code duplication is the advantage of the combined method

mkam: Thanks for entertaining my original idea! Changes looks good to me!

PR #974, opened 2022-06-28
Notifier improvements

hashi-derek:
Improves codebase by reducing the number of notifier implementations and removing unnecessary template override logic.

Fixes race condition bug in notifiers where they could indicate once-mode is completed too early. The old logic relied upon the number of calls encountered to determine whether once-mode had finished. This approach was fundamentally flawed and always prone to race-conditions since it's possible to receive events from any input source multiple times before once-mode is done.

Most notably, the catalog services inputs triggered this bug frequently, because CTS was self-registering with the catalog services in a separate goroutine.


mkam (approved):
Really like the reduction of all the boilerplate code that had to go into making a notifier! Also glad to the override notifier removed too.

PR #976, opened 2022-06-28
upgraded go minimum version to 1.18

wilkermichael:
The pipeline now builds CTS with version 1.18, so upgraded the go mod to reflect this as the minimum supported version


lornasong (approved):
nice catch!

PR #977, opened 2022-06-29
Run Full E2E Matrix (CI)

jwair:
This change ensures that when a single part of e2e tests fails, the other parts continue executing and are not skipped.

Tested with an intentional failure:
image


lornasong (approved):
thanks for adding this + the commit and revert to demo that fail fast is off!

PR #983, opened 2022-07-06
Remove task merge behavior from store.

hashi-derek:


state/in_memory_store.go
```diff
@@ -95,15 +94,15 @@ func (s *InMemoryStore) SetTask(newTaskConf config.TaskConfig) error {
...
+			// overwrite the existing task
+			updatedTaskConf := newTaskConf.Copy()
 			(*taskConfs)[ix] = updatedTaskConf
 			return nil
 		}
 	}
 
 	// add as a new task
-	*taskConfs = append(*taskConfs, &newTaskConf)
+	*taskConfs = append(*taskConfs, newTaskConf.Copy())
```
wilkermichael, line 105: Doesn't this mean the taskConfs will grow each time, even if overwriting?

wilkermichael (approved):
looks good to me!

state/in_memory_store.go
```diff
@@ -95,15 +94,15 @@ func (s *InMemoryStore) SetTask(newTaskConf config.TaskConfig) error {
 
 	for ix, taskConf := range *taskConfs {
 		if config.StringVal(taskConf.Name) == newTaskName {
-			// patch update the existing task
-			updatedTaskConf := taskConf.Merge(&newTaskConf)
+			// overwrite the existing task
```
wilkermichael, line 97: I would just add a debug log that this is what happened, only minor suggestion

hashi-derek: Since the rest of the in-memory store doesn't use logs, I'm not sure we should introduce it. It's relatively simple behavior that's covered by unit tests, so I'm not sure that we need to be that verbose about it.

PR #986, opened 2022-07-07
Allow cts run without initial task

wilkermichael:
This PR serves to remove the requirement that CTS be run with initial configuration tasks.

When CTS is run without initial configuration tasks, it will forgo running once mode and monitor for new tasks/serve the API. If a new task is created via the API/CLI, then CTS will monitor and trigger as before.

If CTS is run with initial configuration tasks, runs as usual.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -439,6 +439,12 @@ func TestConfig_Validate(t *testing.T) {
 	*validMultiTask.TerraformProviders = append(*validMultiTask.TerraformProviders,
 		&TerraformProviderConfig{"Y": map[string]interface{}{}})
 
+	validNilTasks := longConfig.Copy()
+	*validNilTasks.Tasks = nil
+
+	validEmptyTasks := longConfig.Copy()
+	*validEmptyTasks.Tasks = TaskConfigs{}
+
 	cases := []struct {
 		name    string
 		i       *Config
@@ -459,6 +465,14 @@ func TestConfig_Validate(t *testing.T) {
 		}, {
 			"multi-task valid",
 			validMultiTask.Copy(),
+			true,
+		}, {
+			"nil tasks valid",
+			validNilTasks.Copy(),
+			true,
+		}, {
+			"empty tasks valid",
+			validEmptyTasks.Copy(),
 			true,
 		}, {
 			"empty provider",
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -99,6 +99,36 @@ func TestE2EBasic(t *testing.T) {
 	_ = cleanup()
 }
 
+// TestE2ENoTasks runs the CTS binary in daemon mode with a configuration with no
+// tasks
+// 1. Verifies that CTS starts without issue
+// 2. Verifies that a CTS can handle task lifecycle operations
+func TestE2ENoTasks(t *testing.T) {
+	setParallelism(t)
+
+	srv := newTestConsulServer(t)
+	defer srv.Stop()
+
+	tempDir := fmt.Sprintf("%s%s", tempDirPrefix, "no_tasks")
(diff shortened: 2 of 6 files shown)


config/config.go
```diff
@@ -283,8 +283,12 @@ func (c *Config) Validate() error {
 		return err
 	}
 
-	if err := c.Tasks.Validate(); err != nil {
-		return err
+	// Acceptable for tasks to be empty, only check if
+	// validity they exist
+	if c.Tasks != nil && len(*c.Tasks) != 0 {
```
wilkermichael, line 288: Will fix this comment

config/config.go
```diff
@@ -283,8 +283,12 @@ func (c *Config) Validate() error {
 		return err
 	}
 
-	if err := c.Tasks.Validate(); err != nil {
-		return err
+	// Acceptable for tasks to be empty, only check if
+	// validity they exist
```
wilkermichael, line 287: Will fix this comment for grammar

config/config_test.go
```diff
@@ -439,6 +439,12 @@ func TestConfig_Validate(t *testing.T) {
 	*validMultiTask.TerraformProviders = append(*validMultiTask.TerraformProviders,
 		&TerraformProviderConfig{"Y": map[string]interface{}{}})
 
+	validNilTasks := longConfig.Copy()
+	*validMultiTask.Tasks = nil
```
lornasong, line 443: I don't quite follow `*validMultiTask.Tasks = nil`. Maybe should "validMultiTask" be "validNilTasks"?

config/config_test.go
```diff
@@ -439,6 +439,12 @@ func TestConfig_Validate(t *testing.T) {
...
+
+	validEmptyTasks := longConfig.Copy()
+	*validMultiTask.Tasks = TaskConfigs{}
```
lornasong, line 446: Similar here, maybe "validMultiTask" should be "validEmptyTasks"?

e2e/e2e_test.go
```diff
@@ -99,6 +99,31 @@ func TestE2EBasic(t *testing.T) {
 	_ = cleanup()
 }
 
+// TestE2ENoTasks runs the CTS binary in daemon mode with a configuration with no
+// tasks
+func TestE2ENoTasks(t *testing.T) {
```
lornasong, line 104: This is some super thorough testing with all the e2e test updates. I noticed that TestE2E_CreateTaskCommand and TestE2E_CreateDeleteCreateTrigger were also updated. For me, the no-task case is a detail-related to the command tests but not one of its core test case. But let me know your thought process.

One option could be to keep the command-related tests as they were and update the TestE2ENoTasks to include command testing. I'm wondering if it could be as simple as having TestE2ENoTasks call runCreateDeleteCreateTrigger()? I'm assuming it covers the test case in TestE2E_CreateTaskCommand but that might be incorrect.

We may have different thoughts / preferences. Let me know what you think!

lornasong (commented):
left a couple of questions - let me know if it would be easier to chat in person!

config/config.go
```diff
@@ -283,8 +283,12 @@ func (c *Config) Validate() error {
 		return err
 	}
 
-	if err := c.Tasks.Validate(); err != nil {
```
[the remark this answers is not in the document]
wilkermichael: That is what I originally did, but then thought it was weird that a Tasks validation call succeeded if the list provided was nil?

I can change it back

config/config_test.go
wilkermichael: yes you're right, copy paste error

e2e/e2e_test.go
wilkermichael: yeah, I can do that, then remove that level of testing from the commands

lornasong (approved):
looks good, thanks for the updates!

PR #987, opened 2022-07-08
Intentions module input configuration 

noejbrown:
Change:  added new  configuration for `module_input` `"intentions"` and updated necessary module input dependencies 

Commits :
- Added intentions monitor configuration and test
- Added intentions services configuration for `source_services` and `destination_services` used in the `"intentions"` block
- Added intentions module input configuration and test
- Updated module_input.go and module_input_test.go to include the new  `module_input` `"intentions"`

##### Sample `"intentions" module_input`

```
task {
	name = "module_input_task"
	module = "..."
	condition "schedule" {
		cron = "* * * * * * *"
	}
	module_input "intentions" {
		datacenter = "datacenter"
		namespace = "namespace"
		source_services {
			regexp = "^web.*"
		}
		destination_services {
			regexp = "^api.*"
		}
	}
}`
```


noejbrown:
Did a rebase, this change does not require release engineers to review anymore. Removing you all as reviewers thanks.

config/monitor_intentions_test.go
```diff
@@ -0,0 +1,493 @@
...
+			&IntentionsMonitorConfig{SourceServices: &IntentionsServicesConfig{Regexp: String("same")}},
+		},
+		{
+			"regexp_empty_two",
+			&IntentionsMonitorConfig{},
+			&IntentionsMonitorConfig{SourceServices: &IntentionsServicesConfig{Regexp: String("same")}},
+			&IntentionsMonitorConfig{SourceServices: &IntentionsServicesConfig{Regexp: String("same")}},
+		},
+		{
+			"regexp_empty_same",
```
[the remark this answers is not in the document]
noejbrown: Empty is not the right term, that is a typo. It should be "regexp_same". I included this test because it aligns with the other module inputs having the same test. When you say it test the the same code path can you elaborate on that?

config/module_input_test.go
```diff
@@ -335,10 +444,10 @@ func TestModuleInputConfigs_Merge(t *testing.T) {
 		{
 			"happy_path_different_type",
 			&ModuleInputConfigs{&ServicesModuleInputConfig{}},
-			&ModuleInputConfigs{&ConsulKVModuleInputConfig{}},
```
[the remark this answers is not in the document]
noejbrown: Yes, I wanted to test that the "happy_path_different_type" different worked with Intentions module input and for this test it only allowed for two configs. Adding a third config would be too many values for the struct literal.

config/monitor_intentions_test.go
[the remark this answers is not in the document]
hashi-derek: Sure. So what I'm saying is that the `regexp_empty_same` and `regexp_overrides` functions performed the same actions and followed the same if-else statements. The only difference was the data. So it's not necessary to create the `regexp_empty_same` test.

config/module_input_test.go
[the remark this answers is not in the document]
noejbrown: Thanks for explaining that, I added back the previous test and added another test case specifically for intentions.

config/monitor_intentions_test.go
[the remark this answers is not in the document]
hashi-derek: Oh okay. Maybe I misunderstood something. Is it possible to have duplicate the data for those string fields? I thought that was only possible for lists.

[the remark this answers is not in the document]
noejbrown: I assumed so when I was looking at the other module input test as reference and figured that would be the main reason for that test case.

hashi-derek:
This PR looks good, but not marking it as approved yet. We will utilize this as the feature branch for all intentions work, so there will be other PRs merged into this.

config/intentions_service.go
```diff
@@ -0,0 +1,130 @@
...
+	// Check that either regex or names is configured but not both
+	namesConfigured := c.Names != nil && len(c.Names) > 0
+	regexConfigured := c.Regexp != nil
+
+	if namesConfigured && regexConfigured {
+		return fmt.Errorf("regexp and names fields cannot both be " +
+			"configured. If both are needed, consider including the list of " +
+			"names as part of the regex or creating separate tasks")
+	}
+	if !namesConfigured && !regexConfigured {
```
jwair, line 88:
```suggestion

	if !namesConfigured && !regexConfigured {
```

Space out if blocks to improve readability. This might be misread as an `else if` because it feels part of the previous if block

config/intentions_service.go
```diff
@@ -0,0 +1,130 @@
...
+	return nil
+}
+
+func (c *IntentionsServicesConfig) GoString() string {
+	if c == nil {
+		return ": (*IntentionsServicesConfig)(nil)"
+	}
+
+	if len(c.Names) > 0 {
+		return fmt.Sprintf(
```
jwair, line 118: Can you please fix the style for Sprintf() here (and in other place in this PR) to be on the same line instead of splitting it to multiple lines?

config/module_input_intentions.go
```diff
@@ -0,0 +1,93 @@
...
+	return nil
+}
+
+// GoString defines the printable version of this struct.
+func (c *IntentionsModuleInputConfig) GoString() string {
+	if c == nil {
+		return "(*IntentionsModuleInputConfig)(nil)"
+	}
+
+	return fmt.Sprintf("&IntentionsModuleInputConfig{"+
```
jwair, line 88: Same as the previous comment about inlining Sprintf()

config/module_input_intentions_test.go
```diff
@@ -0,0 +1,277 @@
...
+					},
+				},
+			},
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			r := tc.a.Copy()
+			if tc.a == nil {
```
jwair, line 52: Can you please elaborate on this?

config/module_input_intentions_test.go
```diff
@@ -0,0 +1,277 @@
...
+		})
+	}
+}
+
+func TestIntentionsModuleInputConfig_Merge(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name string
+		a    *IntentionsModuleInputConfig
```
jwair, line 67: I suggest improving these names for better readability

config/intentions_service.go
```diff
@@ -0,0 +1,130 @@
...
+
+	return nil
+}
+
+func (c *IntentionsServicesConfig) GoString() string {
+	if c == nil {
+		return ": (*IntentionsServicesConfig)(nil)"
+	}
+
+	if len(c.Names) > 0 {
```
jwair, line 117: I am not sure about the correctness of this logic
I understand that we want to specify either Names OR Regex, but that is business logic, and the struct doesn't know about this restriction. Therefore, IF I create an IntentionsServicesConfig object with both Names and Regex defined, I expect the `GoString()` method to return a string with both fields, not just Names.

config/intentions_service.go
```diff
@@ -0,0 +1,130 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+func (c *IntentionsServicesConfig) GoString() string {
+	if c == nil {
+		return ": (*IntentionsServicesConfig)(nil)"
```
jwair, line 114: What is the `: ` for?
Compared with https://github.com/hashicorp/consul-terraform-sync/pull/987/files#diff-9d9d2ad5226133163e12e5eebac672fc37d2b51772b1b680928036381c9fc9c7R85 which doesn't have a `: ` prefix

config/intentions_service.go
```diff
@@ -0,0 +1,130 @@
...
+		return
+	}
+
+	if c.Names == nil {
+		c.Names = []string{}
+	}
+}
+
+func (c *IntentionsServicesConfig) Validate() error {
+	if c == nil { // config not required, return early
```
jwair, line 75: **config** not required?

config/intentions_service.go
```diff
@@ -0,0 +1,130 @@
...
+			"Names:%s",
+			c.Names,
+		)
+
+	} else {
+		return fmt.Sprintf(
+			"Regexp:%s",
+			StringVal(c.Regexp),
+		)
+
```
jwair, line 128:
```suggestion
```

config/intentions_service.go
hashi-derek: I believe that the nil check here is because the parent config struct could technically have a nil pointer to an undefined child struct. The parents recursively call Validate on their children. It's a pattern seen elsewhere in the code.

config/intentions_service.go
noejbrown: Thank you will update!

config/module_input_intentions.go
```diff
@@ -0,0 +1,93 @@
...
+	c.IntentionsMonitorConfig.Finalize()
+}
+
+// Validate validates the values and required options. This method is recommended
+// to run after Finalize() to ensure the configuration is safe to proceed.
+func (c *IntentionsModuleInputConfig) Validate() error {
+	if c == nil { // config not required, return early
+		return nil
+	}
+	if err := c.IntentionsMonitorConfig.Validate(); err != nil {
```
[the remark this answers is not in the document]
noejbrown: Updated

config/intentions_service.go
noejbrown: This is used for readability. The colon is used specifically for the intentions services (source services and destination services) config being nil. Example  :

- It usually prints like this when all fields are configured 
&IntentionsModuleInputConfig{&IntentionsMonitorConfig{Datacenter:dc, Namespace:namespace, Source Services Regexp:^web.*, Destination Services Regexp:^api.*}}

- If source/destination services fields are not configured it would instead print like 
&IntentionsModuleInputConfig{&IntentionsMonitorConfig{Datacenter:dc, Namespace:namespace, Source Services : (*IntentionsServicesConfig)(nil), Destination Services : (*IntentionsServicesConfig)(nil)}}

config/intentions_service.go
noejbrown: As Derek explained this is a nil check for the parent config

config/intentions_service.go
noejbrown: Updated

config/intentions_service.go
noejbrown: Updated

config/module_input_intentions.go
noejbrown: Updated

config/module_input_intentions_test.go
noejbrown: I chose these names to align with all other module input test files. They all use a, b, and r. I felt a deviation from them would be confusing, but still open to changing if you feel it would be better.

config/module_input_intentions_test.go
noejbrown: An interface only equals nil if both the type and value are nil therefore we can not do an assert.Nil(t, tc.a, r). Interface nil type is not equal to tc.a so first we must check tc.a equals nil and then we can assert nil for t and r.

config/module_input_intentions_test.go
```diff
@@ -0,0 +1,277 @@
...
+
+func TestIntentionsModuleInputConfig_Copy(t *testing.T) {
+	t.Parallel()
+
+	finalizedConf := &IntentionsModuleInputConfig{}
+	finalizedConf.Finalize()
+
+	cases := []struct {
+		name string
+		a    *IntentionsModuleInputConfig
```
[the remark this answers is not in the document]
noejbrown: I chose the field a to align with all other module input test files. They all use a, b, and r in the same manner.

config/intentions_service.go
noejbrown: This logic behind this is for the expected use case that both fields wouldn't be defined. If you create IntentionsServicesConfig object with both field defined, the validate function that is prior to this would fail. I could see putting an error message here instead if both are configured but showing both didn't make sense to me because the output would be : 

`&IntentionsModuleInputConfig{&IntentionsMonitorConfig{Datacenter:dc, Namespace:namespace, Source Services: Names: Regexp:^web.*, Destination Services: Names: Regexp:^api.*}}`  

Could you explain more what the use case for showing both names and regexp would be? I think it would steer people into the wrong direction that you could use both and lead to test failure.

PR #998, opened 2022-07-20
API middleware for intercepting requests

mkam:
General purpose middleware for intercepting an API request based on the given condition. Also removes the client package's dependency on the API package.


mkam:
@wilkermichael Thanks for the feedback about renaming overwrite to intercept! I've made the name changes now, let me know if there's anything else that needs improving!

wilkermichael (approved):
Looks good to me

api/api.go
```diff
@@ -70,10 +70,11 @@ type API struct {
 }
 
 type Config struct {
```
wilkermichael, line 72: Just noticed this, this should have a public comment

PR #999, opened 2022-07-21
Fix issue with task config incorrectly modified

### LINKED ISSUE #1000: Endpoints returning incorrect configuration information
### Describe the bug
Issue where the API responses for `Post /task`, `Get /task` and `Get /task/:name` return objects that differs from what was actually configured. 

### Versions
**Consul-Terraform-Sync**
```
consul-terraform-sync >= v0.5.0 (dbfcb19)
Compatible with Terraform >= 0.13.0, < 0.15
```

**Consul Version**
```
N/A
```

**Terraform Version**
```
N/A
```

### Expected Behavior
Expect the following:

1. The response objects returned by `Get /task/:name`, `Get /tasks` and `Post /tasks` for the same task are equivalent
2. The response objects are reflective of the tasks as they are configured

### Actual Behavior
The response objects for `Get /task/:name` and `Get /tasks` are equivalent for tasks of the same name, however response objects returned by `Post /tasks` are different.

The objects returned also contain transformations that occur after configuration has been processed and converted to an internal task object for use with automation.

Examples:
- `task.working_dir`, if unset, will be populated with the parent `working_dir` which is not what was configured.
- `task.buffer_period`, if unset, will be populated with the parent `buffer_period` which is not what was configured.
- When using a local `task.module` this field is populated with an absolute path, rather than the configured relative path.

The above transformations are required when using the task for automations, so fixing this bug would not change any of the existing behaviour, tasks will still inherit qualities from the parent configuration on usage. This fix will instead create consistency between endpoints and convey to the user exactly what was configured.

### Steps to Reproduce
1. Create a task via the api
2. Assess the response
3. Get the same task
4. Assess the response
5. Get all tasks
6. Assess the object pertaining to the task created in (1)

hashi-derek:
Resolves #1000

### DIFF
--- a/config/buffer_period.go
+++ b/config/buffer_period.go
@@ -78,12 +78,12 @@ func (c *BufferPeriodConfig) Merge(o *BufferPeriodConfig) *BufferPeriodConfig {
 }
 
 // Finalize ensures that the receiver contains no nil pointers. For nil pointers,
-// Finalize sets these values using the passed in "parent" BufferPeriodConfig
-// along with using other defaults where necessary.
-//
-// Example parent param: global-level buffer period uses default values,
-// task-level buffer period uses the global-level buffer period
-func (c *BufferPeriodConfig) Finalize(parent *BufferPeriodConfig) {
+// Finalize sets default values where necessary
+func (c *BufferPeriodConfig) Finalize() {
+	c.inheritParentConfig(DefaultBufferPeriodConfig())
+}
+
+func (c *BufferPeriodConfig) inheritParentConfig(parent *BufferPeriodConfig) {
 	// if disabled, fill in zero values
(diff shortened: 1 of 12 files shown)


config/buffer_period.go
```diff
@@ -115,6 +115,10 @@ func (c *BufferPeriodConfig) Finalize(parent *BufferPeriodConfig) {
 	if c.Max == nil {
 		c.Max = TimeDuration(4 * *c.Min)
 	}
+
+	if *c.Min > *c.Max {
```
wilkermichael, line 119: Can you add a quick unit test for this case?

wilkermichael (approved):
Looks great, thanks so much for working on this with me

PR #1062, opened 2023-05-17
Version bumps

tristanmorgan:
An attempt to update the versions of go modules, Golang and Terraform that CTS works with.


tristanmorgan:
version updated in #1071, closing as a duplicate.

PR #1067, opened 2023-10-13
Increase time.Sleep

absolutelightning:
Issue - NET-4886

Issue - 
When services are registered with checks, initially they are in `critical` status. When health check passes, they get in `passing` status.  Without this delay, CatalogServices may have services that are not yet propagated to HealthServices as healthy services since services are initially set as critical. 

The API call from CTS to Consul has `passing` [set to true](https://github.com/hashicorp/consul-terraform-sync/blob/main/templates/tftmpl/tmplfunc/services_regex.go#L209). It only returns passing health check services or healthy services.

The delay is small. It might take more time to make service healthy. Hence this PR increases the delay.

Tests - 
Manually tested it using Nomad and Consul + CTS setup using Terraform Cloud driver. Changes in PR has resolved the issue.
Also verified by @maksimnosal

### DIFF
--- a/templates/tftmpl/tmplfunc/services_regex.go
+++ b/templates/tftmpl/tmplfunc/services_regex.go
@@ -201,7 +201,7 @@ func (d *servicesRegexQuery) Fetch(clients dep.Clients) (interface{}, *dep.Respo
 	// Without this delay, CatalogServices may have services that are not yet
 	// propagated to HealthServices as healthy services since services are initially
 	// set as critical. https://www.consul.io/docs/discovery/checks#initial-health-check-status
-	time.Sleep(5 * time.Second)
+	time.Sleep(10 * time.Second)
 
 	var services []*dep.HealthService
 	for _, s := range matchServices {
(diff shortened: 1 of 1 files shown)


templates/tftmpl/tmplfunc/services_regex.go
```diff
@@ -201,7 +201,7 @@ func (d *servicesRegexQuery) Fetch(clients dep.Clients) (interface{}, *dep.Respo
 	// Without this delay, CatalogServices may have services that are not yet
 	// propagated to HealthServices as healthy services since services are initially
 	// set as critical. https://www.consul.io/docs/discovery/checks#initial-health-check-status
-	time.Sleep(1 * time.Second)
+	time.Sleep(10 * time.Second)
```
Ganeshrockz, line 204: Any reason behind increasing the sleep duration by 10x? Wondering if there is a way to deterministically figure out rather than following a trial and error based approach

absolutelightning: Tried on 5x. Issue was still seen.
Could not thing of any other way. 😢

Ganeshrockz (approved):

mkam (approved):

PR #1071, opened 2023-10-27
NET 6318 - Go/Terraform Version Upgrade + E2E passing

absolutelightning:
Bump Go Version - 1.20.10
Updated Terraform Version - 1.6.2
Make E2E tests passing

### DIFF
--- a/e2e/condition_schedule_test.go
+++ b/e2e/condition_schedule_test.go
@@ -121,7 +121,7 @@ func TestCondition_Schedule_Basic(t *testing.T) {
 
 			port := cts.Port()
 			taskSchedule := 10 * time.Second
-			scheduledWait := taskSchedule + 7*time.Second // buffer for task to execute
+			scheduledWait := taskSchedule + 57*time.Second // buffer for task to execute
 
 			// 0. Confirm at least one event for once-mode
 			eventCountBase := eventCount(t, taskName, port)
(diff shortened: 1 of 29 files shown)


e2e/condition_schedule_test.go
```diff
@@ -157,6 +157,7 @@ func TestCondition_Schedule_Basic(t *testing.T) {
 			// check scheduled task did not trigger immediately and ran only on schedule
 			api.WaitForEvent(t, cts, taskName, registerTime, scheduledWait)
 			checkScheduledRun(t, taskName, registerTime, taskSchedule, port)
+			time.Sleep(defaultWaitForTestReadiness)
```
absolutelightning, line 160: Had to add this and the same one below this to make the E2E test pass.
Will this be because of https://github.com/hashicorp/consul-terraform-sync/pull/1067 ?

mkam: Most likely! This is fine in the short term to get the release back on track, but could you follow up later and add a more dynamic wait? Like in `validateService`, keep checking for the expected state with more of a polling mechanism.

mkam (approved):
LGTM. You'll also need to add entries in the changelog for the go version upgrade and the Terraform support either in this PR or in a follow up PR.

e2e/condition_schedule_test.go
mkam: Also might be better to have a shorter wait that's closer to the delay you added. So maybe 15 seconds instead of 20 seconds. But not a big deal right now since you're trying to release.

PR #1087, opened 2024-03-18
Update go versions to 1.21.8

kisunji:


dduzgun-security (commented):
@kisunji We may need to edit the constraint range to support newer terraform versions here: https://github.com/hashicorp/consul-terraform-sync/blob/main/version/terraform.go#L19 and bumping the action to v3 `hashicorp/setup-terraform@v3`. That should fix the E2E tests

dduzgun-security (approved):

PR #1089, opened 2024-05-22
bump go version to 1.22.4

dhiaayachi:
Upgrade go CI to use 1.22.4


dhiaayachi:
@DanStough CI has been failing for a while now. I tried to fix it but I don't really understand any of those tests and it will take me a long time. I will merge this as is (after your approval) and then we will decide what to do about those failing tests.

DanStough (approved):

PR #1091, opened 2024-06-28
Update build.yml

NicoletaPopoviciu:
remove macOS runners from buildworkflow

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -80,6 +80,7 @@ jobs:
           - { go: "1.22.4", goos: "windows", goarch: "amd64" }
           - { go: "1.22.4", goos: "solaris", goarch: "amd64" }
           - { go: "1.22.4", goos: "darwin", goarch: "amd64" }
+          - { go: "1.22.4", goos: "darwin", goarch: "arm64" }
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -79,6 +79,7 @@ jobs:
           - { go: "1.22.4", goos: "windows", goarch: "386" }
           - { go: "1.22.4", goos: "windows", goarch: "amd64" }
           - { go: "1.22.4", goos: "solaris", goarch: "amd64" }
+          - { go: "1.22.4", goos: "darwin", goarch: "amd64" }
```
DanStough, line 82: I know this wasn't here before but we might as well add it since M1 macbook support is quite common now. It goes to show you how little milage this repo gets.
```suggestion
          - { go: "1.22.4", goos: "darwin", goarch: "amd64" }
          - { go: "1.22.4", goos: "darwin", goarch: "arm64" }
```

DanStough (changes requested):

DanStough (approved):
As I understand, the acceptance tests have been broken for some time, so this is safe to merge.

DanStough (approved):

